PR #2, opened 2011-10-06
Add riak-erlang-client as a dependency

reiddraper:
AZ820


reiddraper:
At this point you should be able to do something like this:

``` bash
curl -X GET -v \
  -H "authorization: AWS foo:bar"\
   "http://0.0.0.0:8080/bucket"
```

and have it return an error that you aren't authorized.

kellymclaughlin:
+1 to merge.

PR #8, opened 2011-10-12
Create the beginnings of a pluggable authentication mechanism.

kellymclaughlin:
- Add configuration entries to the app.config file to pass an
  authentication module name to the webmachine resources.
- Add an entry to the vars.config file so that the app.config for a
  release will specify the passthru auth module for all resources.
- Create an auth module behavior, riak_moss_auth.
- Create three auth modules: one to pass all requests, one to block
  all requests, and a stub for s3 authentication.
- Change the service, bucket, and key resources to get the
  authentication module from the config during init and add it to the
  context.
- Add a context record to the riak_moss.hrl file.

This actually ended up encapsulating the work described in [az829](https://agilezen.com/project/17206/story/829) and [az834](https://agilezen.com/project/17206/story/834) as well as [az822](https://agilezen.com/project/17206/story/822), but it actually would have been less efficient to separate them out into more individual pieces.

### DIFF
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -6,8 +6,10 @@
           buckets = []}).
 
 -record(rs3_bucket, {
-          name :: binary(), 
+          name :: binary(),
           creation_date :: term()}).
+
+-record(context, {auth_mod :: atom()}).
 
 -define(USER_BUCKET, <<"moss.users">>).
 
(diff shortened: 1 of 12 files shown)


kellymclaughlin:
Agreed, I just planned to save that for a later refactoring pass, but I can go ahead and do it now just as easy.

reiddraper:
+1

PR #9, opened 2011-10-12
Remove Apache license header from all source files.

kellymclaughlin:
`riak_moss` is not currently open source so we do not need the Apache
license on the source files.


slfritchie:
Seems to match the pattern of the riak_repl repo, +1 to merge, thanks!

PR #10, opened 2011-10-13
Rework setup for webmachine dispatch and authentication.

kellymclaughlin:
Added a `riak_moss_web` module to handle the setup of the webmachine dispatch table. Also use an `auth_bypass` configuration option to enable authentication to be disabled.


reiddraper:
+1

PR #11, opened 2011-10-13
Az845 refactor client

reiddraper:
These public APIs have now been refactored and manually tested.
- `get_user`
- `create_user`
- `get_buckets`
- `create_bucket`
- `delete_bucket`
- `get_object`
- `put_object`


kellymclaughlin:
I dont' see any reason not to go ahead and get rid of `riak_moss_client` and `riak_moss_resource` and delete the `moss_client` function from the `riak_moss` module.

src/riak_moss_riakc.erl
```diff
@@ -32,14 +32,18 @@
...
+%% ====================================================================
+%% Public API
+%% ====================================================================
 
 start_link() ->
     gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).
 
 init([]) ->
-    {ok, #state{}}.
+    {ok, Pid} = riak_moss:riak_client(),
```
kellymclaughlin, line 45: Just having the single riak connection is going to be insufficient. Maybe we don't have to change this for this pull request, but we at least need a kanban to implement a riak connection pool in the near future.

src/riak_moss_riakc.erl
```diff
@@ -50,74 +54,156 @@ create_user(UserName) ->
...
+%% the KeyID?
+put_object(KeyID, BucketName, Key, Val, Metadata) ->
+    gen_server:call(?MODULE, {put_object, KeyID, BucketName, Key, Val, Metadata}).
 
-do_save_user(User) ->
-    {ok, Pid} = riak_moss:riak_client(),
-    UserObj = riakc_obj:new(?USER_BUCKET,list_to_binary(User#rs3_user.key_id),User),
-    ok = riakc_pb_socket:put(Pid, UserObj),
-    riakc_pb_socket:stop(Pid),
+handle_call({get_user, KeyID}, _From, State=#state{riakc_pid=RiakcPid}) ->
```
kellymclaughlin, line 80: I wonder if we're doing too much in this module in synchronous call handling. Could we limit the `gen_server` activity to handing out and checking back in riak connection pids. _e.g._ `do_create_user` could become an exported function for public consumption and then we wouldn't make the synchronous server call until right before the actual call to `riakc_pb_socket:put` in `do_save_user` just to obtain a connection pid. Afaict, that's the only reason for using the `gen_server` here.

src/riak_moss_riakc.erl
reiddraper, line 45: I've created [az858](https://agilezen.com/project/17206/story/859).

src/riak_moss_riakc.erl
reiddraper, line 80: I like this idea. I've added it to [az859](https://agilezen.com/project/17206/story/859)

kellymclaughlin:
The `moss_client` function in the `riak_moss` module can be remove too.

kellymclaughlin:
+1 to merge

PR #14, opened 2011-10-17
Support multiple simultaneous auth schemes 

kellymclaughlin:
Reorganize how authentication is handled to allow multiple authentication schemes to be used at one time. The `parse_header` function is now what determines which authentication scheme is being used and it now lives in the `riak_moss_wm_util` module. The `authenticate` function in the auth modules now has arity 2. The first parameter being the request data and the second a list of scheme-specific terms that are created by `riak_moss_wm_util:parse_header`. Also the `auth_mod` field in the `context` record has been replaced with an `auth_bypass` boolean field to indicate if authentication bypass is enabled.

### DIFF
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -10,8 +10,6 @@
           creation_date :: term()}).
 
 -record(context, {auth_bypass :: atom(),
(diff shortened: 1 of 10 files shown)


reiddraper:
+1 to merge

PR #15, opened 2011-10-17
Change the rs3_user and rs3_bucket records to moss_user and moss_bucket

kellymclaughlin:
Change the name of the `rs3_user` record to `moss_user` and update all references in applicable modules. Likewise, do the same for the `rs3_bucket` record, renaming it to `moss_bucket`.


reiddraper:
+1

PR #16, opened 2011-10-17
Az871 user in test auth

reiddraper:


src/riak_moss_passthru_auth.erl
```diff
@@ -10,14 +10,21 @@
...
+-export([authenticate/2]).
 
--spec authenticate(term()) -> {ok, #rs3_user{}}
-                        | {ok, unknown}
-                        | {error, atom()}.
-authenticate(_RD) ->
+-spec authenticate(term(), [string()]) -> {ok, #rs3_user{}}
+                                              | {ok, unknown}
+                                              | {error, atom()}.
+authenticate(_RD, KeyID) when is_list(KeyID) ->
```
kellymclaughlin: This eliminates the possibility of using passthru with no user id. As it stands only this first function clause will ever match because `is_list` matches even empty lists. Probably want to a add a function clause above this one to match on `[]` as the 2nd parameter. You could leave the final clause as a catch-all, but since we know the input will be a list I think it's ok to drop it.

reiddraper: Good catch, pushed e712172e7de6ad5e2f978632443594d8a9196544 to fix.

src/riak_moss_passthru_auth.erl
```diff
@@ -10,17 +10,23 @@
...
-    %% Even though we're 
+    %% Even though we're
     %% going to authorize
     %% no matter what,
     %% maybe it still makes sense
     %% to pass the user on
-    {ok, unknown}.
-
+    {ok, unknown};
+authenticate(_RD, KeyID) when is_list(KeyID) ->
```
kellymclaughlin: You can get rid of the guard here and also the second parameter should be `[KeyID]` for it to work properly.

reiddraper: I thought I was crazy because I did test this. However I've found that these two work the same (for reasons yet unknown)

``` erlang
riak_moss_riakc:get_user(KeyID).
riak_moss_riakc:get_user([KeyID]).
```

The code should definitely be changed to [KeyID] but I'd like to find out why it seemed to be working.

kellymclaughlin:
+1 to merge.

PR #17, opened 2011-10-18
Az860 connect resources with client

reiddraper:


src/riak_moss_wm_service.erl
```diff
@@ -64,21 +64,20 @@ content_types_provided(RD, Ctx) ->
...
-    %% extract the user from the auth
-    %% and retrieve their list of
-    %% buckets. For now we just return
-    %% an empty list.
-    Return_body = [],
-    {mochijson2:encode(Return_body), RD, Ctx}.
+-spec to_json(term(), term()) ->
+    {iolist(), term(), term()}.
+
+to_json(RD, Ctx=#context{user=User}) ->
```
kellymclaughlin, line 81: You get an ugly error here if you haven't specified a user. Probably a good time to add handling of `unknown` Users in `forbidden`.

reiddraper, line 81: good point, fixed in 0b10ea3e1bcf05dbd4baab19bf21a8aa61537112

kellymclaughlin:
+1 to merge.

PR #18, opened 2011-10-19
basic packaging for .deb using node_package

argv0:


reiddraper:
worked for me on osx and ubuntu. +1

PR #19, opened 2011-10-20
Az880 boto test

argv0:
python unit tests for parity with prototype.   works with riak-s3-proxy.


reiddraper:
Looks good. My only comment is I just made a `test` folder, and this is in `tests`.

argv0:
Ah, no problem.  I'll change it.

PR #20, opened 2011-10-21
Az870 eunit rest tests

reiddraper:


reiddraper:
Currently one test is known to fail because we don't account for missing objects. The pattern just allows `{ok, Object}` instead of also allowing, `{error, Reason}`.

kellymclaughlin:
Since there is a `test` directory at the top level now, need to add `test` to the list of `Phony` targets in the Makefile; otherwise the test won't run with `make test`.

kellymclaughlin:
I think it'd be good to get rid of lines 12 and 13 in `riak_moss_wm_test_utils` and replace the env section in `riak_moss.app.src` with this:

<pre><code>
  {env, [
         {moss_ip, "127.0.0.1"},
         {moss_port, 8080}
        ]} 
</code></pre>

test/riak_moss_riakc_test.erl
```diff
@@ -0,0 +1,109 @@
...
+
+-module(riak_moss_riakc_test).
+
+-export([riakc_test_/0]).
+
+-include("riak_moss.hrl").
+-include_lib("eunit/include/eunit.hrl").
+
+setup() ->
+    application:set_env(riak_moss, moss_ip, "127.0.0.1"),
```
kellymclaughlin: Could also then get rid of these lines after the change to the `.app.src` file.

test/riak_moss_riakc_test.erl
```diff
@@ -0,0 +1,109 @@
...
+                               B <- riak_moss_riakc:get_buckets(UserAfter)],
+    ?assertEqual([BucketName], AfterBucketNames).
+
+
+%% @doc Make sure that when we create a key
+%%      that is appears in list keys for that
+%%      bucket
+key_appears() ->
+    %% TODO:
+    %% How much of a hack is it
```
kellymclaughlin, line 74: Yes, it's a hack, but I would say leave it as-is for now. There is a lot of list handling currently, but since that's what we get from webmachine via mochiweb, we have to deal with it. Ideally we could get binaries directly from webmachine, but I think we can optimize that later.

kellymclaughlin:
Looks good. +1 to merge.

PR #22, opened 2011-10-24
885 delete object support

reiddraper:


src/riak_moss_wm_key.erl
```diff
@@ -80,12 +99,20 @@ content_types_provided(RD, Ctx) ->
     %% will either come from the value that was
     %% last PUT or, from you adding a
     %% `response-content-type` header in the request.
-    DocCtx = riak_moss_wm_utils:ensure_doc(Ctx),
-    Doc = DocCtx#key_context.doc,
-    ContentType = riakc_obj:get_content_type(Doc),
+    case lists:member(wrq:method(RD), ['GET', 'HEAD']) of
```
kellymclaughlin: I think this would be more readable and flow better as an `if` statement.

kellymclaughlin:
+1 to merge.

PR #27, opened 2011-10-26
Az880 xml resources

argv0:
parity with prototype implementation.


reiddraper:
The WM tests that expect `to_json` don't work anymore. I vote we just remove them. One of the s3 auth changes seems to have broken the `example_upload_test`.

reiddraper:
[here's](https://gist.github.com/64b9418253d291145b10) a patch that makes both eunit and python functional test pass.

reiddraper:
+1 to merge

PR #29, opened 2011-11-12
Changes so s3cmd bucket listing works.

kellymclaughlin:
- Change bucket creation date format to iso8601.
- Normalize all headers before amazon header extraction so that
  matching the `x-amz-date` header always works.


reiddraper:
a couple of stray `io:format`'s, +1 otherwise

PR #30, opened 2011-11-15
Az918 basho bench

reiddraper:
In order to test, I had to do a few things:
- create a user and put the key_id in the driver config
- add my moss repo bench folder to the config `source_dir`

Here is my config, for reference:

``` erlang
{mode, max}.

{duration, 1}.

{concurrent, 1}.

{driver, basho_bench_driver_moss}.

{moss_authorization, "4844552620141D3E1EA71090A9D27F363F9CFA25"}.

{code_paths, ["deps/ibrowse"]}.

{value_generator, {fixed_bin, 1000}}.

{http_raw_port, 8080}.

{operations, [{insert, 1}, {update, 1}, {delete, 1}, {get, 1}]}.

{source_dir, "/Users/reid/Documents/repos/basho/riak_moss/bench"}.
```


dizzyd:
What is basho_bench.hrl providing that you are using? I don't see any debug statements or anything....

reiddraper:
@dizzyd good point, I had just put it in there from the beginning when I was looking at the other drivers. Will remove

bench/basho_bench_driver_moss.erl
```diff
@@ -0,0 +1,262 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2007-2011 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% -------------------------------------------------------------------
+
+-module(basho_bench_driver_moss).
```
kellymclaughlin, line 13: Could you add a comment indicating that `auth_bypass` must be set to `true`. My initial run failed b/c I had it as `false`.

kellymclaughlin:
I would suggest adding the sample config you provided to the `bench` dir as `riak_moss.config.sample` or something similar.

bench/basho_bench_driver_moss.erl
reiddraper, line 13: Apologies, added a comment in 3003ad73dec7d7994e5768fe17469b25a98be7bc

reiddraper:
Added the same config in 3003ad73dec7d7994e5768fe17469b25a98be7bc

kellymclaughlin:
+1 to merge

PR #35, opened 2011-12-07
Az940 get fsm

reiddraper:
## Testing

First add this to the app.config:

```
{lfs_block_size, 2}, % yep, 2 :)
```

Now put a "file" into moss:

``` erlang
{ok, _Pid} = riak_moss_put_fsm_sup:start_put_fsm(node(), [<<"foo">>, <<"bar">>, 10, <<0,1,2,3,4,5,6,7,8,9>>, 10000]).

%% start the get fsm
{ok, Pid} = riak_moss_get_fsm_sup:start_get_fsm(node(), [self(), <<"foo">>, <<"bar">>]).

%% receive the metadata for the object
receive {metadata, Meta} -> io:format("got ~p ~n", [Meta]) after 100 -> io:format("timeout") end.

%% we've decided we actually need the value (ie. not just a HEAD request)
%% so tell the fsm to start retrieving the blocks (or just return the single value)
riak_moss_get_fsm:continue(Pid).

Recv = fun (_) -> receive Foo -> io:format("got ~p ~n", [Foo]) after 1000 -> io:format("timeout") end end.

lists:foreach(Recv, lists:seq(1, 5)).
```
## Todo
- There are currently no timeouts on the fsm
- the the proc that retrieves the procs should likely be a gen_server, at least when we have parallelism
- the proc that receives the blocks uses `riak_moss_riakc`
- replace the dummy gets module with meck usage


src/riak_moss_get_fsm.erl
```diff
@@ -0,0 +1,214 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2007-2011 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% -------------------------------------------------------------------
+
+%% get fsm for Riak Moss.
```
kellymclaughlin: Add edoc notation (@doc) here.

src/riak_moss_get_fsm.erl
```diff
@@ -0,0 +1,214 @@
...
+    %% message format
+        false ->
+            %% send metadata back to
+            %% the `from` part of
+            %% state
+            %% TODO:
+            %% we don't deal with siblings here
+            %% at all
+            Metadata = riakc_obj:get_metadata(Value),
+            CachedValue = riakc_obj:get_value(Value),
```
kellymclaughlin: Why call `riakc_obj:get_value` at L92 and again here? You could store the binary value as `RawValue` and then change L92 to `binary_to_term(RawValue)` and save the extra lookup.

src/riak_moss_get_fsm.erl
```diff
@@ -0,0 +1,214 @@
...
+    %% fetch the value, be it manifest
+    %% or regular object
+    spawn_link(?MODULE, normal_retriever, [self(), GetModule, Bucket, Key]),
+    {next_state, waiting_value, State}.
+
+waiting_value({object, Value}, #state{from=From}=State) ->
+    %% determine if the object is a normal
+    %% object, or a manifest object
+    DecodedValue = binary_to_term(riakc_obj:get_value(Value)),
+    %%DecodedValue = riakc_obj:get_value(Value),
```
kellymclaughlin: Delete this

kellymclaughlin:
The `StateName` functions are actually `gen_fsm` callbacks, so organizationally you might want to group them with the other `gen_fsm` callbacks, both in the export list and the definitions.

src/riak_moss_get_fsm.erl
reiddraper: addressed in 3290c2d6f13a9a24e916157d2c47266d1e675424

src/riak_moss_get_fsm.erl
reiddraper: good catch. changed in 3e8b04407719857230a060cbf8a92ba04a096d41

src/riak_moss_get_fsm.erl
reiddraper: deleted in 3e8b04407719857230a060cbf8a92ba04a096d41

test/riak_moss_get_fsm_test.erl
```diff
@@ -0,0 +1,83 @@
...
+    application:stop(riak_moss),
+    application:stop(sasl).
+
+get_fsm_test_() ->
+    {setup,
+     fun setup/0,
+     fun teardown/1,
+     [
+      fun receives_metadata/0,
+      test_n_chunks_builder(1),
```
kellymclaughlin: You could use this to save a few lines and make it easier to add different numbers of chunks.
<code>
[test_n_chunks_builder(X)   || X <- [1,2,5,9,100,1000]]
</code>

test/riak_moss_get_fsm_test.erl
```diff
@@ -0,0 +1,83 @@
...
+receive_with_timeout(Timeout) ->
+    receive
+        Anything ->
+            Anything
+    after
+        Timeout -> timeout
+    end.
+
+expect_n_chunks(N) ->
+    lists:foreach(fun (_) ->
```
kellymclaughlin: This would be more clear as a list comprehension:

`[?assertMatch({chunk, _}, receive_with_timeout(100)) || _ <- lists:seq(1, N-1)],`

src/riak_moss_get_fsm.erl
```diff
@@ -0,0 +1,232 @@
...
+            %% we don't deal with siblings here
+            %% at all
+            Metadata = riakc_obj:get_metadata(Value),
+            From ! {metadata, Metadata},
+            {next_state, waiting_chunk_command, State#state{value_cache=RawValue}, NextStateTimeout};
+        true ->
+            Metadata = riak_moss_lfs_utils:metadata_from_manifest(DecodedValue),
+            From ! {metadata, Metadata},
+            StateWithMani = State#state{manifest=DecodedValue},
+            {next_state, waiting_chunk_command, StateWithMani, NextStateTimeout}
```
kellymclaughlin: It would be a little more clear and concise if you assign `NextState` in the case clause and move the message sending and next state specification after the case. So all the case would do would be to assign `Metadata` and `NextState`.

src/riak_moss_get_fsm.erl
```diff
@@ -0,0 +1,232 @@
...
+    end.
+
+waiting_chunk_command(stop, State) ->
```
kellymclaughlin, line 128: Add a function clause to handle the timeout. Since it's valid to only get the head and for the timeout to occur, you can just transition using a normal stop.

kellymclaughlin:
+1 to merge

PR #36, opened 2011-12-07
Az992 wm lfs put

argv0:
integrate the put_fsm and webmachine.  changes:
- writer incrementally calculates md5sum and stores it in manifest during finalization
- first draft at a finalization protocol for the put fsm: riak_moss_put_fsm:finalize(Pid) -> {ok, lfs_manifest()}.
- refactoring of ensure_doc logic in riak_moss_wm_key
- various changes to adapt to new bucket naming conventions.


src/riak_moss_put_fsm.erl
```diff
@@ -82,6 +86,21 @@ augment_data(Pid, Data) ->
             end
     end.
 
+%% @doc Augment the file data being written by a `riak_moss_put_fsm'.
```
kellymclaughlin: Comment needs to be updated.

src/riak_moss_put_fsm.erl
```diff
@@ -82,6 +86,21 @@ augment_data(Pid, Data) ->
...
+-spec finalize(pid()) -> ok | {error, term()}.
+finalize(Pid) ->
```
kellymclaughlin, line 82: We might as well go ahead and break this out into an internal `send_sync_event` function to avoid code duplication. I guess you could argue to just make `send_sync_event` exported, but I like having the specific functions for clarity and to keep the events opaque.

src/riak_moss_put_fsm.erl
```diff
@@ -246,6 +274,10 @@ handle_sync_event({augment_data, NewData},
             NextState = StateName
     end,
     {reply, ok, NextState, UpdState};
+handle_sync_event(finalize, From, StateName, State=#state{final_manifest=undefined}) ->
+    {next_state, StateName, State#state{waiter=From}};
+handle_sync_event(finalize, From, StateName, State=#state{final_manifest=M}) ->
```
kellymclaughlin: s/From/_From

src/riak_moss_writer.erl
```diff
@@ -69,7 +70,7 @@ initialize(Pid, FsmPid, Bucket, FileName, ContentLength) ->
                           FileName,
                           ContentLength}).
 
-%% @doc Write a new root block
+%% @doc Write a new root b
```
kellymclaughlin: `block` got truncated for some reason

src/riak_moss_wm_key.erl
```diff
@@ -197,22 +183,25 @@ content_types_accepted(RD, Ctx) ->
 
 -spec accept_body(term(), term()) ->
     {true, term(), term()}.
+accept_body(RD, Ctx=#key_context{bucket=Bucket,key=Key,
+                                 putctype=_CType,size=Size}) ->
+    {ok, Pid} = riak_moss_put_fsm:start_link(list_to_binary(Bucket), Key, Size, <<>>, 60000),
+    accept_streambody(RD, Ctx, Pid, wrq:stream_req_body(RD, ?DEFAULT_LFS_BLOCK_SIZE)).
```
kellymclaughlin, line 189: Use `riak_moss_lfs_util:block_size()` instead of the macro directly here.

kellymclaughlin:
I don't think calculating the content-md5 on the server side is correct (see [this](http://tools.ietf.org/html/rfc1864)). Seems like just more header data we extract and store.

kellymclaughlin:
+1 to merge.

PR #39, opened 2011-12-12
Az999 get fsm backpressure

reiddraper:
I've refactored the GET fsm to have the caller _ask_ for chunks instead of receiving them in the process mailbox. This adds implicit backpressure and makes writing the WM code easier because we don't have to deal with the potential variability in the WM proc pid. The tests have been updated to reflect this. Take a look at the tests to see the new usage.

### DIFF
--- a/src/riak_moss_get_fsm.erl
+++ b/src/riak_moss_get_fsm.erl
@@ -24,7 +24,8 @@
 -export([start_link/3,
          stop/1,
          continue/1,
-         get_metadata/1]).
+         get_metadata/1,
+         get_next_chunk/1]).
 
 %% exported to be used by
 %% spawn_link
@@ -36,8 +37,11 @@
          prepare/2,
          waiting_value/2,
          waiting_metadata_request/3,
-         waiting_chunk_command/2,
+         waiting_chunk_request/3,
+         waiting_continue_or_stop/2,
          waiting_chunks/2,
+         waiting_chunks/3,
+         sending_remaining/3,
          handle_event/3,
          handle_sync_event/4,
          handle_info/3,
@@ -49,6 +53,7 @@
                 key :: term(),
                 value_cache :: binary(), 
                 metadata_cache :: term(), 
+                chunk_queue :: term(),
                 reply_pid :: pid(),
                 manifest :: term(),
                 blocks_left :: list(),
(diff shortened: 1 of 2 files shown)


src/riak_moss_get_fsm.erl
```diff
@@ -46,6 +54,9 @@
                 bucket :: term(),
                 key :: term(),
                 value_cache :: binary(), 
+                metadata_cache :: term(), 
+                chunk_queue :: term(),
+                reply_pid :: pid(),
```
argv0: reply_pid seems to be unused?

src/riak_moss_get_fsm.erl
```diff
@@ -63,24 +74,32 @@ stop(Pid) ->
...
+
 %% ====================================================================
 %% gen_fsm callbacks
 %% ====================================================================
 
-init([From, Bucket, Key]) ->
-    State = #state{from=From, bucket=Bucket, key=Key,
-                   get_module=riak_moss_riakc},
+init([Bucket, Key]) ->
+    Queue = queue:new(),
```
argv0, line 89: not an issue for this PR, but we might want to consider a variant of riak_repl's bounded_queue in the future.

src/riak_moss_get_fsm.erl
```diff
@@ -92,6 +111,10 @@ prepare(timeout, #state{bucket=Bucket, key=Key, get_module=GetModule}=State) ->
     spawn_link(?MODULE, normal_retriever, [self(), GetModule, Bucket, Key]),
```
argv0: we probably want to stash this pid and match on it in handle_info - under heavy load pid reuse can become an issue.

src/riak_moss_get_fsm.erl
```diff
@@ -63,24 +74,32 @@ stop(Pid) ->
 continue(Pid) ->
     gen_fsm:send_event(Pid, continue).
 
+get_metadata(Pid) ->
+    gen_fsm:sync_send_event(Pid, get_metadata).
+
+get_next_chunk(Pid) ->
+    gen_fsm:sync_send_event(Pid, get_next_chunk).
```
argv0, line 82: consider moving riak_moss_put_fsm:send_sync_event to a utility module and using that for get_metadata/get_next_chunk

src/riak_moss_get_fsm.erl
reiddraper: fixed in 2d92fba424a4223995bf630aa864ad9a2ca26c1e

src/riak_moss_get_fsm.erl
reiddraper, line 89: agreed.

src/riak_moss_get_fsm.erl
reiddraper: fixed in fe870137e56ec1e010c39a25a1c63fc047116df2

argv0:
+1 to merge

PR #41, opened 2011-12-12
Az1004 put fsm backpressure

argv0:
A new riak_moss app variable, put_fsm_buffer_size_max, controls how much data the put FSM will buffer.

Once the buffer reaches the max size, calls to augment_data will block until the buffer drains to below the max.

### DIFF
--- a/src/riak_moss_put_fsm.erl
+++ b/src/riak_moss_put_fsm.erl
@@ -48,7 +48,11 @@
                 next_block_id=0 :: non_neg_integer(),
                 raw_data :: undefined | binary(),
                 final_manifest :: undefined | riak_moss_lfs_utils:lfs_manifest(),
-                waiter :: undefined | {reference(), pid()},
+                buffer_size :: non_neg_integer(),
+                max_buffer_size :: non_neg_integer(),
+                block_size :: pos_integer(),
+                block_waiter :: undefined | {reference(), pid()},
+                manifest_waiter :: undefined | {reference(), pid()},
                 timeout :: timeout()}).
 -type state() :: #state{}.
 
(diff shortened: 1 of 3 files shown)


src/riak_moss_put_fsm.erl
```diff
@@ -269,9 +296,14 @@ handle_sync_event({augment_data, NewData},
...
+        true ->
+            {next_state, NextState, UpdState#state{block_waiter=From, buffer_size=NewBufSz}};
+        false ->
+            {reply, ok, NextState, UpdState#state{buffer_size=NewBufSz}}
+    end;
 handle_sync_event(finalize, From, StateName, State=#state{final_manifest=undefined}) ->
-    {next_state, StateName, State#state{waiter=From}};
+    {next_state, StateName, State#state{manifest_waiter=From}};
 handle_sync_event(finalize, _From, StateName, State=#state{final_manifest=M}) ->
     {reply, {ok, M}, StateName, State};
```
kellymclaughlin: This is not directly related to this pr, but would it be better to return `{stop,Reason,Reply,NewStateData}` here? I guess the timeout will eventually initiate the transition from `client_wait` to `stop`, but putting the `stop` here seems more clear and direct.

src/riak_moss_put_fsm.erl
```diff
@@ -177,9 +188,20 @@ write_root(writer_ready, State=#state{writer_pid=WriterPid,
...
     {next_state, write_block, State, Timeout};
 write_root({block_written, BlockId}, State=#state{writer_pid=WriterPid,
+                                                  buffer_size=BufSz,
+                                                  max_buffer_size=MaxBufSz,
+                                                  block_size=BlockSz,
+                                                  block_waiter=From,
                                                   timeout=Timeout}) ->
     riak_moss_writer:update_root(WriterPid, {block_ready, BlockId}),
-    {next_state, write_block, State, Timeout}.
+    NewState = case (NewBufSz = BufSz - BlockSz) >= MaxBufSz of
```
kellymclaughlin: I can't see that this causes any direct harm, but if `BufSz` is < `BlockSz`, `NewBufSz` will be a negative integer and that value could grow somewhat large if there are many transitions between `write_block` and `write_root` before another data chunk is received. Is it worth it to check for that case so we don't get a runaway negative integer there or perhaps I'm over-thinking it.

PR #42, opened 2011-12-13
Set the content type on puts.

kellymclaughlin:
- Use `riakc_obj:new/4` to set the content type when the manifest is
  initially written.
- Update all related functions so the content type is passed from
  webmachine to the put fsm and onto the writer process.
- Update appropriate tests to use a content type parameter.


reiddraper:
+1 to merge

PR #43, opened 2011-12-14
Az997 wm get fsm

reiddraper:
I've hooked up the WM code with the get fsm and also changed to get fsm to work properly with notfounds. I've successfully tested a full put/get cycle with `s3cmd`.

### DIFF
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -13,7 +13,8 @@
                   user :: #moss_user{}}).
 
 -record(key_context, {context :: #context{},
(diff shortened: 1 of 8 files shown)


kellymclaughlin:
The block `notfound` problem I experienced while testing was attributed to a low file descriptor limit, but even so it might be useful to check for an error response from the riak `get` in `riak_moss_get_fsm:blocks_retriever` and log a message indicating the problem block before throwing an error.

kellymclaughlin:
+1 to merge

PR #44, opened 2011-12-15
Riak connection work

kellymclaughlin:
- No longer use single shared riak connection through `riak_moss_riakc`.
- Riak connections are established and closed as needed, but a single connection is reused for getting and putting large files.
- Functions from  `riak_moss_riakc` and `riak_moss` have been merged into `riak_moss_utils`.
- `riak_moss_object` has been removed as it is unused.
- Test suites have been updated where appropriate.


reiddraper:
+1 to merge

PR #47, opened 2011-12-19
Handle object and file deletion

kellymclaughlin:
- New `riak_moss_delete_fsm` module coordinates deletion.
- `riak_moss_deleter` module performs actual deletion from `riak`. 
- `riak_moss_delete_fsm_eqc` module tests state transitions of the delete fsm.
- `riak_moss_deleter_test` tests for expected responses from the deleter module for a given input message.
### To Test
1. Create a moss bucket:
   `s3cmd mb s3://testbucket`
2. Put a file into the bucket:
   `s3cmd put <FILE_NAME_HERE> s3://testbucket/`
3. Verify the contents of the Riak buckets using erlang console for **Riak**:
   <pre>
   <code>
   {ok, C}=riak:local_client().
   C:list_keys(<<"objects:testbucket">>).
   C:list_keys(<<"blocks:testbucket">>).
   </code>
   </pre>
4. Delete the file on the **MOSS** console:
   `riak_moss_delete_fsm_sup:start_delete_fsm(node(), [<<"testbucket">>, <<"FILE_NAME_HERE">>, 300000]).`
5. Re-run the 2 `list_keys` calls from step 3 on the **Riak** console and each should now return empty lists (assuming there were no pre-existing files in the bucket).


src/riak_moss_delete_fsm.erl
```diff
@@ -0,0 +1,266 @@
...
+                                 {next_state,
+                                  waiting_blocks_delete,
+                                  state(),
+                                  timeout()}.
+waiting_root_update(root_inactive, State=#state{block_count=BlockCount,
+                                                deleter_pid=DeleterPid,
+                                                timeout=Timeout}) ->
+
+    %% Send request to the deleter to delete the file blocks
+    %% @TODO Backpressure or concurrency of deletion needed. Currently
```
reiddraper, line 162: I think this is OK for now, but as you noted, we could use some back-pressure. I could see someone doing deletes in a loop and having a bunch of procs with mailboxes of hundreds or thousands of messages. I think this will sort of fix itself when we add concurrency.

src/riak_moss_delete_fsm.erl
```diff
@@ -0,0 +1,266 @@
...
+                 string(),
+                 timeout()) ->
+                        {ok, pid()} | ignore | {error, term()}.
+start_link(Bucket, Name, Timeout) ->
+    Args = [Bucket, Name, Timeout],
+    gen_fsm:start_link(?MODULE, Args, []).
+
+%% @doc Send an event to a `riak_moss_delete_fsm'.
+-spec send_event(pid(), term()) -> ok.
+send_event(Pid, Event) ->
```
reiddraper, line 63: Not worth changing now, but I think it'd be nice to encapsulate all of the messages events the deleter proc sends into their own funcs

reiddraper:
+1

kuenishi:
https://github.com/basho/riak_cs/blob/master/client_tests/python/boto_test.py#L120
Is this test can be uncommented by this fix?

PR #48, opened 2011-12-19
Az998 wm deletes

reiddraper:
Testing can be done like:

```
s3cmd del s3://bucket/key
```
## known issues
- currently returns 204 instead of 404 when the BKey doesn't exist


kellymclaughlin:
+1 to merge

PR #54, opened 2012-01-09
Changes so that s3cmd subcommands sync, du, and rb work.

kellymclaughlin:
Some of these changes are pragmatic and we will want to evaluate the possibility of improving efficiency in the future. Getting the metadata for object size and checksum is particularly expensive, but should suffice for the near-term. Also changes will be required when we begin enforcing globally-unique bucket names. I have tried to notate places that could be affected with `TODO` statements.


reiddraper:
I've successfully used `s3cmd du`, `s3cmd sync`, and `s3cmd rb` (`rb` properly returns conflict when the bucket isn't deleted. +1 to merge

PR #55, opened 2012-01-09
make buckets truly private by default

argv0:
add a check in riak_moss_s3_auth:authenticate to match the request bucket to the list of buckets owned by the user.  exceptions:
- requests to the service-level resource (no bucket to validate)
- requests to create buckets


kellymclaughlin:
Looks good. +1 to merge.

PR #56, opened 2012-01-11
URL decode keys on put so they are represented correctly.

kellymclaughlin:
This eliminates confusion when objects with spaces in their names are listed and when attempting to access them.


reiddraper:
Tested and works for me. +1 to merge

PR #58, opened 2012-01-13
support the prefix query parameter

argv0:
this is needed by the blackflag loader script.  from the python repl using boto:

```>>> b.get_all_keys()
[<Key: baz,acbd18db4cc2f85cedef654fccc4a4d8>, <Key: baz,p1>, <Key: baz,p2>, <Key: baz,p3>]

> > > b.get_all_keys(prefix="p")
> > > [<Key: baz,p1>, <Key: baz,p2>, <Key: baz,p3>]
> > > b.get_all_keys(prefix="")
> > > [<Key: baz,acbd18db4cc2f85cedef654fccc4a4d8>, <Key: baz,p1>, <Key: baz,p2>, <Key: baz,p3>]```

### DIFF
--- a/src/riak_moss_utils.erl
+++ b/src/riak_moss_utils.erl
@@ -17,7 +17,7 @@
          delete_object/2,
          from_bucket_name/1,
          get_buckets/1,
-         get_keys_and_objects/1,
+         get_keys_and_objects/2,
          get_object/2,
          get_object/3,
          get_user/1,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/riak_moss_utils.erl [hunk 1]
```diff
@@ -175,15 +175,15 @@ get_buckets(#moss_user{buckets=Buckets}) ->
...
+-spec get_keys_and_objects(binary(), binary()) -> {ok, [binary()]}.
+get_keys_and_objects(BucketName, Prefix) ->
     case riak_connection() of
         {ok, RiakPid} ->
             case list_keys(BucketName, RiakPid) of
                 {ok, Keys} ->
                     KeyObjPairs =
                         [{Key, get_object(BucketName, Key, RiakPid)}
-                         || Key <- Keys],
+                         || Key <- prefix_filter(Keys, Prefix)],
```
kellymclaughlin, line 186: Might be better to check if a prefix is specified and only do the prefix filtering if necessary since it's unnecessary expense if the key list is large.

kellymclaughlin:
+1 to merge

PR #59, opened 2012-01-18
fixes needed for customer script

argv0:
fixes for :
- connections not being reaped
- incomplete keys appearing in bucket listing
- zero byte file crash


kellymclaughlin:
+1 to merge

PR #60, opened 2012-01-24
S71 get fsm eqc

reiddraper:
This branch has a few more changes than I would've liked, but I guess that's what a week of QuickCheck does. The biggest changes this branch has are:
- refactor the GET FSM to use a "reader" gen_server instead of spawning procs itself
- create a test reader that doesn't connect to Riak
- clean up process dependencies so that processes get shutdown nicely when they're deps or creator go away
- add an EQC test for the GET FSM. This test has some known limitations, namely that the number of commands generated doesn't always drive the GET FSM through all of its phases. ie, it doesn't ask for the maximum number of blocks. I intend to fix this in the future by changing the test to be a "regular" EQC test instead of an `eqc_fsm` test.
## Tests
- all of these tests with `make test` should pass
- `prototype_parity.py` tests should also pass

### DIFF
--- a/src/riak_moss_reader.erl
+++ b/src/riak_moss_reader.erl
@@ -1,0 +1,131 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2007-2011 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc Module to write data to Riak.
+
+-module(riak_moss_reader).
+
+-behaviour(gen_server).
+
+-include("riak_moss.hrl").
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+
+%% Test API
+-export([test_link/0]).
+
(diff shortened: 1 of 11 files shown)


src/riak_moss_reader.erl
```diff
@@ -0,0 +1,138 @@
...
+         init/2,
+         handle_call/3,
+         handle_cast/2,
+         handle_info/2,
+         terminate/2,
+         code_change/3]).
+
+-record(state, {riakc_pid :: pid(),
+                caller_pid :: pid(),
+                monitor_ref :: reference()}). %% TODO: is ref() the right type here?
```
kellymclaughlin: `reference()` is the correct type here if you want to get rid of this `TODO`

reiddraper: removed in b037f151fd71f0eda9779e42065a99317edbbe1c

kellymclaughlin:
+1 to merge.

PR #61, opened 2012-01-24
clean up riak pids from deleter

argv0:
(sprintly ticket 79)

minimal changes to make sure that many deletes don't exhaust connections.


reiddraper:
For the GET FSM I went the route of also adding a `process_flag(trap_exit, true)` call and handling 'EXIT' calls in handle_info. I also terminated the gen_server through the supervisor, so it doesn't complain. That being said, I think this approach is OK too. +1

PR #62, opened 2012-01-26
Fix standalone test run failure of riak_moss_s3_auth tests.

kellymclaughlin:
- Move all test cases under a single test suite.
- Set `moss_root_host` environment variable for the tests needed by the
  `bucket_from_host` function.
- These changes also make it provide finer-grained test case output when running `rebar eunit apps=riak_moss suite=riak_moss_s3_auth -v`.
### Testing

Running `rebar eunit apps=riak_moss suite=riak_moss_s3_auth -v` on _master_ should fail and running again with _s80-s3auth-test-fix_ should pass.

### DIFF
--- a/src/riak_moss_s3_auth.erl
+++ b/src/riak_moss_s3_auth.erl
@@ -14,17 +14,15 @@
 -include_lib("eunit/include/eunit.hrl").
 -endif.
 
-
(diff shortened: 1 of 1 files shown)


reiddraper:
+1 to merge

PR #63, opened 2012-01-28
Use file creation time instead of `now` for last-modified time when listing files.

kellymclaughlin:
- Use file creation time as value for last-modified element in object
  listing responses.
- Use ISO 8601 time format in creation field of file manifest records.

### DIFF
--- a/src/riak_moss_lfs_utils.erl
+++ b/src/riak_moss_lfs_utils.erl
@@ -41,6 +41,7 @@
          still_waiting/1,
          content_md5/1,
          content_length/1,
+         created/1,
          is_active/1]).
 
 -export_type([lfs_manifest/0]).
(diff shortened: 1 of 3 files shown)


reiddraper:
This change isn't backward compatible with manifests stored before the change. The result is `s3cmd` not being able to parse the date format, with an exception like:

```
ValueError: time data 'Mon, 30 Jan 2012 14:51:53 GMT' does not match format '%Y-%m-%dT%H:%M:%S.000Z'
```

Looks like we need some way to change an "old" `httpd_util:rfc1123_date().` format time into a `riak_moss_wm_utils:iso_8601_datetime().` format time. Or we could have some clause that says if we don't understand the format, set it to `now()`?

kellymclaughlin:
I suppose we will have to do something about that. _grumble, mutter, grumble, grumble_

reiddraper:
Awesome, tested and works for me. +1 to merge

PR #64, opened 2012-02-01
Fix handling of HEAD requests when the object does not exist.

kellymclaughlin:
`HEAD` requests for non-existent objects caused a 500 error. This change fixes that so that a proper 404 error is returned instead. This change affects the usage the `s3cmd` subcommand `info`.


slfritchie:
Yup, now returning HTTP status code 404.  +1 to merge, thanks!

PR #66, opened 2012-02-02
Include Lookup Errors in Access Usage Query (S94)

beerriot:
This branch adds errors encountered in archive lookup to the access usage response.  The idea is to let people know if the archive lookup failed, instead of being obtuse about whether or not there is any information for a slice.

See also documentation updates in the wiki branch `s94-lookup-errors`.

### DIFF
--- a/src/riak_moss_wm_usage.erl
+++ b/src/riak_moss_wm_usage.erl
@@ -129,15 +129,25 @@ produce_json(RD, Ctx) ->
 
 mochijson_access(Msg) when is_atom(Msg) ->
     Msg;
-mochijson_access(Access) ->
+mochijson_access({Access, Errors}) ->
     orddict:fold(
       fun(Node, Samples, Acc) ->
               [{struct, [{node, Node},
                          {samples, [{struct, S} || S <- Samples]}]}
                |Acc]
       end,
-      [],
+      [{struct, [{errors, [ {struct, mochijson_access_error(E)}
+                            || E <- Errors ]}]}],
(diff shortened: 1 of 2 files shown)


src/riak_moss_wm_usage.erl
```diff
@@ -154,16 +164,28 @@ produce_xml(RD, Ctx) ->
...
                 [xml_access_sample(S) || S <- Samples]}
                |Acc]
       end,
-      [],
+      [{'Errors', [ xml_access_error(E) || E <- Errors ]}],
       Access).
 
+xml_access_error({{Start, End}, Reason}) ->
+    %% cheat to make errors structured exactly like samples
+    FakeSample = [{<<"start_time">>, riak_moss_access:iso8601(Start)},
```
kellymclaughlin, line 179: [`riak_moss_wm_utils`](https://github.com/basho/riak_moss/blob/master/src/riak_moss_wm_utils.erl) on master has some iso8601 formatting functions as well. After the overarching story for access logs is merged to master perhaps we can reconcile and minimize the code related to this.

kellymclaughlin:
+1

src/riak_moss_wm_usage.erl
beerriot, line 179: Agreed.  I will handle that before the final feature merge.  I'll be there's even room for a `riak_moss_fmt` to take over parts of the generic "utils".  Tracking here: https://sprint.ly/product/969/item/110

PR #67, opened 2012-02-07
Fix bucket creation when a content-type header with parameters is used.

kellymclaughlin:
The webmachine resource for handling bucket creation requests was not separating the media type from the parameters that were included in the _content-type_ header. This was causing the media type to be unrecognized and resulting in `415 Unsupported Media Type` errors being returned.

### DIFF
--- a/src/riak_moss_wm_bucket.erl
+++ b/src/riak_moss_wm_bucket.erl
@@ -80,9 +80,9 @@ content_types_accepted(RD, Ctx) ->
         undefined ->
             {[{"application/octet-stream", accept_body}], RD, Ctx};
         CType ->
-            {[{CType, accept_body}], RD, Ctx}
+            {Media, _Params} = mochiweb_util:parse_header(CType),
+            {[{Media, accept_body}], RD, Ctx}
     end.
-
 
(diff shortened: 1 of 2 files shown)


src/riak_moss_wm_bucket.erl
```diff
@@ -80,10 +80,10 @@ content_types_accepted(RD, Ctx) ->
         undefined ->
             {[{"application/octet-stream", accept_body}], RD, Ctx};
         CType ->
-            {[{CType, accept_body}], RD, Ctx}
+            Media = hd(string:tokens(CType, ";")),
```
beerriot: I'd suggest using `mochiweb_util:parse_header/1` instead of `string:tokens`, since it also handles whitespace stripping and such.

kellymclaughlin:
Great idea. Changed that in both the buckets and keys resource modules.

slfritchie:
+1 to merge.  Please use the test script at
https://gist.github.com/a22d5ab4215de9e109e0

... to test.  Note that the script will fail if the Webmachine PR #55 isn't
also applied.

PR #68, opened 2012-02-07
Add ability to manually trigger the access log archival (S96)

beerriot:
When a MOSS node is being taken out of service, a user will likely want to ensure that all of its access logs are flushed to Riak before turning it off.  This PR adds `riak_moss flush_access` to do just that.

This PR also includes code to check & set `allow_mult=true` on the archive bucket at MOSS startup time, to ensure that multiple writes to a user-slice record don't clobber each other (which could happen by flushing multiple times during the same slice).

### DIFF
--- a/rel/files/riak_moss
+++ b/rel/files/riak_moss
@@ -251,19 +251,8 @@ case "$1" in
         fi
         echo "config is OK"
         ;;
-    flush_access)
-        # Make sure the local node IS running
-        RES=`$NODETOOL ping`
-        if [ "$RES" != "pong" ]; then
-            echo "Node is not running!"
-            exit 1
-        fi
-        shift
-
-        $NODETOOL rpc riak_moss_console flush_access $@
-        ;;
     *)
-        echo "Usage: $SCRIPT {start|stop|restart|reboot|ping|console|attach|chkconfig|flush_access}"
+        echo "Usage: $SCRIPT {start|stop|restart|reboot|ping|console|attach|chkconfig}"
         exit 1
         ;;
 esac
--- a/rel/files/riak_moss
+++ b/rel/files/riak_moss
@@ -251,8 +251,19 @@ case "$1" in
         fi
         echo "config is OK"
         ;;
+    flush_access)
+        # Make sure the local node IS running
+        RES=`$NODETOOL ping`
+        if [ "$RES" != "pong" ]; then
+            echo "Node is not running!"
+            exit 1
+        fi
+        shift
+
+        $NODETOOL rpc riak_moss_console flush_access $@
+        ;;
     *)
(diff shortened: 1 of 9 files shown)


rel/files/riak_moss
```diff
@@ -251,8 +251,19 @@ case "$1" in
...
+            echo "Node is not running!"
+            exit 1
+        fi
+        shift
+
+        $NODETOOL rpc riak_moss_console flush_access $@
+        ;;
     *)
-        echo "Usage: $SCRIPT {start|stop|restart|reboot|ping|console|attach|chkconfig}"
+        echo "Usage: $SCRIPT {start|stop|restart|reboot|ping|console|attach|chkconfig|flush_access}"
```
kellymclaughlin: Ran into problems trying to test the new option due to something totally outside the scope of this change, but wanted to mention it here anyway just to record it. Since we're using `node_package` to build, the `riak_moss` script in rel/files never actually gets touched. So if I build a package the `/sbin/riak_moss` script is actually the `runner` script from `node_package` after it's run through the template engine. I've talked to Jared about some options to be able to use node_package script as a base, but add customized options, but all would require work that is not done. For right now two options I see are to copy the runner script from node_package to the moss repo and customize it or create an `moss-admin` script for the custom options. Thoughts?

beerriot: Doh.  You caught me.  I never tested that `make rel` moved this over, and just assumed that it worked like it always had in the past.  :P

Both of the alternatives suck, though.  There has been lots of complaining about how the runner scripts in `riak`, `riak_ee`, and now `riak_moss` have been maintained at unpredictable versions.  Similarly, there's constant grouching about how the division of `riak` and `riak-admin` makes no sense.

I think I'd be tempted to choose a `moss-admin`-like path if pressed to do so.  I might suggest a name more like `moss-usage`, though, to make it obvious that the commands there relate to usage records, and not controlling MOSS.  There may be additional commands to add anyway (like `status` to see what the logger/archiver is up to, `lookup` to avoid the HTTP frontend, or `force_storage` to do a storage-calculation run outside of the schedule).  What do you think?

kellymclaughlin: Yeah I dislike both too and that's one big reason for the existence of `node_package` I think. I'm with you, a `usage` script might be the best short-term path. It'd be great if we could figure a better solution before the 1.0 release so we don't have to retrain all the users if we get the node_package situation figured out. I made task [#109](https://sprint.ly/product/969/item/109) to keep track of this.

beerriot:
I've moved the `flush_access` command to a new `moss_usage` script, as discussed. I've also merged this branch and others into `s6-access-log`.  I plan to submit a mega-pull-request when the whole access+storage feature is ready for mainline.

PR #71, opened 2012-02-11
Change user record structure and lay ground work for ACLs

kellymclaughlin:
This set of changes updates the user record structure to include _email_, _display_name_, and _canonical_id_ fields. These fields are necessary to mirror the information used by amazon for S3 and to implement ACLs. The _buckets_ field is removed as all bucket information will be stored in the `<<moss.buckets>>` bucket instead of with each users's data. The changes are also a prerequisite for the serialization of user creation requests which is required to enforce the uniqueness of email addresses in the system. 

Also included is `riak_moss_acl_utils`, a utility module for ACL handling. A default ACL is created when a bucket is created in `riak_moss_wm_bucket`, but it's currently just a placeholder. The `create_bucket` function in `riak_moss_utils` does not currently do anything with it. This is just more groundwork for more ACL work that will be done in the near future.

Creating a user is still accomplished via a POST to the `/user` resource, but in addition to specifying a name, a properly formatted email address is also required. 

A curl request to create a user would look like this:

<pre><code>
curl http://localhost:8080/user --data "email=joe@bob.com&name=joe%20bob"
</code></pre>

The response JSON would look similar to the following:

<pre><code>
{
    "email": "joe@bob.com",
    "display_name": "joe",
    "key_id": "324ABC0713CD0B420EFC086821BFAE7ED81442C", 
    "key_secret": "5BE84D7EEA1AEEAACF070A1982DDA74DA0AA5DA7", 
    "name": "joe bob",
    "id":"8d6f05190095117120d4449484f5d87691aa03801cc4914411ab432e6ee0fd6b"
}
</code></pre>

We may need a script to assist existing users in migrating to this new user representation, but that has been left as a separate piece of work.

### DIFF
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -1,6 +1,8 @@
 
 -record(moss_user, {
           name :: string(),
+          display_name :: string(),
+          email :: string(),
           key_id :: string(),
           key_secret :: string(),
           canonical_id :: string(),
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -1,17 +1,23 @@
-
 -record(moss_user, {
+          name :: string(),
+          key_id :: string(),
+          key_secret :: string(),
+          buckets = []}).
+
+-record(moss_user_v1, {
           name :: string(),
           display_name :: string(),
           email :: string(),
           key_id :: string(),
           key_secret :: string(),
-          canonical_id :: string(),
-          buckets = []}).
--type moss_user() :: #moss_user{}.
+          canonical_id :: string()}).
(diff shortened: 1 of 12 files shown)


beerriot:
Removing the `buckets` field from the user record will make it significantly more work to calculate any single user's total storage space.  Producing a list of the user's buckets will require either a MapReduce or 2i query (if there is an "owner" index) across `<<moss.buckets>>`, instead of simply fetching the user's record.

kellymclaughlin:
After discussion it has been decided that the _buckets_ field will not be removed at this time. A commit is forthcoming to reflect that fact.

PR #75, opened 2012-02-18
`s3cmd info` fails due to KeyError

slfritchie:
Fixes #83

Add last-modified time to the dictionary that's stored in the
riak_moss_get_fsm's #state.metadata_cache.  Do some minor refactoring
to use lists:foldl/3 instead of explicit dict:store/3 and
wrq:set_resp_header/3 calls.

I also added riak_moss_wm_util:iso_8601_to_rfc_1123/1.  It's
a bit brute force, but I didn't see any utility functions lying
around to help.

Verification script: https://gist.github.com/07de3132a16ba5457100


kellymclaughlin:
+1 to merge.

PR #76, opened 2012-02-18
"s3cmd info" fetches entire object

slfritchie:
Fixes #119

Added HTTP method to #key_context record to allow starting
of the riak_moss_get_fsm so that that FSM can avoid fetching
all of the object's chunks.

### DIFF
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -20,6 +20,7 @@
                       putctype :: string(),
                       bucket :: list(),
                       key :: list(),
+                      method :: atom(),
                       size :: non_neg_integer()}).
 
 -define(USER_BUCKET, <<"moss.users">>).
--- a/src/riak_moss_get_fsm.erl
+++ b/src/riak_moss_get_fsm.erl
@@ -21,7 +21,7 @@
 -include("riak_moss.hrl").
 
 %% API
--export([start_link/2,
+-export([start_link/3,
          stop/1,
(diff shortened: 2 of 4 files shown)


reiddraper:
The GET fsm already supports this. The fsm will only retrieve the manifest by default, and then you have to explicitly call `riak_moss_get_fsm:continue/`` for it to start retrieving the individual chunks. I was thinking we could just have [this](https://github.com/basho/riak_moss/blob/62d7e494f8aca688dc83dd4aca4119e3f6dae960/src/riak_moss_wm_key.erl#L148-154) part of the code only happen on GET requests, and stop the FSM on HEAD requests. Lemme know if this makes sense.

reiddraper:
I talked with @beerriot and it turns out we're going to want to make a change to webmachine (probably just our own branch for now) to allow us to fulfill HEAD requests without producing the entire body in memory.

beerriot:
If I'm reading (webmachine_request)[https://github.com/basho/webmachine/blob/wrd-content-length-with-streaming/src/webmachine_request.erl#L349-350] correctly, I think @reiddraper's earlier `known_content_length` change should already be enough on the Webmachine side.  Can you just move the `riak_moss_get_fsm:continue/1` call inside the anonymous function, so it's not called for HEAD?

reiddraper:
Another approach was taken [here](https://github.com/basho/riak_moss/pull/94)

PR #77, opened 2012-02-22
Bucket-level ACLs

kellymclaughlin:
This pull request encompasses the changes needed to support bucket-level ACLs similar to S3's bucket-level ACLs. An overview of those can be found [here](http://docs.amazonwebservices.com/AmazonS3/latest/dev/ACLOverview.html).

The following are now true:
1. Buckets are truly private by default. _Note_ The use of the `x-amz-acl` header to specify a non-default ACL at bucket creation time is not yet supported, but a `TODO` placeholder is in the code.
2. Bucket ACLs may be updated via a `PUT` to the `?acl' *subresource* of a bucket. This corresponds to the`setacl`subcommand of`s3cmd`.
3. Bucket ACLs may be retrieved by a `GET` to the `?acl' *subresource* of a bucket. This corresponds to part of the information returned by the`info`subcommand of`s3cmd`.
4. ACL permissions are granted using a user's email address or the canonical id created at user creation time.
5. Access grants to _everyone_ or _all authenticated users_ should work, but still needs to be thoroughly tested.

These changes also require `stanchion` branch `s44-bucket-acl-update`. A pull requests for those changes is [here](https://github.com/basho/stanchion/pull/3).

An automated test to exercise the ACL behavior will hopefully be forthcoming shortly, but is unfortunately not done yet. 

Just for reference here is an example s3cmd to grant read permissions to a user on `testbucket1`:

```
s3cmd setacl s3://testbucket1 --acl-grant=read:b8becb0e04990dcbb15c946f6e394994c5b95da160f2c1508a333faec5f36b37
```


beerriot:
"I haven't run the code yet, but..."

It looks like there's some confusion about the return types of `riak_moss_acl:object_access` and `:anonymous_object_access`. The specs for each say they return `{true, string()}` along one of their happy paths.  I think that's not true for `anonymous_object_access`, though, since it ends in a call to `has_permission`, which only returns `boolean()`.  At the same time, it looks like the case in `riak_moss_wm_key:forbidden` is missing a match for `{true, string()}` returned from `object_access`.

FWIW, the access logging code is _always_ going to be interested in who the owner of the key is, if authentication+permissions succeed, so `{true, KeyID}` or some other clear indication of what owner the bucket lookup found would be preferred.

reiddraper:
Moss returns 503 for an authorized user when you start moss/stanchion/riak in quick succession.

kellymclaughlin:
@beerriot In the wm resource modules for buckets and keys, the owner of the entity will end up set in the `user` field of the context record, but it seems like you might want something more like a `get_owner` function exported from one of the acl modules. Is that correct?

beerriot:
@kellymclaughlin I only see the `user` field being set if authentication was included, but not on anonymous activity.  It also looks like `user` is set to the actor for the request, not the owner of the key.  I _think_ I'm going to want the owner, and since you'll find that in the object you're looking up to check permissions, it seems like a good idea to save a second roundtrip to Riak.

kellymclaughlin:
@reiddraper 403 issue resolved with [f4276c8](https://github.com/basho/riak_moss/commit/f4276c8b0c36a8af94c57eb590831209996a99c0).

src/riak_moss_wm_bucket.erl
```diff
@@ -40,17 +40,82 @@ malformed_request(RD, Ctx) ->
 %%      we'd use the `authorized` callback,
 %%      but this is how S3 does things.
 forbidden(RD, Ctx=#context{auth_bypass=AuthBypass}) ->
```
reiddraper, line 42: It would be nice to break this function up, it's getting a bit hard to follow at ~75 lines w/ lots of case-statement nesting.

src/riak_moss_wm_key.erl
```diff
@@ -55,20 +55,59 @@ service_available(RD, Ctx) ->
 %%      authenticated. Normally with HTTP
 %%      we'd use the `authorized` callback,
 %%      but this is how S3 does things.
-forbidden(RD, Ctx=#key_context{context=#context{auth_bypass=AuthBypass}}) ->
+forbidden(RD, Ctx=#key_context{bucket=Bucket,
```
reiddraper, line 58: this func might also benefit from being split-up into smaller components

src/riak_moss_acl_utils.erl
```diff
@@ -69,13 +71,122 @@ acl_to_xml(Acl) ->
...
            {'AccessControlList', grants_xml(Acl?ACL.grants)}
           ]}],
-    unicode:characters_to_list(
+    unicode:characters_to_binary(
       xmerl:export_simple(XmlDoc, xmerl_xml, [{prolog, ?XML_PROLOG}])).
 
+%% @doc Convert an internal representation of an ACL into
+%% erlang terms that can be encoded using `mochijson2:encode'.
+-spec acl_to_json_term(acl_v1()) -> term().
+acl_to_json_term(#acl_v1{owner={DisplayName, CanonicalId},
```
reiddraper: can you use the `stanchion:acl_to_json_term/1` func instead of redefining it?

reiddraper:
2654f7e5ddac4a00a882a7ad43f3b91943202f08 addresses the duplicated funcs from stanchion.

src/riak_moss_wm_key.erl
kellymclaughlin, line 58: I think I will defer this until after the object ACL work. I'll handle the refactoring of the `forbidden` function in `riak_moss_wm_bucket` at that time as well.

reiddraper:
+1 to merge

PR #78, opened 2012-02-22
Add url-safe base64 encoding module and change all base64 encoding to use it.

kellymclaughlin:
Since there are occasions where the user key ids will be used in URLs, let's make them URL-safe.


reiddraper:
+1 to merge

PR #79, opened 2012-02-24
Change runner script path from bin to sbin so riak_moss start will work.

kellymclaughlin:
The runner script was looking in `bin` instead of `sbin` for the `riak_moss` script so moss would not start.


jaredmorrow:
+1

PR #80, opened 2012-02-25
Server-side signature calculation can fail when using the base64url module  and the key secret contains characters that are not url-safe.

kellymclaughlin:
Just use the standard base64 module for signature calculation. It works for existing credentials as well as credentials generated with the base64url module.


reiddraper:
+1 to merge

PR #82, opened 2012-02-25
Server-side signature calculation can fail when using the base64url modu...

jaredmorrow:
Just use the standard base64 module for signature calculation. It works for existing credentials as well as credentials generated with the base64url module.


jaredmorrow:
Copying https://github.com/basho/riak_moss/pull/80 to the 0.1 branch rather than master.

PR #84, opened 2012-02-26
Changes to properly handle anonymous bucket access.

kellymclaughlin:
1. Change atom used for `key_id` when no auth header is specified in
   riak_moss_wm_utils:parse_auth_header.
2. Fix match condition used in riak_moss_acl:has_permission/2. It should look for a match for the `'AllUsers'` atom instead of the `AllUsers` URI.
## Testing
1. Create a bucket using an existing account: `s3cmd -c ~/.mosscfg mb s3://testbucket`.
2. Set the bucket to have public read access: `s3cmd -c ~/.mosscfg setacl -P s3://testbucket`.
3. List the bucket info: `curl http://localhost:8080/testbucket` and you should see something similar to the following output:

```
<ListBucketResult>
  <Name>testbucket</Name>
  <Prefix/>
  <Marker/>
  <MaxKeys>1000</MaxKeys>
  <Delimiter>/</Delimiter>
  <IsTruncated>false</IsTruncated>
</ListBucketResult>
```

Using `master` step 3 would result in a `500 Internal Server Error`.

### DIFF
--- a/src/riak_moss_acl.erl
+++ b/src/riak_moss_acl.erl
@@ -42,7 +42,12 @@ anonymous_bucket_access(Bucket, RequestedAccess) ->
                     %% Only owners may delete buckets
                     false;
                 _ ->
-                    has_permission(Acl, RequestedAccess)
+                    case has_permission(Acl, RequestedAccess) of
+                        true ->
+                            {true, owner_id(Acl)};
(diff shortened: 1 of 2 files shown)


reiddraper:
+1 to merge

PR #85, opened 2012-02-27
Changes to make maximum content length configurable.

kellymclaughlin:
- Create `max_content_len/0` function in `riak_moss_lfs_utils`.
- Change `MAX_CONTENT_LENGTH` macro to `DEFAULT_MAX_CONTENT_LENGTH`.
- Allow non-default maximum content length to be specified via
  `max_content_length` setting in the `app.config` file.


reiddraper:
+1 to merge

PR #90, opened 2012-03-02
Change no_such_user atom in riak_moss_wm_service:forbidden to no_user_key

kellymclaughlin:
Unauthorized requests to the service resource (_i.e._ `GET /`) currently return a `503` error code when the more appropriate (imo) error response is `403`. This is due to an incorrect atom used to check responses from a call to `riak_moss_utils:get_user`. 
## Testing

Start riak, stanchion, and moss. On `master` of moss run `curl localhost:8080/ -v`. The response should be a `503`. Now checkout the `s158-fix-wm-service-user-atom` branch and rerun the curl query. The response this time should be `403`.

### DIFF
--- a/src/riak_moss_wm_bucket.erl
+++ b/src/riak_moss_wm_bucket.erl
@@ -113,9 +113,13 @@ forbidden(RD, Ctx=#context{auth_bypass=AuthBypass}) ->
                     %% Anonymous access not allowed, deny access
                     riak_moss_s3_response:api_error(access_denied, RD, Ctx)
             end;
+        {error, notfound} ->
+            %% Access not allowed, deny access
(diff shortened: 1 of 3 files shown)


reiddraper:
This works for anonymous requests, but I still get 503 using signed requests with bogus creds. The moss logs look like:

```
Retrieval of user record for "sdf" failed. Reason: notfound
```

kellymclaughlin:
`503` for bogus creds addressed by [3832217](https://github.com/basho/riak_moss/commit/38322178fab8bb878f1768050356ec82e858c412). Now requests with bogus creds at the service, bucket, or key level should result in a `403`.

reiddraper:
woot, +1 to merge

PR #91, opened 2012-03-02
Remove obsolete nodetool and riak_moss scripts.

kellymclaughlin:
`node_package` brings in these scripts for us now.


jaredmorrow:
+1

PR #95, opened 2012-03-06
Fix dependencies to match those of 0.1.0

jaredmorrow:
0.1.1 had its dependencies get squashed by a mistaken merge somewhere.   This is setting the dependencies back to those of 0.1.0.


jaredmorrow:
Merging for 0.1.2 release.

PR #96, opened 2012-03-08
Allow for non-default canned ACL to be set at bucket creation time. 

kellymclaughlin:
- Add `canned_acl/3` function to `riak_moss_acl_utils`. This function will be reused for objects as well which is the reason for the 3rd parameter.
- Unit test for `canned_acl/3`.
- Change bucket resource to use `canned_acl` instead of `default_acl` so that the value of `x-amz-acl` header is not ignored.
- Update the `riak_moss_passthru_auth` module. `s3cmd` only supports the `public-read` option when creating a bucket so the easiest way to test the changes is to set `auth_bypass` to `true` in the app.config file and use curl to issue requests.
## Testing

The supported values for the `x-amz-acl` are: _private_, _public-read_, _public-read-write_, _authenticated-read_, 
_bucket-owner-read_, and _bucket-owner-full-control_. There is also _log-delivery-write_, but we're not supporting that initially. Each of these maps to a canned ACL. Descriptions are [here](http://docs.amazonwebservices.com/AmazonS3/latest/dev/CannedACL.html).
- Create a bucket with _public-read_ with s3cmd and then verify the ACL is set properly: `s3cmd mb --acl-public s3://somebucket` and then `s3cmd info s3://somebucket` to fetch the ACL. The ACL should grant your user full control and the public group (_http://acs.amazonaws.com/groups/global/AllUsers_) read access.
- Stop moss, set `auth_bypass` to `true`, and restart moss.
- Use curl to create buckets with `x-amz-acl` set to each of _public-read-write_, _authenticated-read_, 
  _bucket-owner-read_, and _bucket-owner-full-control_ and then fetch the ACL for each and verify the expected permissions are granted. Note that _bucket-owner-read_ and _bucket-owner-full-control_ are treated the same as _private_ for buckets. 

Create a bucket: 
`curl localhost:8080/abucket -X PUT -H "authorization: <key_id_here>" -H "x-amz-acl: authenticated-read"`
Fetch the ACL: 
`curl localhost:8080/abucket?acl -H "authorization:  <key_id_here>"`

### DIFF
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -60,7 +60,7 @@
 -define(USER_BUCKET, <<"moss.users">>).
 -define(BUCKETS_BUCKET, <<"moss.buckets">>).
 -define(FREE_BUCKET_MARKER, <<"0">>).
--define(MAX_CONTENT_LENGTH, 5368709120). %% 5 GB
+-define(DEFAULT_MAX_CONTENT_LENGTH, 5368709120). %% 5 GB
 -define(DEFAULT_LFS_BLOCK_SIZE, 1048576).%% 1 MB
 -define(XML_PROLOG, "<?xml version=\"1.0\" encoding=\"UTF-8\"?>").
 -define(DEFAULT_STANCHION_IP, "127.0.0.1").
--- a/src/riak_moss_acl_utils.erl
+++ b/src/riak_moss_acl_utils.erl
@@ -20,6 +20,7 @@
 %% Public API
 -export([acl/3,
          default_acl/2,
+         canned_acl/3,
          acl_from_xml/1,
          acl_to_xml/1,
          empty_acl_xml/0,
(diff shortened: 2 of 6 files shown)


reiddraper:
I think there's an issue with "public-read-write" (and maybe "public-read"):

```
# turn auth_bypass on
curl localhost:8080/bbucket -X PUT -H "authorization: V1O0FH5GRERBDPLYMDXE" -H "x-amz-acl: public-read-write"
# success
curl localhost:8080/bbucket

#returns
<?xml version="1.0" encoding="UTF-8"?><Error><Code>InvalidAccessKeyId</Code><Message>The AWS Access Key Id you provided does not exist in our records.</Message><Resource>/bbucket</Resource><RequestId></RequestId></Error>
```

As I understand public-read and public-reader-write, it should allow access to the AllUsers group, which includes anonymous requests.

reiddraper:
Ah, it seems to work OK when I set auth_bypass back to false

reiddraper:
+1 to merge

PR #99, opened 2012-03-13
Updates to get_fsm to work with sibling resolution changes.

kellymclaughlin:
- Update `riak_moss_get_fsm` to work with the object sibling resolution changes from the `s24-lfs-siblings` branch. 
- Change get_fsm to support retrieval of a limited number of file blocks concurrently. Blocks are retrieved and buffered if any of the blocks is returned out of order. The concurrency level can be adjusted by setting the `fetch_concurrency` property in the app.config file. The default is 1.


reiddraper:
+1 to merge

PR #100, opened 2012-03-13
Refactor the forbidden/2 functions to ease access tracking

beerriot:
The `forbidden/2` functions of `riak_moss_wm_key` and `riak_moss_wm_bucket` are where the hooks for access logging need to go.  The versions of these functions on the master branch are deeply nested and full of redundant code.  I had to refactor them on my logging branch to be able to understand where the hooks belonged.  This PR is my refactor without the logging hooks included, such that the refactor can be reviewed on its own.

While the diffs don't lie, I think they make the change look like more of a rewrite than it is.  The goal was to decrease nesting and remove redundant code.  It means there is a little bit of spaghetti, as functions call other functions instead of nesting their cases directly, but I think it has made the phases of authorization a little bit more componentized.  Please read and test.

No user-visible functionality should have changed, so if something _is_ different, yell loudly.

### DIFF
--- a/src/riak_moss_wm_bucket.erl
+++ b/src/riak_moss_wm_bucket.erl
@@ -39,87 +39,72 @@ malformed_request(RD, Ctx) ->
 %%      authenticated. Normally with HTTP
 %%      we'd use the `authorized` callback,
 %%      but this is how S3 does things.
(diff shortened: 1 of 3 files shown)


kellymclaughlin:
Looks really good. There will be some work to merge my object acl changes with this work, but nothing that should cause too much trouble. +1 to merge.

PR #101, opened 2012-03-13
S24 lfs siblings

reiddraper:
These changes make file storage in Moss "sibling-aware". This gives us eventual consistency semantics like s3. Most notably, block-level garbage collection (even on explicit delete) is missing. This shouldn't be noticeable from the perspective of someone using the API.

The best documentation for the manifest resolution is on the wiki, [here](https://github.com/basho/riak_moss/wiki/Manifest-Sibling-Resolution)

Documentation about the roles of the Get/Put/Delete coordinators, manifest fsm, and blocks server are [here](https://github.com/basho/riak_moss/wiki/Large-file-process-responsibilities).

There is decent EQC test coverage of the pure sibling resolution code. Tests for the Put fsm are quite bare, and currently write directly to Riak (though only a few bytes). This can be fixed when the Put FSM tests are beefed up.

For testing, you'll want to make sure to set `{allow_mult, true}`. All of the usual s3cmd suspects should work as expected. Concurrent writes to the same key should yield the last written one, on a get.


src/riak_moss_block_server.erl
```diff
@@ -0,0 +1,192 @@
...
+%%%===================================================================
+%%% gen_server callbacks
+%%%===================================================================
+
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% Initializes the server
+%%
+%% @spec init(Args) -> {ok, State} |
```
kellymclaughlin, line 68: Are these big comment blocks from a template? I find them a bit noisy. I have specs for the `gen_server` callbacks in the stanchion_server module you could copy if you want. The blocks are fine if you like them, but it'd be good to convert to real specs at some point.

reiddraper, line 68: Yep, just from a template. I agree, they're just line noise. I'm happy to just can them for the time being, and reintroduce when we do a dialyzer-day.

kellymclaughlin, line 68: Nah, don't worry about them for now. It's a detail we can fix later.

kellymclaughlin:
Looks good. Nice job! +1 to merge.

PR #102, opened 2012-03-13
Change LFS chunk naming scheme

### LINKED ISSUE #175: Metadata header with &nbsp in the value causes 403
Marcel reported this migrating some Java code from S3 to CS. The following is the relevant Java snippet:

ObjectMetadata metaData= new ObjectMetadata();
metaData.addUserMetadata("title", "Test\u00A0");
s3.putObject(new PutObjectRequest(bucketName, key, new FileInputStream(createSampleFile()), metaData));

slfritchie:
Fixes #175, partially.
- Change KV bucket naming convention for S3 object bucket and LFS block bucket.
- Change KV key naming convention for LFS block keys.

Still missing: changes to stanchion to create a UUID for each S3 bucket so that the KV bucket name can be a fixed size.

### DIFF
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -25,7 +25,14 @@
           creation_date :: string(),
           modification_time :: erlang:timestamp(),
           acl :: acl_v1()}).
--type moss_bucket() :: #moss_bucket_v1{}.
+-record(moss_bucket_v2, {
+          name :: string(),
+          last_action :: created | deleted,
+          bucket_id :: binary(),
+          creation_date :: string(),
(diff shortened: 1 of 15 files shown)


kellymclaughlin:
Looks good. +1 to merge.

kellymclaughlin:
The bucket name mapped to uuid scheme will not play well with object ACLs and access to objects when the requesting party is not the bucket owner. Closing this request in favor of [this](https://github.com/basho/riak_moss/pull/107) pr to use a hash of the actual bucket name instead.

PR #104, opened 2012-03-14
Check group grants even if user has other grants specified in the ACL.

kellymclaughlin:
Group grants were ignored if the user issuing the request had other
permissions in the ACL. This change makes sure group grants are also
check in that case.
## Testing
1. Create a bucket with public read access: `s3cmd mb -P s3://testbucket`
2. Add `READ_ACP` permissions for a user who is not the bucket owner: `s3cmd setacl --acl-grant=read_acp:<user_email_here> s3://testbucket`
3. Attempt to get the bucket's info using the credentials of the non-bucket-owner user: `s3cmd info s3://testbucket`

On `master` this should fail with a `403 Forbidden` error, but should work using the `s177-fix-group-grant-check` branch.


reiddraper:
+1

PR #105, opened 2012-03-14
S157 multi backend

slfritchie:
- Add multi backend source that has a bucket name prefix matching feature
- Add README.multi-backend.txt file with (still buggy) instructions for configuring the multi backend for Riak
- Add the still-unfinished fs2 backend, based originally on the Riak 0.14 riak_fs_backend.erl and updated
  enough capability to put & get objects and to use on-disk checksums.

### DIFF
--- a/README.multi-backend.txt
+++ b/README.multi-backend.txt
@@ -44,7 +44,7 @@ Step 2: Edit the Riak "app.config" file to configure the multi backend
                 /opt/riak/data
 
             Riak developer source data directory path:
-                ./rel/riak/data
+                ./data
 
     Step b: Use a text editor to edit the 'riak_kv' section of the Riak
             "app.config" file.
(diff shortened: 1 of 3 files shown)


README.multi-backend.txt
```diff
@@ -0,0 +1,80 @@
...
+    Step a: Identify the directory that will be storing Riak's data.
+
+            Riak Linux data directory path:
+                /var/lib/riak
+
+            Riak Solaris data directory path:
+                /opt/riak/data
+
+            Riak developer source data directory path:
+                ./rel/riak/data
```
reiddraper: I _think_ this should just be `./data`, as the path will be relative from the app.config location.

reiddraper: Rather, when you refer to it later as the substitution for DATADIR.

slfritchie: Yup, you're right.  I'll fix it.

reiddraper:
I wasn't sure if there was a better way to test, but I've been able to confirm that the Right Backend is used by just deleting the whole _other_ data dir and having data still persist. Looks good to me, +1. Nice clear instructions too, btw.

PR #106, opened 2012-03-15
Object Acls

kellymclaughlin:
This changeset does the following:
- Adds object ACLs to the file manifest and ensure a default ACL is created with each object (unless a canned ACL is specified at creation time)
- Adds interface support for setting and retrieving object ACLs to `riak_moss_wm_key`.
- Change `acl_v1()` spec type to `acl()`
- Add `s3api_verification_test.py` script as a building block for s3 API compatability.
- Remove the `prototype_parity.y` script. It is deprecated in favor of `s3api_verification_test.py`.
## Testing
- There are 4 test cases in `s3api_verification_test.py` that deal with ACLs. Running the test requires the full `riak_cs` system to be up and running.
- Use `s3cmd setacl` and `s3cmd info` to verify that object ACLs behave as expected. I made some on the [ACL wiki page](https://github.com/basho/riak_moss/wiki/ACLs) about an interesting case with `READ_ACP` and `WRITE_ACP` permissions on objects.


reiddraper:
I get an XML parse error trying to do `s3cmd info s3://bucket/key` on a bucket/key that I _don't_ have permission to view.
[Here's](https://gist.github.com/69f9f789f27e63763ef3) the traceback.

reiddraper:
> I get an XML parse error trying to do s3cmd info s3://bucket/key on a bucket/key that I don't have permission to view.

Fixed with manually installing new s3cmd from [here](https://github.com/s3tools/s3cmd). Was an s3cmd bug, not Moss.

reiddraper:
Excellent, +1 to merge

PR #107, opened 2012-03-16
Fixed-length bucket names for objects and blocks buckets.

kellymclaughlin:
Use an md5 hash of the bucket prefix and the `riak_cs` bucket name as the `riak` bucket name for the `objects` and `blocks` buckets for a `riak_cs` bucket. This ensures that the `riak` bucket names have a fixed size and that straightforward to determine them fromthe `riak_cs` bucket name.

This does not include any checks for collisions where the requested bucket may hash to the same digest as another bucket in the system and result in unauthorized access to the existing bucket. This check will be much more straightforward to add once the object ACL pull request has been merged. A separate task is on the sprintly board for that.
## Testing

Ensure that put, get, and delete of a file work as expected.

### DIFF
--- a/src/riak_moss_block_server.erl
+++ b/src/riak_moss_block_server.erl
@@ -8,6 +8,8 @@
 
 -behaviour(gen_server).
 
+-include_lib("riakc/include/riakc_obj.hrl").
+
 %% API
 -export([start_link/0,
          get_block/5,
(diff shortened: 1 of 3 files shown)


slfritchie:
One thing missing from this PR is that the addition of the S3 bucket and file name isn't propagated to the block metadata dictionary.

kellymclaughlin:
Thanks @slfritchie. [d8e67fb](https://github.com/basho/riak_moss/commit/d8e67fb4d304414ab7d82e4b4e6c6db2a1382570) takes care of that.

reiddraper:
+1

PR #108, opened 2012-03-16
Set default admin credentials that can be used to bootstrap a real admin user

kellymclaughlin:
- Set default admin credentials so that the process of bootstrapping
  an admin user is less of a hassle and so that `riak_cs` and `stanchion`
  will work together with their respective default configurations.
- Set `stanchion_ssl` to `false` as a default to also support the idea
  that `riak_cs` and `stanchion` work together with their default
  configurations.


reiddraper:
+1

PR #110, opened 2012-03-16
Update empty bucket check so when a bucket has no active objects it may be deleted

kellymclaughlin:
Use the manifest_fsm to check active manifests to determine if a bucket may be deleted.

This pr depends on [this](https://github.com/basho/stanchion/pull/8) stanchion pr. 
## Testing

Using `master` of moss and stanchion, create a bucket, store a file in the bucket, delete the file from the bucket, and then attempt to delete the bucket. It will fail. Switch to the `s180-fix-bucket-deletion` branch for moss and stanchion and now the bucket deletion should succeed.

### DIFF
--- a/src/riak_moss_utils.erl
+++ b/src/riak_moss_utils.erl
@@ -457,18 +457,33 @@ bucket_acl_json(ACL, KeyId)  ->
 %% @doc Check if a bucket is empty
 -spec bucket_empty(binary()) -> boolean().
 bucket_empty(Bucket) ->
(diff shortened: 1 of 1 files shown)


kellymclaughlin:
`TODO` added with [c21d4f2](https://github.com/basho/riak_moss/commit/c21d4f228fcb64c9c4225c997f36a28b83977d5f).

reiddraper:
+1 to merge

PR #112, opened 2012-03-19
Usage Logging

beerriot:
Usage Logging

This PR encapsulates the entire usage logging feature, including both
access and storage tracking.  I had intended to squash this into a
single commit for review, but there are other outstanding PRs (object
ACLs, most notably) that will also require merging.  Github's diff
shows pretty well what the changes are, so I'll leave it unsquashed
until the final merge, if that's not too much trouble.

_Please ping me when you start reviewing this PR._ There are a couple
 of small outstanding items that likely aren't _needed_ for 1.0, but I
 may finish if review is delayed.

The documentation of these features should also be considered as part
of this PR.  It might even be a good place to start:
- https://github.com/basho/riak_moss/wiki/Logging-Access-Stats
- https://github.com/basho/riak_moss/wiki/Querying-Access-Stats
- https://github.com/basho/riak_moss/wiki/Logging-Storage-Stats
- https://github.com/basho/riak_moss/wiki/Querying-Storage-Stats

Beyond regular code review, I'm seeking feedback on the following
topics:
- Default configurations.  For example, is a 1-hour archive period
  too long?
- Querying URI.
  - The `/usage/User/Options/Start/End` uri is the first thing I
    could make work with `s3cmd`. I've been kicking around something
    a bit more parameter-oriented, like
    `/usage/User-Options-sStart-eEnd` (note the `s` and `e` denoting
    which time is which, and allowing for reordering or omission,
    similar to the query-param version). Is there another format that
    Amazon uses for a similar style of query?
  - Should the query-param version (`/usage/User?a&b&s=Start&e=End`)
    even stay around?  I'm thinking not, since it's an extra code
    path to maintain, which probably won't be used as often?
- The bucket names and one of the stored fields still contain a
  `moss` prefix. Should this be changed to `rcs`?
- Does anyone know of a reason that I have overlooked, which would
  make my [ab]use of siblings a problem?

I'm also open for discussion of anything else, of course, but these
were the major questions left in my mind.

The list of outstanding issues is below.  I think that 1.0 could ship
without these, but feel free to argue otherwise.
- https://sprint.ly/#!/product/969/item/89
- https://sprint.ly/#!/product/969/item/90
- https://sprint.ly/#!/product/969/item/91
- https://sprint.ly/#!/product/969/item/93
- https://sprint.ly/#!/product/969/item/95
- https://sprint.ly/#!/product/969/item/183
- https://sprint.ly/#!/product/969/item/126
- https://sprint.ly/#!/product/969/item/181


kellymclaughlin:
Minor typo on [Logging Storage Stats](https://github.com/basho/riak_moss/wiki/Logging-Storage-Stats) page: The storage archival period is configured **the byte** riak_moss application environment variable storage_archive_period.

src/riak_moss_access_archiver.erl
```diff
@@ -0,0 +1,226 @@
...
+        %% TODO: might be a good idea to check archiver mailbox
+        %% backlog here, in case stats logging is taking much longer
+        %% than expected
+        ets:give_away(Table, whereis(?SERVER), Slice)
+    catch error:badarg ->
+            lager:error("~p was not available, access stats for ~p lost",
+                        [?SERVER, Slice]),
+            %% if the archiver had been alive just now, but crashed
+            %% during operation, the stats also would have been lost,
+            %% so also losing them here is just an efficient was to
```
kellymclaughlin: s/was/way

src/riak_moss_wm_usage.erl
```diff
@@ -0,0 +1,457 @@
...
+         }).
+
+init(Config) ->
+    %% Check if authentication is disabled and
+    %% set that in the context.
+    AuthBypass = proplists:get_value(auth_bypass, Config),
+    {ok, #ctx{auth_bypass=AuthBypass}}.
+
+service_available(RD, Ctx) ->
+    case riak_moss_utils:riak_connection() of
```
kellymclaughlin, line 148: Doesn't look like this connection ever gets closed and I see the number of open file handles for moss increase with each query.

kellymclaughlin:
If I set `{storage_schedule, "0600"}` in the moss app.config and start riak and moss and then halt riak, moss crashes with the following console output: https://gist.github.com/42a6064939d647f54ed1

priv/sbin/riak-cs-access
```diff
@@ -0,0 +1,29 @@
...
+        if [ "$RES" != "pong" ]; then
+            echo "Node is not running!"
+            exit 1
+        fi
+        shift
+
+        $NODETOOL rpc riak_cs_access_console flush $@
+        ;;
+    *)
+        echo "Usage: $SCRIPT { flush_access }"
```
kellymclaughlin: Should this just be `flush` instead of `flush_access`?

src/riak_moss_storage.erl
```diff
@@ -0,0 +1,107 @@
...
+            {error, Error}
+    end.
+
+object_size_map({error, notfound}, _, _) ->
+    {0,0};
+object_size_map(Object, _, _) ->
+    %% TODO: use Reid's sibling resolution code to choose
+    %% correct value
+    Manifest = binary_to_term(hd(riak_object:get_values(Object))),
+    case riak_moss_lfs_utils:is_active(Manifest) of
```
kellymclaughlin: This function no longer exists so you'll have to access the state field of the manifest directly.

src/rts.erl
```diff
@@ -0,0 +1,419 @@
...
+              12,24,48,96,
+              60,600,3600,21600, % 1min, 10min, 1hr, 6hr
+              86400,86401,86500]). % 1day, over 1 day (will be trunc'd)
+
+%% Check that all elements in the list are boolean true, or 2-tuples
+%% where the second element is true; return 'true' if this is the
+%% case, or the input List if it is not.  This is a useful alternative
+%% to many 'andalso' clauses at the end of an EQC test, as it's easier
+%% to tell which clause failed.
+conj(List) ->
```
kellymclaughlin: You may already be aware of this so feel free to totally ignore me in that case, but there is also an `eqc:conjunction` function that is useful for similar purposes.

src/rts.erl
```diff
@@ -0,0 +1,419 @@
...
+                                        binary_to_list(Postfix))},
+                      {key_time, 0 /= string:str(
+                                        binary_to_list(riakc_obj:key(Obj)),
+                                        binary_to_list(iso8601(SliceStart)))},
+                      {ctype, "application/json" ==
+                           riakc_obj:md_ctype(
+                             riakc_obj:get_update_metadata(Obj))},
+
+                      {start_time, rts:iso8601(Start) ==
+                           proplists:get_value(<<"start_time">>, MJ)},
```
kellymclaughlin: This property fails because `?START_TIME` and `?END_TIME` are defined as `<<"StartTime">>` and `<<"EndTime">>` respectively in `rts.hrl`.

src/rts.erl
```diff
@@ -0,0 +1,419 @@
...
+
+%% check that an archive object is in the right bucket, with a key
+%% containing the end time and the username, with application/json as
+%% the content type, and a value that is a JSON representation of the
+%% sum of each access metric plus start and end times
+make_object_prop() ->
+    ?FORALL({Bucket, Postfix,
+             T0, T1, Period},
+            {string_g(), string_g(),
+             datetime_g(), datetime_g(), valid_period_g()},
```
kellymclaughlin, line 348: I would use `eqc_gen:noshrink/1` around the start and end time generators. I don't see any real value in shrinking the dates and it makes tracking down failures more difficult imo.

beerriot, line 348: I think we just decided to leave this for now.  It will be a quick thing to change if troubleshooting this test comes up again.

src/riak_moss_access_logger.erl
```diff
@@ -0,0 +1,416 @@
...
+-define(STAT(Name), {access, Name}).
+-define(EXPECT_BYTES_OUT, expect_bytes_out).
+-define(BYTES_IN, bytes_in).
+
+%% @doc Set the MOSS user for this request.  Stats are not recorded if
+%% the user is not set.
+set_user(?MOSS_USER{key_id=KeyID}, RD) ->
+    wrq:add_note(?STAT(user), KeyID, RD);
+set_user(#moss_user{key_id=KeyID}, RD) ->
+    %% TODO: does this support for old user records need to be here?
```
kellymclaughlin: Not really, at&t has already been migrated to the new user record structure so I think it's safe to remove if you want.

kellymclaughlin:
+1 to merge from me pending reid's review. Very nice work!

reiddraper:
It'd be convenient to expose a "riak_moss_usage" (maybe a better name) section in the app.config for things like `access_log_flush_interval`

reiddraper:
+1 from me too. Excellent work!

reiddraper:
+1 on 8e9c3dc

beerriot:
Squashed and merged as https://github.com/basho/riak_moss/commit/079772bfb1deabca98c6027bbb0b0cdf866b4620

PR #113, opened 2012-03-20
Monitor webmachine process during file download.

kellymclaughlin:
Monitor the calling webmachine process from the get fsm so that if the
client connection is prematurely terminated then it can be detected by
the get fsm and the retrieval of file blocks can be halted.
## Testing
1. Start `riak_cs` in console mode or attach to the console.
2. Load a large file into `riak_cs`. _i.e._ One that will take at least several seconds to download.
3. Set the lager console backend to debug level: `lager:set_loglevel(lager_console_backend, debug).`
4. Begin to download the file from `riak_cs` using `s3cmd`, but hit `CTRL+C` to terminate the download before it completes. 

On the `master` branch the console output will show that the blocks of the file continue to be fetched from Riak until they have all been retrieved. Using the `s184-wm-monitor-on-download` branch the block retrieval will halt once the webmachine process dies and the notification of its death is received and processed by the get fsm.

### DIFF
--- a/src/riak_moss_get_fsm.erl
+++ b/src/riak_moss_get_fsm.erl
@@ -21,7 +21,7 @@
 -include("riak_moss.hrl").
 
 %% API
--export([start_link/2,
+-export([start_link/3,
          stop/1,
          continue/1,
          manifest/2,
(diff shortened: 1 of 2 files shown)


reiddraper:
+1 to merge

PR #115, opened 2012-03-21
Overhaul basho_bench plugin to use S3 auth

slfritchie:
When using the new basho_bench config items as shown in the sample
config file, the RCS server's `auth_bypass` config knob can keep its
default `false` value.


kellymclaughlin:
Look good! +1 to merge.

PR #116, opened 2012-03-22
gen_server & related call timeout review

slfritchie:
Fixes #191

Most of the calls are gen_fsm sync event calls.  For both gen_server
and gen_fsm servers, all procs are intra-VM and therefore there's no
reason that I know of why we shouldn't be using an infinity timeout.

I know that these changes fix at least one bug: the gen_fsm sync call
to fetch manifests for all objects in a bucket (triggered by `s3cmd ls`
of a bucket with many thousands of files can take more than 5 seconds
on a slow machine.


reiddraper:
+1

PR #117, opened 2012-03-22
Stop the manifest fsm on list keys

reiddraper:
Fixes #[192](https://sprint.ly/#!/product/969/item/192)

Previously we left the manifest
fsm running for each object in a
list keys. This shuts the
fsm down after each object.


slfritchie:
+1, much less spam, yay.

PR #120, opened 2012-03-23
Resolve race condition when transitioning from initialization to the first state.

kellymclaughlin:
The most straightforward thing to do here is just to eliminate the `prepare` state and handle all of the initialization in the `init` function. Using a `prepare` state does not seem to have much advantage in this case so removing it seems like the logical solution. There is also a change to the `init` function clause that handles test setup so that the get_fsm tests still work as expected.
## Testing

I believe `basho_bench` was used to generate the load initially revealed this problem. Also [here](https://gist.github.com/d39e7657b6bb5b39e2f4) is more information on a minimal set of modules that can be used to reproduce the root problem.

### DIFF
--- a/src/riak_moss_get_fsm.erl
+++ b/src/riak_moss_get_fsm.erl
@@ -32,8 +32,8 @@
 %% gen_fsm callbacks
 -export([init/1,
          prepare/2,
+         prepare/3,
          waiting_value/3,
-         waiting_metadata_request/3,
          waiting_continue_or_stop/2,
          waiting_chunks/2,
          waiting_chunks/3,
@@ -99,7 +99,6 @@ init([Bucket, Key, Caller]) ->
     %%    terminations, so this process would still
     %%    live.
     CallerRef = erlang:monitor(process, Caller),
-
     %% we want to trap exits because
     %% `erlang:link` isn't atomic, and
     %% since we're starting the reader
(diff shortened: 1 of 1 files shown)


reiddraper:
Looks like this is going to have a minor conflict with f64b014bc4a1993971f44d964a1c9e2c04380f70, wanna merge/rebase with it and then I'll review after?

kellymclaughlin:
Rebased and resolved issue that arose. Basically if a manifest is not found there is no manifest_fsm or block servers to shut down so I make that distinction.

slfritchie:
Hrm.  In Riak Core & KV land, the prepare state was introduced to help purify the state machine to make it easier to test.
It's also there to do things that might take a long time, because OTP behavior init() functions are supposed to be quick ... because they're frequently started by a supervisor which is does all of its work serially.

If you get an event or message before you've done your stage-two-might-take-a-while initialization, then I think there should be info in #state that says that you aren't fully initialized yet.  Then handling any timeout race is easy, at the expense of an extra bit of code for each of several gen_fsm callbacks, e.g.,

```
handle_event(Event, StateName, #state{prepared = false} = S) ->
    %% prepare/1 will flip prepared to true
    %% We assume that prepare always succeeds ... need more code here if it can fail
    handle_event(Event, Statename, prepare(S));
handle_event(FirstRealEvent, ....) ->
    ....
```

reiddraper:
> Then handling any timeout race is easy, at the expense of an extra bit of code for each of several gen_fsm callbacks, e.g.,

I guess this boilerplate code needs to be added to a single `handle_event/3` clause and a single `handle_sync_event/4` clause? Certainly a bit annoying, but perhaps our best option, considering @slfritchie's remark about supervisors handling `init` funs serially.

kellymclaughlin:
Yeah the point about supervisors is a good one. I'll rework this a little. If I understand the `gen_fsm` code correctly I think explicitly sending the `timeout` event in the `init` function instead of relying on the 0 timeout trigger avoids the possibility of any events getting onto the message queue before the `timeout` event and maybe that's a way to have an `init` that is lightweight and also avoid lots of extra annoying code.

reiddraper:
Nice job, +1 to merge

kellymclaughlin:
@reiddraper I went ahead and removed the `waiting_metadata_request` state since it's no longer used. Could you just verify that I am correct in that assertion. Thanks.

reiddraper:
@kellymclaughlin yep, +1

PR #121, opened 2012-03-23
Fix process and socket leaks

reiddraper:
Moss had a process (and therefore socket)
leak that was apparent when using
persistent HTTP connections. It wasn't
as obvious when connections were torn down,
because the processes were terminated because
of monitors. Webmachine reuses the same process
throughout a persistent HTTP connection, so the
processes would continue to leak until
the socket was closed.

The fix is two-fold:
1. Stop the get_fsm used for object
   ACLs during PUT and DELETE requests. This
   was simply an oversight previously.
2. Shutdown the manifest_fsm and
   block servers in the get_fsm
   terminate call.

Fixes sprintly #189


kellymclaughlin:
Looks fixed to me. Nice work! +1 to merge.

PR #122, opened 2012-03-27
Dss riakc pool

reiddraper:


src/riak_moss_utils.erl
```diff
@@ -407,24 +406,13 @@ put_object(BucketName, Key, Value, Metadata) ->
...
-    end,
-    riak_connection(Host, Port).
+    Worker = poolboy:checkout(riakc_pool, true, infinity),
+    {ok, Worker}.
 
 %% @doc Get a protobufs connection to the riak cluster.
--spec riak_connection(string(), pos_integer()) -> {ok, pid()} | {error, term()}.
-riak_connection(Host, Port) ->
-    riakc_pb_socket:start_link(Host, Port).
+%% -spec riak_connection(string(), pos_integer()) -> {ok, pid()} | {error, term()}.
```
kellymclaughlin, line 413: Delete this.

slfritchie:
The PR looks pretty straightforward.  My problem is that I don't know enough about the additional ways that poolboy may complicate things that aren't addressed in this PR.  Does the PB socket have an inactivity timer?  If yes, then some RCS proc could get a poolboy-assigned PB connection that was dead.  I don't believe that it's possible for close_riak_connection() to be called when a PB call times out (60 seconds?), but if it were, then a later poolboy checkout could get a reply sent by somebody else.  I guess I need to read more riakc_pb_socket code.

slfritchie:
FWIW, see also the later comments in https://sprint.ly/product/969/item/193 ... connection pooling is a huge help there.

slfritchie:
Hrm, I think we definitely need more work.  -1, sorry.  If you shut down Riak while RCS is under any kind of load (e.g. `s3cmd ls://bucket-with-35K-files-in-it`), then you'll get a supervisor shutdown and then app shutdown like this:

```
Supervisor riak_moss_sup had child riakc_pool started with poolboy:start_link([{name,{local,riakc_pool}},{worker_module,riak_moss_riakc_pool_worker},{size,10},{max_overflow,100},...]) at <0.82.0> exit with reason reached_max_restart_intensity in context shutdown
2012-03-27 22:11:11.606 [error] <0.11922.0> gen_server <0.11922.0> terminated with reason: no match of right hand value {error,{tcp,econnrefused}} in poolboy:new_worker/2
2012-03-27 22:11:11.623 [error] <0.11922.0> CRASH REPORT Process <0.11922.0> with 0 neighbours crashed with reason: no match of right hand value {error,{tcp,econnrefused}} in poolboy:new_worker/2
2012-03-27 22:11:11.639 [info] <0.7.0> Application riak_moss exited with reason: shutdown
2012-03-27 22:11:11.640 [info] <0.131.0> alarm_handler: {clear,system_memory_high_watermark}
```

kellymclaughlin:
Yeah I noticed the same thing this afternoon.

kellymclaughlin:
There is also a deadlock condition that can occur if all the connections in the pool are used up by concurrent actors. Since the fsms hold connections and multiple connections are used per request for things like access control and authentication, it is possible to get all the workers blocking and waiting for a connection to be returned to the pool when none will ever be freed. This behavior can be seen by using `riakc_pool` settings of `{3, 0}` and 3 concurrent b_b workers. The solution for this finer granularity of connection control as has been discussed already this week.

reiddraper:
I can confirm I've seen the deadlock happen In The Wild.

PR #123, opened 2012-03-27
ACL ownership information update

kellymclaughlin:
These changes add an `acl_v2` record and make the changes necessary to include the user's `key_id` as part of the acl ownership information. The reason behind this is to eliminate the need for an expensive 2I query on each object put request and to only use the secondary index to look up user by `canonical_id` when absolutely necessary. These changes are backward compatible so existing buckets and objects with `acl_v1` ACLs will continue to work. 

These changes also rely on [this](https://github.com/basho/stanchion/pull/10) stanchion pr. When testing be sure to use the `s209-acl-owner-update` branch of stanchion both to run stanchion and checked out in the deps of riak_moss.

### DIFF
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -42,7 +42,7 @@
                       putctype :: string(),
                       bucket :: binary(),
                       key :: list(),
-                      owner :: moss_user(),
+                      owner :: string(),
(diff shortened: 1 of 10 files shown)


reiddraper:
+1

PR #128, opened 2012-03-29
One connection per request

kellymclaughlin:
Changes so that a single get or put request for a bucket or an object uses the same connection to riak throughout the lifecycle of the request. This should avoid some deadlock issues that have been seen using the connection pool and be a more efficient way to use the connection pool.


reiddraper:
Excellent work, +1 to merge

PR #129, opened 2012-03-29
Fix PUTs for zero-byte files

reiddraper:
S212

Skip all of the block writing
state if the content length
of the file is 0.


kellymclaughlin:
Looks good and tests with 0 bytes files pass. +1 to merge.

PR #132, opened 2012-03-30
skip recalculations on scheduled runs and manual trigger without -r flag

beerriot:
This patch changes the behavior of the batch storage calculation
process.  By default, the process now skips calculation for any user
that already has a calculation stored for the given time period.  This
allows calculation to be stopped and restarted without chugging through
already calculated users again.  This also lays the groundwork for
multiple MOSS nodes calculating storage for all users concurrently.

Recalculating and storing new results can be forced by providing the -r
or --recalc flags on the command line.


src/riak_moss_storage_d.erl
```diff
@@ -73,8 +75,17 @@ status() ->
...
-start_batch() ->
-    gen_fsm:sync_send_event(?SERVER, start_batch, infinity).
+%%
+%% Allowed options are:
+%% <dl>
+%%   <dt>`recalc'</dt>
+%%   <dd>Recalculate the storage for each user, even if that user
+%%   already has a calculation stored for this time period. Default is
+%%   `false', such that restarting a canceled batch does not require
+%%   redoing the work that happened before cancelation.</dd>
```
kellymclaughlin: s/cancelation/cancellation

src/riak_cs_storage_console.erl
```diff
@@ -116,6 +129,8 @@ human_detail(elapsed, Elapsed) ->
     {"Elapsed time of current run", integer_to_list(Elapsed)};
 human_detail(users_done, Count) -> 
     {"Users completed in current run", integer_to_list(Count)};
+human_detail(users_skipped, Count) -> 
+    {"Users skippeded in current run", integer_to_list(Count)};
```
kellymclaughlin: Extra _ed_ on _skippeded_.

kellymclaughlin:
Inclusion of `getopt` causes release building issues since the eper dependency already includes it. After talking with Bryan, we will just rely on the `getopt` provided by eper.

kellymclaughlin:
Can you just make sure to add a snippet about the new `-r` option to the wiki page too.

kellymclaughlin:
+1 to merge.

PR #133, opened 2012-03-30
access archiver and storage daemon shouldn't use pool connections

beerriot:
we don't want them to be blocked by user requests, nor to block user
requests


reiddraper:
+1

PR #134, opened 2012-03-31
Return no_user_key when there is no user for a request.

cmeiklejohn:
Ensure that we return no_user_key when the user making the request is unauthenticated.

Fixes the following server error, when accessing a public asset without authentication.

```
{error,badarg,
       [{erlang,list_to_binary,[undefined]},
        {riak_moss_utils,get_user,2},
        {riak_moss_wm_service,forbidden,2},
        {webmachine_resource,resource_call,3},
        {webmachine_resource,do,3},
        {webmachine_decision_core,resource_call,1},
        {webmachine_decision_core,decision,1},
        {webmachine_decision_core,handle_request,2}]}
```

I wasn't sure the correct way to approach the eunit test for this functionality.  Should I just unit test the get_user method, or do we prefer to test this a different way?


cmeiklejohn:
Reopening against 1.0 branch.

PR #138, opened 2012-04-02
allow dot separators in usage path instead of slashes (fixes #219)

beerriot:
Some S3 libraries auto-quote extra slashes in the path, so
    /usage/Key/Opts/Start/End
becomes
    /usage/Key%2fOpts%2fStart%2fEnd
so the patch allows
    /usage/Key.Opts.Start.End
instead.  This also helps simple s3cmd usage, as it will automatically
write to a file named Key.Opts.Start.End instead of just End.


reiddraper:
+1

PR #140, opened 2012-04-03
Trigger access archive on log size (helps fix #220)

beerriot:
This is an attempt to avoid the situation where a single user with _many_ accesses takes a _long_ time to archive.  I've set the default trigger for one million accesses.  That's probably high for an underpowered cluster, and low for beefy hardware.  I'm open to suggestion for a better default.

To test:

Set `{access_log_flush_size, 10}` in your app.config.  Start riak-cs.  Run:

```
s3cmd get s3://usage/YOUR_KEY/abj/20120304T000000Z/20120304235959Z
```

ten times.  Pause for a breath, then run it an eleventh time.  On the eleventh, you should see a new access sample in the result.  (be sure you modify the dates in that example to cover the time you're experimenting)

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -38,6 +38,11 @@
               %% twice per period, etc.)
               {access_log_flush_factor, 1},
               
+              %% Additional access log flush trigger - flush after
+              %% this many accesses are recorded, even if the flush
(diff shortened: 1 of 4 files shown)


reiddraper:
+1

PR #141, opened 2012-04-04
limit amount of usage stats allowed in one request (fixes #201)

beerriot:
Set the `riak_moss` application environment variable `usage_request_limit` to the number of archive periods that a user is allowed to request in a single usage read.  Default is 744, which is one month's worth of usage data at the default archive period of one hour.

Should this be expressed as a number of seconds instead of a number of periods?

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -55,7 +55,13 @@
               %% chosen such that each storage_schedule entry falls in
               %% a different period; integer number of seconds
               %% (86400 == 1 day)
-              {storage_archive_period, 86400}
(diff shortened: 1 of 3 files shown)


beerriot:
To test:

```
s3cmd get s3://usage/KEY/abj/20120403T000000Z/20120503T000000Z
```

should succeed (one month 3.Apr - 3.May)

```
s3cmd get s3://usage/KEY/abj/20120403T000000Z/20120603T000000Z
```

should fail (two months 3.Apr - 3.Jun)

reiddraper:
+1

PR #148, opened 2012-04-05
Set content-type in context when using the default type.

kellymclaughlin:
If no content-type header is specified in a request the default header type is not set in the key_context in the `riak_moss_wm_key` webmachine resource. This causes a badarg error when the accept_body function calls `list_to_binary` on the content-type from the context and the value is undefined.
## Testing

Use the following ruby script to reproduce or send a request with the Content-Type header not set: 

https://gist.github.com/db9006729bedd3b92ab7

### DIFF
--- a/src/riak_moss_wm_key.erl
+++ b/src/riak_moss_wm_key.erl
@@ -233,7 +233,10 @@ delete_resource(RD, Ctx=#key_context{bucket=Bucket,
 content_types_accepted(RD, Ctx) ->
     case wrq:get_req_header("Content-Type", RD) of
         undefined ->
-            {[{"application/octet-stream", accept_body}], RD, Ctx};
+            DefaultCType = "application/octet-stream",
+            {[{DefaultCType, accept_body}],
+             RD,
+             Ctx#key_context{putctype=DefaultCType}};
         %% This was shamelessly ripped out of
         %% https://github.com/basho/riak_kv/blob/0d91ca641a309f2962a216daa0cee869c82ffe26/src/riak_kv_wm_object.erl#L492
         CType ->


kellymclaughlin:
True, the only reason for that is that webmachine did not like when `binary/octet-stream` was used. I didn't ever dig deep enough to find out why, but it wasn't working for me using the real default.

reiddraper:
I've started from a fresh user and `rm -rf rel/riak/data/*`, and I'm still getting [these errors](https://gist.github.com/9f2f0fa6afcdfef1c5c5).

slfritchie:
I see the same errors as Reid.

kellymclaughlin:
First off, there's a problem that the `stanchion` tag referenced by the `riak_moss` `1.0` branch has the `0.2.0` tag of `node_package` referenced whereas the `riak_moss` repo references `0.2.1` in its rebar.config. This can cause some issues getting a good build. I just edited the rebar.config of the stanchion dep to reference `node_package` `0.2.1` to avoid this problem.

Here's an updated test scenario:
1. Check out the `1.0` branch of `riak_moss` and `stanchion` and ensure all deps are updated appropriately.
2. Build a release of `riak_moss` and `stanchion`.
3. Set `auth_bypass` to true in the app.config of `riak_moss` and `stanchion`.
4. Create a user for testing if one does not already exist.
5. Create a bucket for this test by replacing the value of the `Authorization` header with the `key_id` of the user created in the previous step:

```
 curl -X PUT s3.amazonaws.com/testbucket --proxy1.0 "localhost:8080" -H 'Authorization: Q5U52KT6JYEF2EUPL8DP' -v
```
1. Attempt to put an object into the test bucket with a request that does not include a `Content-Type` header.

```
curl -X PUT s3.amazonaws.com/testbucket/testfile --proxy1.0 "localhost:8080" --data-binary @testfile -H 'Authorization: Q5U52KT6JYEF2EUPL8DP' -H 'Content-Type:' -v
```

The curl request should return output similar to this: https://gist.github.com/51610fc3dc7907977836
7. Now check out the `gh143-default-ct-not-set` branch of `riak_moss`, rebuild, and update the release. 
8. Rerun step 6 and this time the curl request should be successful and return a `204`.

slfritchie:
+1 to merge, using this script to automate the bug provoking recipe:
https://gist.github.com/a4aceeef657dd6d1c5ee

PR #161, opened 2012-04-09
Use warnings_as_errors directive in rebar.config.

kellymclaughlin:
- Use `warnings_as_errors` directive in rebar.config.
- Comment out riak_kv_backend behavior declaration in riak_kv_fs2_backend since we are not building under riak_kv and do not wish to have riak_kv as a dependency.


slfritchie:
+1 to merge, thanks!

PR #165, opened 2012-04-10
Correct inappropriate use of riak_moss_utils:get_user.

kellymclaughlin:
`riak_moss_acl_utils:get_owner_data` calls `riak_moss_utils:get_user` when it needs to call `riak_moss_utils:get_user_by_index`. Additionally it needs a riak connection pid to use for the `get_user_by_index` call. These changes correct `riak_moss_acl_utils:get_owner_data` so it calls `get_user_by_index` and also plumbs through changes so a riak connection pid is available to use for that call.


slfritchie:
+1 to merge, thanks: I've merged this patch into my gh144-dialyzer-warning-review with good results.

Issue #146, opened 2012-04-05
Timeout access archiver


beerriot:
The archiver process should abort the archival of the current timeslice if the total time for archiving it exceeds some configured time, to prevent compounded backlog. Currently timeouts only happen on a per-user-slice storage operation basis.

This issue was originally filed as sprint.ly 95, a sub-task of the initial usage-tracking story, and later as sprint.ly 202.

PR #168, opened 2012-04-13
Archiver Timeout and Client Disconnect Handling

beerriot:
Fixes issue #146.

This PR does two main things:
1. It prevents the backlog of logs needing to be archived from growing too large.  It does this by dropping work on the current log if too many new logs arrive before it's finished.
2. It handles the death or disconnection of the Riak client its using, without dropping the entire rest of the log (and any logs in its mailbox) on the floor.

Accomplishing these goals required two redesigns:
1. The archiver is now "more fluid", in that it stays in the `archiving` state, but sends itself `continue` messages after archiving each user.  This allows other messages, such as `ETS-TRANSFER` and `DOWN` to be received in the during a log archival.
2. The archiver is now started under its own supervisor, and instead of `start_link`ing its Riak client directly, it asks the supervisor to start it instead.  This prevents the death-by-failed-startup Scott described last week.  To manage the association between the archiver and the client, the supervisor is set up for `rest_for_one` restarts.  This means that if the archiver dies, its client will be torn down by the supervisor (because the archiver was started first), but if the client dies, the archiver stays alive (so it doesn't lose the logs its working on).  The client is started `temporary` and the archiver `monitor`s it, allowing the archiver to manage the restart of the client when appropriate.

This design will also allow the easy addition of some basic stats, and a `status` console command to go along with the `flush` command.  I believe a similar arrangement for the starting and monitoring of the storage daemon's riak client should also be used.


slfritchie:
Conditioning from my previous gig makes a reflex quite strong, one that takes strong interest in anything that loses billing information ... 'cause billing info == money.  As a short term thing, dropping work sounds better than blowing up a bit later.  However, is it feasible to drop work by scribbling it to disk and slurping it up sometime later?  Billing data loss due to a file system exploding/disk dying is generally viewed differently than losing billing data due to mere overload.  {shrug}

beerriot:
@slfritchie Completely agreed.  There's TODOs all over for using a disk-based log instead of ETS.  This PR doesn't make us any _more_ likely to lose billing information, and I'd prefer to handle that problem as part of solving those other TODOs.

src/riak_moss_access_archiver.erl
```diff
@@ -14,21 +14,38 @@
...
 -define(SERVER, ?MODULE). 
+-define(DEFAULT_MAX_BACKLOG, 2). % 2 ~= forced and next
 
--record(state, {}).
+-record(state,
+        {
+          riak,      %% the riak connection we're using
+          mon,       %% the monitor watching our riak connection
+          table,     %% the logger table being archived
+          slice,     %% the logger slice bieng archived
```
kellymclaughlin: minor typo: _bieng_

src/riak_moss_access_archiver.erl
```diff
@@ -67,14 +81,18 @@ archive(Table, Slice) ->
             false %% opposite of ets:give_away/3 success
     end.
 
-%% @doc Find out what the archiver is up to.  Should return `{ok, N}`
-%% where N is the number of logs waiting to be archived.
+%% @doc Find out what the archiver is up to.  Should return `{ok,
```
kellymclaughlin, line 84: Is the use of the markdown style _``_ instead of the edoc _`'_ intentional?

kellymclaughlin:
Finally got through all the tests I wanted to do, sorry for the delay. +1 to merge.

PR #174, opened 2012-04-16
Use {notfound_ok, false} for r=1 block requests

### LINKED ISSUE #170: Use of r=1 in block_server causes file retrieval to fail when any Riak cluster node is down.
The block_server fetches chunks using r=1. If any Riak node is down there is a high probability that the downed node houses a primary vnode for one of the file blocks. This leads to a scenario where the responding vnode may respond with {error, notfound} and this causes the file retrieval to fail. If the requesting client is s3cmd then this manifests as s3cmd hanging and chewing up cpu time.

reiddraper:
Fixes #170

Without this patch, requests for
sufficiently large files will
fail when a single Riak node is
down. This happens (presumably) because
the fallback node is the quickest to
respond, and thus the block request
gets a 404. This commit adds
{notfound_ok, false} and {basic_quorum, false}
to the request so that the request
will only return 404 is N (3) vnodes
return 404.


kellymclaughlin:
+1 to merge. Thanks.

PR #177, opened 2012-04-16
Fix test builds broken by enabling warnings_as_errors.

kellymclaughlin:
Enabling `warnings_as_errors` in the rebar.config broke the building and execution of the eunit tests. These changes fix those problems.

I commented-out the test cases in `riak_cs_kv_multi_backend` and `riak_kv_fs2_backend` so that they could still be uncommented and run if copied into the `riak_kv` repo, but maybe they should just be deleted. Feedback on that welcome.
## Testing

With these changes, running `make tests` works again for me.


slfritchie:
+1 to merge, thanks.  NOTE: Riak KV needs to be listening on its default port, 8087 on localhost, for the `riak_moss_put_fsm_test` to avoid errors on its startup phase.

PR #182, opened 2012-04-18
Make a PR without an extra Issue

cmeiklejohn:
It can be a little confusing to have two cards in the Kanban, one for the Issue, and one for the PR that solved the issue.  Github automatically makes an issue for each PR, but it looks like the "hub" helper script has a chance of getting around this.  Try it out.

### DIFF
--- a/README.org
+++ b/README.org
@@ -9,4 +9,3 @@ application that provides an S3 type interface onto Riak.
    system to compile and run =riak_moss=.
 
    *TODO*
-
--- a/README.org
+++ b/README.org
@@ -9,3 +9,4 @@ application that provides an S3 type interface onto Riak.
    system to compile and run =riak_moss=.
 
    *TODO*
+


beerriot:
Looks like this is pretty simple.  Just

```
hub pull-request -i 182
```

and it automatically converts issues/182 into pull/182, right @cmeiklejohn?  Hub comes from https://github.com/defunkt/hub, though you have to check it out and use the `api-v3` branch right now, because Github has disabled the token auth that master uses.

cmeiklejohn:
That's correct.  Also, according to the docs, you should be able to associate an already-open pull with an issue by switching to the branch with the open pull and running that command as well, but we haven't tested that one yet.

beerriot:
Cool.  I think we can close this PR out then, unless you want to experiment more.

beerriot:
@Vagabond also just pointed out that we can do this from the kanban board, using the `pull-request` link on the bottom of the ticket (next to the `ready-to-pull` button)

PR #188, opened 2012-04-20
Gh145 stats and webmachine

slfritchie:
Merged Dizzy's original stats work, then added /riak-cs/stats routing for Webmachine and admin user authentication check.

### DIFF
--- a/rel/reltool.config
+++ b/rel/reltool.config
@@ -15,6 +15,7 @@
          mochiweb,
          webmachine,
          lager,
+         folsom,
          riak_moss
         ]},
        {rel, "start_clean", "",
@@ -29,6 +30,7 @@
        {excl_archive_filters, [".*"]},
        {app, sasl, [{incl_cond, include}]},
        {app, lager, [{incl_cond, include}]},
+       {app, folsom, [{incl_cond, include}]},
        {app, riak_moss, [{incl_cond, include}]}
       ]}.
 
--- a/src/riak_cs_stats.erl
+++ b/src/riak_cs_stats.erl
@@ -11,7 +11,11 @@
 -export([start_link/0,
          update/2,
          update_with_starttime/2,
-         report/0]).
+         report/0,
+         report_json/0,
+         report_pretty_json/0,
+         report_str/0,
+         get_stats/0]).
 
 %% gen_server callbacks
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
@@ -40,17 +44,44 @@
 %% API
 %% ====================================================================
 
+-spec start_link() -> {ok, pid()} | {error, term()}.
 start_link() ->
     gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
 
+-spec update(atom(), integer()) -> ok | {error, {unknown_id, atom()}}.
 update(BaseId, ElapsedUs) ->
     gen_server:call(?MODULE, {update, BaseId, ElapsedUs}).
 
+-spec update_with_starttime(atom(), erlang:timestamp()) ->
+                                   ok | {error, {unknown_id, atom()}}.
 update_with_starttime(BaseId, StartTime) ->
-    gen_server:call(?MODULE, {update, BaseId, timer:now_diff(os:timestamp(), StartTime)}).
+    gen_server:call(?MODULE, {update, BaseId,
+                              timer:now_diff(os:timestamp(), StartTime)}).
 
+-spec report() -> ok.
 report() ->
-    [report_item(I) || I <- ?IDS].
+    [report_item(I) || I <- ?IDS],
+    ok.
+
+-spec report_str() -> [string()].
+report_str() ->
+    [lists:flatten(report_item_str(I)) || I <- ?IDS].
+
+-spec report_json() -> string().
+report_json() ->
+    lists:flatten(mochijson2:encode({struct, get_stats()})).
+
+-spec report_pretty_json() -> string().
+report_pretty_json() ->
+    lists:flatten(riak_moss_utils:json_pp_print(report_json())).
+
+-spec get_stats() -> [{legend, [atom()]} |
+                      {atom(), [number()]}].
+get_stats() ->
+    [{legend, [meter_count, meter_rate, latency_mean, latency_median,
+               latency_95, latency_99]}]
+    ++
+    [raw_report_item(I) || I <- ?IDS].
 
 %% ====================================================================
 %% gen_server callbacks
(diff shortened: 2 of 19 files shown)


src/riak_cs_wm_stats.erl
```diff
@@ -0,0 +1,142 @@
+%% -------------------------------------------------------------------
+%%
+%% stats_http_resource: publishing RiakCS runtime stats via HTTP
+%%
+%% Copyright (c) 2007-2010 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
```
kellymclaughlin: Remove the Apache License

src/riak_cs_wm_stats.erl
```diff
@@ -0,0 +1,142 @@
+%% -------------------------------------------------------------------
+%%
+%% stats_http_resource: publishing RiakCS runtime stats via HTTP
+%%
+%% Copyright (c) 2007-2010 Basho Technologies, Inc.  All Rights Reserved.
```
kellymclaughlin: Might as well update the copyright to 2012.

src/riak_cs_wm_stats.erl
```diff
@@ -0,0 +1,142 @@
...
+          auth_bypass :: boolean(),
+          riakc_pid :: pid(),
+          path_tokens :: [string()]
+         }).
+
+init(Props) ->
+    AuthBypass = proplists:get_value(auth_bypass, Props, false),
+    {ok, #ctx{auth_bypass = AuthBypass}}.
+
+%% @spec encodings_provided(webmachine:wrq(), context()) ->
```
kellymclaughlin, line 38: Does webmachine complain too much about using `-spec` for these?

src/riak_moss_web.erl
```diff
@@ -17,10 +17,17 @@ dispatch_table() ->
         undefined ->
             AuthBypass = false
     end,
+    StatsProps = stats_props(AuthBypass),
+
     [
      {[], riak_moss_wm_service, [{auth_bypass, AuthBypass}]},
+     {["riak-cs", '*'], riak_cs_wm_stats, StatsProps},
+     {["stats"], riak_cs_wm_stats, StatsProps},
```
kellymclaughlin: I think we want to get rid of this line to avoid collision with someone trying to use `stats` as a bucket name and change the previous line to match only for `["riak-cs", "stats"]`

src/riak_cs_wm_stats.erl
```diff
@@ -0,0 +1,142 @@
...
+                    {true, RD, Ctx#ctx{riakc_pid = Pid}};
+                _ ->
+                    {false, RD, Ctx}
+            end
+    end;
+service_available(RD, Ctx) ->
+    {false, wrq:append_to_response_body("Unrecognized path.\n", RD), Ctx}.
+
+produce_body(RD, Ctx) ->
+    Body = mochijson2:encode({struct, get_stats()}),
```
kellymclaughlin: Any reason not to just call `riak_cs_stats:get_stats` directly here?

src/riak_cs_wm_stats.erl
```diff
@@ -0,0 +1,142 @@
...
+
+content_types_provided(RD, Context) ->
+    {[{"application/json", produce_body},
+      {"text/plain", pretty_print}],
+     RD, Context}.
+
+ping(RD, Ctx) ->
+    {pong, RD, Ctx#ctx{path_tokens = path_tokens(RD)}}.
+
+service_available(RD, #ctx{path_tokens = ["stats"]} = Ctx) ->
```
kellymclaughlin: I think I would remove this hardcoded path assumption since the webmachine dispatcher should already have checked this.

slfritchie: Kelly, the `/riak-cs` part has already been stripped off, so ["stats"] should remain.  The 2nd clause for `service_available/2` would then take care of other URIs under `/riak-cs`.

However, I did discover that the append_to_response_body calls aren't doing anything when the URI is `/riak-cs/stats`, so I've removed them.

src/riak_cs_wm_stats.erl
slfritchie: Nope, I figured the `{struct,...}` wrapper was necessary, but it isn't after all: JSON is identical with & without the wrapper.

src/riak_cs_wm_stats.erl
slfritchie: Ja, that makes sense, I should've done it that way in the first place.

... though I ran into an icky problem/bug in `riak_moss_s3_auth:canonicalize_resource/1` that I worked around, please take a peek at it to double-check sanity.  After commit 4cef9f8fe, S3-style authentication was definitely broken until I added 912ce699's fix.

kellymclaughlin: Yes, I know this bug of which you speak. I've stumbled on this a few times and I think this may be the time to fix it. The reason I have previously deferred fixing it is that referencing the bucket names for S3 is most often done by prepending the bucket name onto the host name in the `Host` header and all the clients we have tested use this method and thus laziness prevailed.

slfritchie:
Ok, I believe that I've got all of the HTTP exposure stuff done, Dialyzer warnings fixed (when rebased onto master, which doesn't go 100% cleanly, sorry), and added code to actually increment all counters in Dizzy's list.

slfritchie:
Oops.  Forgot to mention this script, https://gist.github.com/cf4b264128097f0a80d6, that triggers all S3 client ops that will increment all available counters, eventually.  Except for the delete block counter, because its implementation isn't finished yet.

src/riak_cs_stats.erl
```diff
@@ -0,0 +1,162 @@
...
+    lists:flatten(mochijson2:encode({struct, get_stats()})).
+
+-spec report_pretty_json() -> string().
+report_pretty_json() ->
+    lists:flatten(riak_moss_utils:json_pp_print(report_json())).
+
+-spec get_stats() -> [{legend, [atom()]} |
+                      {atom(), [number()]}].
+get_stats() ->
+    [{legend, [meter_count, meter_rate, latency_mean, latency_median,
```
kellymclaughlin, line 80: I'm wondering if the legend should always be included. Maybe it should be optionally enabled via some query parameter or only returned when the `Accept` header is `text/plain` since that's more likely to be a human that needs it. Thoughts?

rel/reltool.config
```diff
@@ -16,6 +16,7 @@
          webmachine,
          lager,
          poolboy,
+         folsom,
```
kellymclaughlin, line 19: Since folsom doesn't use normal otp app behavior for starting up we don't actually need this. Starting and stopping the application has no effect on its functioning. Really just an observation, doesn't matter to me if it's removed or not.

src/riak_cs_stats.erl
kellymclaughlin, line 80: Since folsom is hopefully a temporary solution, ignore my previous comment about the legend.

src/riak_moss_utils.erl
```diff
@@ -744,7 +799,9 @@ serialized_bucket_op(Bucket, ACL, User, VClock, BucketOp, RiakPid) ->
                             OpResult;
                         {ok, UpdUser} ->
                             save_user(UpdUser, VClock, RiakPid)
-                    end;
+                    end,
+                    riak_cs_stats:update(
```
kellymclaughlin: I think we only want to do this if the return from the `update_user_buckets` call is not `{ok, ignore}` since that indicates nothing was actually done. _i.e._ The bucket was already created or deleted.

slfritchie: Fixed in 5b9034c.  I also then shortened all stats function calls slightly in 35fd0eb, then discovered that the `bucket_set_acl` stat wasn't incrementing and fixed it in f875019.

kellymclaughlin:
Looks good. +1 to merge, thanks!

slfritchie:
Merged via rebase onto master and then merge --no-ff, which confused poor lil' GitHub.  Closing.

PR #190, opened 2012-04-23
Fetching info on a bucket fails.

beerriot:
Doing a GET on a bucket resource fails because we compare the bucket name in binary form against the string representation.


kellymclaughlin:
https://github.com/basho/riak_moss/blob/master/src/riak_moss_wm_bucket.erl#L166

kellymclaughlin:
+1 to merge, thanks!

PR #142, opened 2012-04-27
Improve the performance of object listing

beerriot:
Stream 2I key query results into a mapreduce job to fetch the objects and extract the metadata. This could potentially be cheaper than having all the object fetches done from moss.

### DIFF
--- a/src/riak_moss_utils.erl
+++ b/src/riak_moss_utils.erl
@@ -19,6 +19,7 @@
          get_admin_creds/0,
          get_buckets/1,
          get_keys_and_manifests/3,
+         map_keys_and_manifests/3,
          get_object/3,
          get_user/2,
          get_user_by_index/3,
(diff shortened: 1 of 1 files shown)


beerriot:
This PR improves the performance of object listing by using MapReduce to lookup object details, instead of fetching each object from the RCS side.

Some details:
- This has not yet been perf tested on Real Hardware.  On 4 laptop nodes, listing 10k objects takes 65% of the time using the MR version (11sec. vs. 17sec.).  On 4 devtestbox VM nodes, listing 10k objects takes 34% of the time using the MR version (10sec. vs. 29sec.), and listing 100k objects takes 41% of the time (2.25min. vs. 5.5min).
- Concurrent listing appears to impact listing speed exactly linearly to the number of concurrent listers.  On 4 devtestbox VM nodes, one lister listing 10k objects takes 10sec, two listers listing 10k objects each takes 20sec (both starting and finishing at the same time, ~20sec total wall clock elapsed), three listers … 30sec, six listers … 1min.  This could be simple cpu/io time sharing on my devbox, though.
- This change means that a Riak cluster _must_ have the `add_paths` configuration set up to reference `riak_moss_utils.beam` in order to list objects.
- Using MapReduce also means that a cluster must have its pipe worker cap set to ~3x the number of expected concurrent object-listing operations (1 worker for listing keys, 1 for retrieving their values, 1 for extracting the active manifest).  In operation, less that 3x will often suffice because of jitter introduced in the key listing coverage operation.
- The limit imposed by pipe workers likely means we also want to limit the number of object lists any particular RCS node might request.  One possible option is to use a separate pool of `riakc_pb_socket` workers for listing requests than the pool used for object retrieval/update requests.
- This implementation uses the filtered-keylist MapReduce input to do prefix queries because it is almost as fast as the equivalent 2i range query, but has less complexity since it doesn't have to re-filter keys to work around the fact that 2i's ranges are inclusive, not exclusive.

beerriot:
Maybe worth a little bit more explanation of how I got the numbers I mentioned above…
1. Create a bucket: `s3cmd mb s3://100ko`
2. Make a dumb file to appease s3cmd: `echo "this is a dumb file" > dummy.txt`
3. Fill the bucket: `for i in {1..100000}; do s3cmd put dummy.txt s3://100ko/$i.txt; done`
4. Test listing 10k objects: `time s3cmd ls s3://100ko/2 > /dev/null`
5. Test listing 100k objects: `time s3cmd ls s3://100k/ > /dev/null`
6. Test listing 10k object from two concurrent clients: `time s3cmd ls s3://100ko/2 > dev/null & time s3cmd ls s3://100ko/3 > /dev/null`

src/riak_moss_utils.erl
```diff
@@ -212,31 +213,68 @@ stanchion_data() ->
...
+        {ReqId, {error, Reason}} ->
+            {error, Reason}
+    after 60000 ->
+            %% timing out after complete inactivity for 1min
+            %% TODO: would shorter be better? should there be an
+            %% overall timeout?
+            {error, timeout}
+    end.
 
+%% MapReduce function, runs on the Riak nodes, should therefor use
```
reiddraper: s/therefor/therefore

src/riak_moss_utils.erl
```diff
@@ -212,31 +213,68 @@ stanchion_data() ->
...
+                         || V <- riak_object:get_values(Object) ],
+        Resolved = riak_moss_manifest_resolution:resolve(AllManifests),
+        case riak_moss_manifest:active_manifest(Resolved) of
+            {ok, Manifest} ->
+                [{riak_object:key(Object), {ok, Manifest}}];
+            _ ->
+                []
+        end
+    catch _:_ ->
+            []
```
reiddraper, line 278: What do you think about a `lager:warn` or `lager:error` here?

beerriot, line 278: Good idea. It will be most helpful in debugging configuration errors to see a nice 'undefined' in a log.

reiddraper:
+1 to merge

PR #196, opened 2012-04-27
don't bomb if the object being deleted doesn't exist

beerriot:
This case was making `s3api_verification_test.py` barf when the tests ran in a
different order than expected (because one of the tests attempted to
clean up something that shouldn't have been there).  Beyond the test
case, though, I think we don't want RCS to log an error when a user
attempts to delete something that doesn't exist.

Simple test:

```
s3cmd mb s3://deltest
s3cmd del s3://deltest/nonexistent
```

Without this patch, you'll see a parse error from s3cmd, and `{error,{badmatch,{error,notfound}},[{riak_moss_wm_key,delete_resource,2}…` in the RCS console.  With the patch, you'll just see s3cmd report `File s3://deltest/nonexistent deleted`

The `s3api_verification_test.py` script also passes once again after this patch is applied.


kellymclaughlin:
+1 to handling the other errors better later and +1 to merge

PR #153, opened 2012-04-28
Expand User API so admin user can list all users

kellymclaughlin:
These changes adds one webmachine resource, `riak_moss_wm_users` and updates another, `riak_moss_wm_user`.

The changes to `riak_moss_wm_user` to allow a user to list their own account information or the admin user to list the user information of any individual user.

`riak_moss_wm_users` allows the admin user to get a list of all users on the system. This information is gathered from riak using a streaming list keys and is streamed back to the requesting client using a multipart response. This is to avoid having to accumulate all of the user records in memory when the number of users could potentially be very large.
## Testing

I recommend the [s3-curl](http://aws.amazon.com/code/128) tool for these tests and that's what I'll use to show testing examples with. [Here's](https://gist.github.com/23cda6221da4995c15be) a sample s3-curl config file. The following testing suggestions assume that the admin user credentials are associated with the admin id in the `.s3curl` file.
1. List the cs user's information. All of the following should yield the same document containing the user account info for the user associated with the `cs` id in the `.s3curl` file.
   
   ```
   ./s3curl.pl --id cs -- --proxy1.0 "localhost:8080" -v -H "Accept: application/json"  http://s3.amazonaws.com/riak-cs/user/TMEQSQBMR59IZCTYRXQY
   ./s3curl.pl --id cs -- --proxy1.0 "localhost:8080" -v -H "Accept: application/json"  http://s3.amazonaws.com/riak-cs/user
   ./s3curl.pl --id cs -- --proxy1.0 "localhost:8080" -v -H "Accept: application/json"  http://riak-cs.s3.amazonaws.com/user/TMEQSQBMR59IZCTYRXQY
   ./s3curl.pl --id cs -- --proxy1.0 "localhost:8080" -v -H "Accept: application/json"  http://riak-cs.s3.amazonaws.com/user
   ```
2. List another user using the admin credentials. Both of the following should yield the same results and that result should be a document containing the cs user's record.
   
   ```
   ./s3curl.pl --id admin -- --proxy1.0 "localhost:8080" -v -H "Accept: application/json"  http://s3.amazonaws.com/riak-cs/user/TMEQSQBMR59IZCTYRXQY
   ./s3curl.pl --id admin -- --proxy1.0 "localhost:8080" -v -H "Accept: application/json"  http://riak-cs.s3.amazonaws.com/user/TMEQSQBMR59IZCTYRXQY
   ```
3. Fail to list another user using non-admin credentials. This assumes that the user credentials represented by the `test` id in the `.s3curl` file is distinct from the `cs` user and also not the admin for the system.
   
   ```
   ./s3curl.pl --id test -- --proxy1.0 "localhost:8080" -v -H "Accept: application/json"  http://s3.amazonaws.com/riak-cs/user/TMEQSQBMR59IZCTYRXQY
    ./s3curl.pl --id test -- --proxy1.0 "localhost:8080" -v -H "Accept: application/json"  http://riak-cs.s3.amazonaws.com/user/TMEQSQBMR59IZCTYRXQY
   ```
4. List all users using the admin credentials. Both of the following should yield the same results and that result should be a list of all user records.
   
   ```
   ./s3curl.pl --id admin -- --proxy1.0 "localhost:8080" -v -H "Accept: application/json"  http://s3.amazonaws.com/riak-cs/users
   ./s3curl.pl --id admin -- --proxy1.0 "localhost:8080" -v -H "Accept: application/json"  http://riak-cs.s3.amazonaws.com/users
   ```
5. Fail to list all users using non-admin credentials
   
   ```
   ./s3curl.pl --id cs -- --proxy1.0 "localhost:8080" -v -H "Accept: application/json"  http://s3.amazonaws.com/riak-cs/users
   ./s3curl.pl --id cs -- --proxy1.0 "localhost:8080" -v -H "Accept: application/json"  http://riak-cs.s3.amazonaws.com/users
   ```

Also repeat all of the above steps using `application/xml` as the `Accept` header value and verify that the xml results are correct as well.

### DIFF
--- a/src/riak_moss_wm_user.erl
+++ b/src/riak_moss_wm_user.erl
@@ -1,6 +1,6 @@
 %% -------------------------------------------------------------------
 %%
-%% Copyright (c) 2007-2011 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2007-2012 Basho Technologies, Inc.  All Rights Reserved.
 %%
 %% -------------------------------------------------------------------
 
@@ -8,25 +8,68 @@
 
 -export([init/1,
          service_available/2,
+         forbidden/2,
+         content_types_provided/2,
+         allowed_methods/2,
+         produce_json/2,
+         produce_xml/2,
          process_post/2,
-         allowed_methods/2]).
+         finish_request/2]).
 
 -include("riak_moss.hrl").
 -include_lib("webmachine/include/webmachine.hrl").
 
-init(_Config) ->
-    {ok, #context{}}.
+%% -------------------------------------------------------------------
+%% Webmachine callbacks
+%% -------------------------------------------------------------------
+
+init(Config) ->
+    %% Check if authentication is disabled and
+    %% set that in the context.
+    AuthBypass = proplists:get_value(auth_bypass, Config),
+    {ok, #context{auth_bypass=AuthBypass}}.
 
 -spec service_available(term(), term()) -> {true, term(), term()}.
 service_available(RD, Ctx) ->
-    {true, RD, Ctx}.
+    riak_moss_wm_utils:service_available(RD, Ctx).
 
 -spec allowed_methods(term(), term()) -> {[atom()], term(), term()}.
 allowed_methods(RD, Ctx) ->
-    {['POST'], RD, Ctx}.
+    {['GET', 'HEAD', 'POST'], RD, Ctx}.
 
+forbidden(RD, Ctx) ->
+    case wrq:method(RD) of
+        'POST' ->
+            {false, RD, Ctx};
+        _ ->
+            Next = fun(NewRD, NewCtx=#context{user=User}) ->
+                           AccessRD = riak_moss_access_logger:set_user(User, NewRD),
+                           forbidden(AccessRD, NewCtx, User)
+                   end,
+            riak_moss_wm_utils:find_and_auth_user(RD, Ctx, Next)
+    end.
 
-%% @doc Create a user from a POST
+content_types_provided(RD, Ctx) ->
+    {[{?XML_TYPE, produce_xml}, {?JSON_TYPE, produce_json}], RD, Ctx}.
+
+produce_json(RD, #context{user=User}=Ctx) ->
+    MJ = {struct, riak_moss_wm_utils:user_record_to_proplist(User)},
+    Body = mochijson2:encode(MJ),
+    Etag = etag(Body),
+    RD2 = wrq:set_resp_header("ETag", Etag, RD),
+    {Body, RD2, Ctx}.
+
+produce_xml(RD, #context{user=User}=Ctx) ->
+    XmlUserRec =
+        [{Key, [binary_to_list(Value)]} ||
+            {Key, Value} <- riak_moss_wm_utils:user_record_to_proplist(User)],
+    Doc = [{'User', XmlUserRec}],
+    Body = riak_moss_s3_response:export_xml(Doc),
+    Etag = etag(Body),
+    RD2 = wrq:set_resp_header("ETag", Etag, RD),
+    {Body, RD2, Ctx}.
+
+%% @doc Create a user from a POST.
 %%      and return the user object
 %%      as JSON
 -spec process_post(term(), term()) -> {true, term(), term}.
@@ -38,11 +81,70 @@ process_post(RD, Ctx) ->
     case riak_moss_utils:create_user(UserName, Email) of
         {ok, UserRecord} ->
             PropListUser = riak_moss_wm_utils:user_record_to_proplist(UserRecord),
-            CTypeWritten = wrq:set_resp_header("Content-Type", "application/json", RD),
+            CTypeWritten = wrq:set_resp_header("Content-Type", ?JSON_TYPE, RD),
             WrittenRD = wrq:set_resp_body(list_to_binary(
                                             mochijson2:encode(PropListUser)),
                                           CTypeWritten),
             {true, WrittenRD, Ctx};
         {error, Reason} ->
             riak_moss_s3_response:api_error(Reason, RD, Ctx)
+    end.
+
+finish_request(RD, Ctx=#context{riakc_pid=undefined}) ->
+    {true, RD, Ctx};
+finish_request(RD, Ctx=#context{riakc_pid=RiakPid}) ->
+    riak_moss_utils:close_riak_connection(RiakPid),
+    {true, RD, Ctx#context{riakc_pid=undefined}}.
+
+%% -------------------------------------------------------------------
+%% Internal functions
+%% -------------------------------------------------------------------
+
+%% @doc Calculate the etag of a response body
+etag(Body) ->
+    webmachine_util:quoted_string(
+      riak_moss_utils:binary_to_hexlist(
+        crypto:md5(Body))).
+
+forbidden(RD, Ctx, undefined) ->
+    %% anonymous access disallowed
+    riak_moss_wm_utils:deny_access(RD, Ctx);
+forbidden(RD, Ctx, User) ->
+    UserKeyId = User?MOSS_USER.key_id,
+    UserPathKey = user_key(RD),
+    case UserPathKey of
+        [] ->
+            %% user is accessing own account
+            %% @TODO Determine if logging this is appropriate
+            %% and if we need to classify it differently.
+            AccessRD = riak_moss_access_logger:set_user(User, RD),
+            {false, AccessRD, Ctx};
+         UserKeyId ->
+            %% user is accessing own account
+            %% @TODO Determine if logging this is appropriate
+            %% and if we need to classify it differently.
+            AccessRD = riak_moss_access_logger:set_user(User, RD),
+            {false, AccessRD, Ctx};
+        _ ->
+            case riak_moss_utils:get_admin_creds() of
+                {ok, {Admin, _}} when Admin == UserKeyId ->
+                    %% admin can access any account
+                    case riak_moss_utils:get_user(UserPathKey, Ctx#context.riakc_pid) of
+                        {ok, {ReqUser, _}} ->
+                            {false, RD, Ctx#context{user=ReqUser}};
+                        {error, Reason} ->
+                            _ = lager:warning("Failed to fetch user record. KeyId: ~p"
+                                          " Reason: ~p", [UserPathKey, Reason]),
+                            riak_moss_s3_response:api_error(invalid_access_key_id, RD, Ctx)
+                    end;
+                _ ->
+                    %% no one else is allowed
+                    riak_moss_wm_utils:deny_access(RD, Ctx)
+            end
+    end.
+
+user_key(RD) ->
+    case wrq:path_tokens(RD) of
+        [KeyId|_] -> mochiweb_util:unquote(KeyId);
+        _         -> []
     end.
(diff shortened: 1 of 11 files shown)


beerriot:
The `s3curl` script seems not to care, but `s3cmd get s3://riak-cs/user` complains that the response does not include an ETag.  If we expect people to use existing S3 tools to interact with that URL, we might want to add that.

beerriot:
This seems to have broken creation new users.

If I use the snippet from https://github.com/basho/riak_moss/wiki/Creating-a-User I get:

```
$ curl -X POST http://localhost:8080/user --data "email=newuser@foobar.com&name=newuser"
<?xml version="1.0" encoding="UTF-8"?><Error><Code>AccessDenied</Code><Message>Access Denied</Message><Resource>/user</Resource><RequestId></RequestId></Error>
```

If I use `s3curl`, I can create users as long as I use any existing user's credentials

```
$ ~/Downloads/s3-curl/s3curl.pl -id=regular -- --proxy1.0 "localhost:8080" -X POST http://s3.amazonaws.com/user --data "email=newuser@foobar.com&name=newuser"
{"email":"newuser@foobar.com","display_name":"newuser","name":"newuser","key_id":"JJVIM1UREQ0D9TBCATCU","key_secret":"eu5zDH_0gJ8T_WUWagLREh4AbxZCckLsFmzp8w==","id":"af5e1e41ccf4f5e9eb5031212c59668ebddd08b49f6ceac5d87b86d2019e2888"}
```

But it will fail with fake creds:

```
$ ~/Downloads/s3-curl/s3curl.pl -id=fake -- --proxy1.0 "localhost:8080" -X POST http://s3.amazonaws.com/user --data "email=newuser@foobar.com&name=newuser"
<?xml version="1.0" encoding="UTF-8"?><Error><Code>InvalidAccessKeyId</Code><Message>The AWS Access Key Id you provided does not exist in our records.</Message><Resource>/user</Resource><RequestId></RequestId></Error>
```

beerriot:
Listing a specific user using admin credentials doesn't work for me.  Instead, it just returns the admin user:

```
$ cat ~/.s3curl 
%awsSecretAccessKeys = (
    # admin account
    bogusadmin => {
        id => 'JJVIM1UREQ0D9TBCATCU',
        key => 'eu5zDH_0gJ8T_WUWagLREh4AbxZCckLsFmzp8w==',
    },
);
$ ~/Downloads/s3-curl/s3curl.pl --id bogusadmin -- --proxy1.0 "localhost:8080" http://s3.amazonaws.com/riak-cs/user/OJSQUBDITUPCBLGQEGD5
<?xml version="1.0" encoding="UTF-8"?><User><email>admin@foobar.com</email><display_name>admin</display_name><name>admin</name><key_id>JJVIM1UREQ0D9TBCATCU</key_id><key_secret>eu5zDH_0gJ8T_WUWagLREh4AbxZCckLsFmzp8w==</key_secret><id>af5e1e41ccf4f5e9eb5031212c59668ebddd08b49f6ceac5d87b86d2019e2888</id></User>
```

It will deny my access if I use non-admin creds:

```
$ cat ~/.s3curl 
%awsSecretAccessKeys = (
    # bogus account
    bogus => {
        id => 'K39JBMHDAIGY33-09HGC',
        key => 'L0smMbpFBFqHLrCReTq29l2aM-2XlEUPB9x-gg==',
    },
);
$ ~/Downloads/s3-curl/s3curl.pl --id bogus -- --proxy1.0 "localhost:8080" http://s3.amazonaws.com/riak-cs/user/OJSQUBDITUPCBLGQEGD5
<?xml version="1.0" encoding="UTF-8"?><Error><Code>AccessDenied</Code><Message>Access Denied</Message><Resource>/riak-cs/user/OJSQUBDITUPCBLGQEGD5</Resource><RequestId></RequestId></Error>
```

src/riak_moss_wm_user.erl
```diff
@@ -8,25 +8,65 @@
...
-    {['POST'], RD, Ctx}.
+    {['GET', 'POST'], RD, Ctx}.
+
+forbidden(RD, Ctx) ->
+    Next = fun(NewRD, NewCtx=#context{user=User}) ->
+                   forbidden(NewRD, NewCtx, User)
+           end,
+    riak_moss_wm_utils:find_and_auth_user(RD, Ctx, Next).
 
+content_types_provided(RD, Ctx) ->
```
beerriot: This looks like it's purposely circumventing content-type negotiation … yes?

src/riak_moss_wm_user.erl
```diff
@@ -8,25 +8,65 @@
...
+produce_json(RD, #context{user=User}=Ctx) ->
+    MJ = {struct, riak_moss_wm_utils:user_record_to_proplist(User)},
+    Body = mochijson2:encode(MJ),
+    {Body, RD, Ctx}.
 
-%% @doc Create a user from a POST
+produce_xml(RD, #context{user=User}=Ctx) ->
+    XmlUserRec =
+        [{binary_to_atom(Key, latin1), [binary_to_list(Value)]} ||
+            {Key, Value} <- riak_moss_wm_utils:user_record_to_proplist(User)],
```
beerriot: Could we maybe have `riak_moss_wm_utils:user_record_to_proplist/1` provide atom keys, so we don't have to do call `binary_to_atom/2` here?  I know the keys come from a limited set, but enforcing that implicitly would mean we wouldn't have to worry about leaking a generated binary in here, and slowly filling the atom table.

Also, a small style thing: all XML tags in the other messages RCS produces are CamelCase.  It might be a good idea to keep that theme here.  A simple `xml_key(binary()) -> camel_case_atom()` would even kill both birds with one stone.

src/riak_moss_wm_user.erl
```diff
@@ -8,25 +8,65 @@
...
+    riak_moss_wm_utils:service_available(RD, Ctx).
 
 -spec allowed_methods(term(), term()) -> {[atom()], term(), term()}.
 allowed_methods(RD, Ctx) ->
-    {['POST'], RD, Ctx}.
+    {['GET', 'POST'], RD, Ctx}.
+
+forbidden(RD, Ctx) ->
+    Next = fun(NewRD, NewCtx=#context{user=User}) ->
+                   forbidden(NewRD, NewCtx, User)
```
beerriot: If you toss in a `AccessRD = riak_moss_access_logger:set_user(User, NewRD)` here (plus one more clause in `riak_moss_access_logger:operation/1`, we can log this access as one more thing to bill.

src/riak_moss_wm_users.erl
```diff
@@ -0,0 +1,156 @@
...
+    AuthBypass = proplists:get_value(auth_bypass, Config),
+    {ok, #context{auth_bypass=AuthBypass}}.
+
+-spec service_available(term(), term()) -> {true, term(), term()}.
+service_available(RD, Ctx) ->
+    riak_moss_wm_utils:service_available(RD, Ctx).
+
+-spec allowed_methods(term(), term()) -> {[atom()], term(), term()}.
+allowed_methods(RD, Ctx) ->
+    {['GET'], RD, Ctx}.
```
beerriot: any reason not to also support HEAD?

src/riak_moss_wm_users.erl
```diff
@@ -0,0 +1,156 @@
...
+
+forbidden(RD, Ctx) ->
+    Next = fun(NewRD, NewCtx=#context{user=User}) ->
+                   forbidden(NewRD, NewCtx, User)
+           end,
+    riak_moss_wm_utils:find_and_auth_user(RD, Ctx, Next).
+
+content_types_provided(RD, Ctx) ->
```
beerriot, line 45: Same question as `riak_moss_wm_user`: why circumvent content-type negotiation?

src/riak_moss_wm_users.erl
```diff
@@ -0,0 +1,156 @@
...
+    case riakc_pb_socket:stream_list_keys(RiakPid, ?USER_BUCKET) of
+        {ok, ReqId} ->
+            wait_for_users(Format, RiakPid, ReqId, Boundary);
+        {error, _Reason} ->
+            {<<>>, done}
+    end.
+
+wait_for_users(Format, RiakPid, ReqId, Boundary) ->
+    receive
+        {ReqId, {keys, UserIds}} ->
```
beerriot, line 103: While this streaming setup avoids holding all user records in memory at one time, it does not avoid holding all user keys in memory at once.  The riak connection will not process any requests for user records until it has finished sending the all of the user keys to this process.  So, all user keys will be waiting in this process's mailbox.

This is kind of unavoidable since there's no way for this client to provide backpressure to the keylist, but a stream that didn't wait for the full key list could be set up using a second Riak client.

src/riak_moss_wm_user.erl
```diff
@@ -46,3 +84,49 @@ process_post(RD, Ctx) ->
...
+            %% @TODO Determine if logging this is appropriate
+            %% and if we need to classify it differently.
+            AccessRD = riak_moss_access_logger:set_user(User, RD),
+            {false, AccessRD, Ctx};
+         UserKeyId ->
+            %% user is accessing own account
+            %% @TODO Determine if logging this is appropriate
+            %% and if we need to classify it differently.
+            AccessRD = riak_moss_access_logger:set_user(User, RD),
+            {false, AccessRD, Ctx};
```
kellymclaughlin, line 121: Won't this call and the one similar call that follows already do what you described in a previous comment about using `set_user`. I think I'll just need to make the change to the access logger module you suggested.

beerriot, line 121: Yep, I missed this, you're absolutely right.

src/riak_moss_wm_users.erl
kellymclaughlin, line 103: Not sure I understand the problem here. Are you saying that `RiakPid` may be blocked due to the key listing and unable to get the user record until that completes? It doesn't seem like that's the case, but I want to make sure I understand what you're saying before proceeding with anymore testing.

beerriot, line 103: Yep, that's exactly what I'm saying.  If you check out [riakc_pb_socket:handle_call/3](https://github.com/basho/riak-erlang-client/blob/master/src/riakc_pb_socket.erl#L884), you'll see that when `#state.active` is not `undefined`, it just queues the request.  The [send_request/2](https://github.com/basho/riak-erlang-client/blob/master/src/riakc_pb_socket.erl#L1354) function sets `#state.active` to non-`undefined` when a request is sent, and it doesn't get cleared again until all processing of that request finishes, which for key listing is not until the whole list completes, see [process_response/3](https://github.com/basho/riak-erlang-client/blob/master/src/riakc_pb_socket.erl#L1168).

kellymclaughlin, line 103: Yes of course, it looks at the key listing as one big operation even if streamed. That's unfortunate, but I suppose I can try and grab another connection from the pool and handle it that way. Thanks for the clarification.

kellymclaughlin, line 103: Addressed by [e0bcacb](https://github.com/basho/riak_moss/commit/e0bcacb9bf3d1c0afe3830e29babe07aa87472c3)

kellymclaughlin:
[fb7eee1](https://github.com/basho/riak_moss/commit/fb7eee18aaf8085ee38441a2e0428a16923c6cbb) addresses the bug that listing a specific user using admin credentials doesn't work for me.

kellymclaughlin:
I think I've addressed all the comments and I think it's ready for re-review at this point.

beerriot:
I added a couple of small things, but +1 to merge.  Please update https://github.com/basho/riak_moss/wiki/Creating-a-User with the new CamelCase fields during merge, as I did for s3api_verification_test.py in 6018927.

PR #199, opened 2012-05-07
Slf dtrace round1

slfritchie:
Add USDT DTrace probes to RiakCS.

```
%% The convention that we'll use of these probes is:
%%   param  D arg name  use
%%   -----  ----------  ---
%%   --     arg0         Erlang PID
%%   --     arg1         PID's user-defined tag (if any)
%%   i1     arg2        Application category (see below)
%%   i2     arg3        1 = function entry, 2 = function return
%%                      NOTE! Not all function entry probes have a return probe
%%   i3-i4  arg4-arg5   Varies, zero usually means unused (but not always!)
%%   s1     arg6        Module name
%%   s2     arg7        Function name
%%   s3-4   arg8-arg9   Varies, NULL means unused
```

Here's an example, using R14B04:
For example, running a D script such as this (as the superuser):

```
dtrace -qn 'erlang*:::user_trace* /arg2 == 703/ {printf("pid %s: mod %s op %s: user %s bucket/file %s\n", copyinstr(arg0), copyinstr(arg6), copyinstr(arg7), copyinstr(arg8), copyinstr(arg9));}'
```

... yields this output when using `s3cmd get s3://decision/foo`:

```
pid <0.790.0>: mod riak_moss_wm_key op file_get: user foo bar bucket/file decision,foo
pid <0.790.0>: mod riak_moss_wm_key op file_get: user foo bar bucket/file decision,foo
```

### DIFF
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -253,5 +253,5 @@
 -define(DT_SERVICE_OP,      701).
 -define(DT_BUCKET_OP,       702).
 -define(DT_OBJECT_OP,       703).
--define(DT_AUTH_OP,         704).
+%%-define(DT_AUTH_OP,         704).
 -define(DT_WM_OP,           705).
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -213,3 +213,11 @@
 %% to determine the PUT buffer size.
 %% ex. 2 would mean BlockSize * 2
 -define(DEFAULT_PUT_BUFFER_FACTOR, 1).
+
+%% Major categories of Erlang-triggered DTrace probes
+-define(DT_BLOCK_OP,        1).
+-define(DT_SERVICE_OP,      2).
+-define(DT_BUCKET_OP,       3).
+-define(DT_OBJECT_OP,       4).
+-define(DT_AUTH_OP,         5).
+-define(DT_WM_OP,           6).
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -215,9 +215,43 @@
 -define(DEFAULT_PUT_BUFFER_FACTOR, 1).
 
 %% Major categories of Erlang-triggered DTrace probes
--define(DT_BLOCK_OP,        1).
--define(DT_SERVICE_OP,      2).
--define(DT_BUCKET_OP,       3).
--define(DT_OBJECT_OP,       4).
--define(DT_AUTH_OP,         5).
--define(DT_WM_OP,           6).
+%%
+%% The main R15B01 USDT probe that can be triggered by Erlang code is defined
+%% like this:
+%%
+%% /**
+%%  * Multi-purpose probe: up to 4 NUL-terminated strings and 4
+%%  * 64-bit integer arguments.
+%%  *
+%%  * @param proc, the PID (string form) of the sending process
+%%  * @param user_tag, the user tag of the sender
+%%  * @param i1, integer
+%%  * @param i2, integer
+%%  * @param i3, integer
+%%  * @param i4, integer
+%%  * @param s1, string/iolist. D's arg6 is NULL if not given by Erlang
+%%  * @param s2, string/iolist. D's arg7 is NULL if not given by Erlang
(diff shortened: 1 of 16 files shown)


dizzyd:
Complete and utter +1 on the idea; no code review was part of this message.

On Sun, May 6, 2012 at 11:16 PM, Scott Lystig Fritchie
reply@reply.github.com
wrote:

slfritchie:
NOTE: I've written things intentionally so that each dtrace-message goes through a module-local function before jumping into DTrace-land, so that it will hopefully be easier to scrape the same trace events using Redbug & other Erlang function tracing tools.

One thing that I haven't figured out is if there's a place for the per-process user-defined tag in this scheme.  If RCS isn't going to be using that tag for anything else, then perhaps user and/or bucket/bucket+file info could use that tag?  (or would that paint us into a corner that wouldn't allow us to use that tag for something else later?)

slfritchie:
Reviewer(s): I'd appreciate your comments on whether or not it's valuable to have _all_ of the WM callbacks annotated.  The ones such as `content_types_provided` aren't doing that much interesting computation, and part of me thinks that they're not as useful as the probes that fire during more "interesting" WM callbacks.  (IMO we still want the `init()` probe, so that we can measure end-to-end timing of WM operations.)

kellymclaughlin:
I see a few hoops for R14B04. What if we simplify life and say that cs will only be built with R15B01 or later. We're not bound to Riak's build requirements. I've been building and running with R15 since it came out and haven't had any issues. Just a thought to make things a little cleaner.

slfritchie:
Yes, well, I've a bit of an agenda ... namely to put the same low-level code into Riak, where R14B04 is bound to live for a while longer.

include/riak_moss.hrl
```diff
@@ -213,3 +213,45 @@
...
+%%   i3-i4  arg4-arg5   Varies, zero usually means unused (but not always!)
+%%   s1     arg6        Module name
+%%   s2     arg7        Function name
+%%   s3-4   arg8-arg9   Varies, NULL means unused
+%%
+-define(DT_BLOCK_OP,        700).
+-define(DT_SERVICE_OP,      701).
+-define(DT_BUCKET_OP,       702).
+-define(DT_OBJECT_OP,       703).
+%%-define(DT_AUTH_OP,         704).
```
kellymclaughlin, line 256: Is this a a `TODO`?

kellymclaughlin:
Could we factor out the various flavors of `dt_entry*` functions into the `riak_cs_dtrace` module. Seems like it could DRY out the code at the expense of an extra param to specify the calling module.

include/riak_moss.hrl
slfritchie, line 256: It's a need-to-avoid-spamminess-on-a-single-DTrace-probe thing.  All of the auth decisions are already being logged by WM forbidden() callbacks.  If you think that it'd really be useful to have separate auth op probes, e.g. log different info than the WM forbidden() callbacks are logging?

kellymclaughlin, line 256: I was more wondering if that line was a placeholder for later or could just be deleted.

slfritchie:
> Could we factor out the various flavors of [...]

The reason for the repetition is to make it easier for Erlang tracing users to get the same info as DTrace/SystemTap users can.  It's much more convenient (and easier to remember the syntax) to use `redbug:start({riak_moss_wm_key, dt_entry})` or `redbug:start("riak_moss_wm_key:dt_entry")` than it is to use `redbug:start({riak_cs_dtrace, dtrace, ['_', 2, '_', riak_moss_wm_key, '_', '_']})` or `redbug:start("riak_cs_dtrace:dtrace('_', 2, '_', riak_moss_wm_key, '_', '_')`.

kellymclaughlin:
Ok sounds like a good reason to me.

include/riak_moss.hrl
slfritchie, line 256: I'm thinking placeholder, especially if/when a decent solution arrives for the one-probe-to-rule-them-all problem.

kellymclaughlin:
Good stuff! +1 to merge.

PR #201, opened 2012-05-11
Separate connection pool for bucket listings

kellymclaughlin:
- Add an independent connection pool only for bucket listings and limit it by default to 5.
- Update `riak_moss_wm_bucket` to use this new pool.
- Add functions so that a pool may be specified when requesting a connection or returning one to a pool.
- Update supervisor to create specifications for any connection pool specified in the application environment.


reiddraper:
+1 to merge, nice work

PR #206, opened 2012-05-29
Support query parameter authentication

kellymclaughlin:
This is a quick turnaround pr to support some testing at a potential customer. It adds support for the query parameter authentication described [here](http://docs.amazonwebservices.com/AmazonS3/latest/dev/RESTAuthentication.html). 
## Testing

I added a temporary exported function, `riak_moss_s3_auth:calculate_qs_signature,` to calculate a request signature that can be used in the `Signature` query parameter. The first parameter is the key secret, followed by the HTTP verb, the resource (including bucket), the expiry time in seconds (as a string), and an optional list of headers.

Use:

```
(riak-cs@127.0.0.1)34> riak_moss_s3_auth:calculate_qs_signature("EEZ4kukvdC7qBtULVcnKh022VNIo-kVpYSB7Qw==", "GET", "/bucket/a", "1338324923", []).             
"V2yY4W13ZPpvkOcA1Eqt2urLajE%3D"
```

Then using the resulting signature in a request with curl:

```
curl --proxy1.0 "localhost:8080" "http://s3.amazonaws.com/bucket/a?AWSAccessKeyId=JYIGUADVDOHPDCYGP4K1&Expires=1338324923&Signature=V2yY4W13ZPpvkOcA1Eqt2urLajE%3D" 
```

_Note:_ For creating buckets when testing this remember to checkout the 1.0 branch of stanchion for the `riak-cs` dep and the standalone repo used for running the stanchion app.

### DIFF
--- a/RELEASE-NOTES.org
+++ b/RELEASE-NOTES.org
@@ -1,3 +1,6 @@
+* MOSS 1.0.2 Release Notes
+** Additions
+- Support query parameter authentication as specified in [[http://docs.amazonwebservices.com/AmazonS3/latest/dev/RESTAuthentication.html][Signing and Authenticating REST Requests]].
 * MOSS 1.0.1 Release Notes
 ** Bugs Fixed
 - Default content-type is not passed into function to handle PUT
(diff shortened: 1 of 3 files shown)


reiddraper:
+1, nice work. Tested with boto, a python s3 client, the URLs it generates work, and expire appropriately.

PR #203, opened 2012-06-01
Check for timeout when calling poolboy:checkout

kellymclaughlin:
If Riak is responding very slowly then the synchronous call `poolboy:checkout` can time out. Poolboy calls `gen_fsm:sync_send_state` with a 5 second timeout and our pool worker function calls `riakc_pb_socket:start_link` with the default timeout of infinity. Check for such a timeout in riak_moss_utils:riak_connection` so the condition can be handled gracefully.

Add a catch in `riak_moss_utils:riak_connection/1` to handle cases where the call to `poolboy:checkout/2` times out or otherwise throws an error. The error is handled gracefully and an error code is returned to the client that indicates the request should be retried, but that the request rate is too high.

### DIFF
--- a/src/riak_moss_riakc_pool_worker.erl
+++ b/src/riak_moss_riakc_pool_worker.erl
@@ -29,7 +29,7 @@ riak_host_port() ->
 
 start_link(_Args) ->
     {Host, Port} = riak_host_port(),
-    riakc_pb_socket:start_link(Host, Port).
+    riakc_pb_socket:start_link(Host, Port, [{connect_timeout, 10000}]).
 
 stop(Worker) ->
     riakc_pb_socket:stop(Worker).
--- a/src/riak_moss_utils.erl
+++ b/src/riak_moss_utils.erl
@@ -471,9 +471,11 @@ riak_connection() ->
 %% from the specified connection pool.
 -spec riak_connection(atom()) -> {ok, pid()} | {error, term()}.
 riak_connection(Pool) ->
-    case poolboy:checkout(Pool, false) of
+    case catch poolboy:checkout(Pool, false) of
         full ->
             {error, all_workers_busy};
+        {'EXIT', _Error} ->
+            {error, poolboy_error};
         Worker ->
             {ok, Worker}
     end.


reiddraper:
You think it's also worth (possibly separately) changing the `riakc_pb_socket:start_link` call to _not_ use an `infinity` timeout? Though fwiw, [it may be that we never wait for infinity anyway](http://erlang.org/pipermail/erlang-questions/2009-April/043002.html). My testing on OSX has this effect as well (though 60 seconds). If we _do_ end up waiting for infinity though, are there any downsides?

kellymclaughlin:
Probably a decent idea. I suppose if the call did block forever it could result in `poolboy` blocking in the `init` function forever or each attempt to start a new worker, in the case of worker death, being blocked forever. 60 seconds seems like the standard default, but that seems extremely long for this case. Perhaps somewhere in the 10-30 second range?

reiddraper:
> Perhaps somewhere in the 10-30 second range?

yeah 10 seconds seems good to me

Vagabond:
Have you considered using https://github.com/basho/poolboy/blob/master/src/poolboy.erl#L37 with 'infinity' as the timeout instead?

reiddraper:
@Vagabond we want to return `503` to users in a timely manner if a connection from the pool isn't available in a few seconds. Otherwise we could get a massive queue of requests all waiting for a connection.

Vagabond:
Ah ok, that makes sense. At least if you want a different timeout, you can set one.

kellymclaughlin:
FYI ,the impetus for this change stems from observations of calls to `pooboy:checkout` timing out after 5 seconds in production. On Friday a similar thing was happening with a Riak node as well. The `poolboy` fsm seemed to go unresponsive for some reason and even status calls to poolboy were timing out. I spent some time looking at the poolboy code yesterday, but I didn't come across any issues that would seem to cause that. I suppose the fsm could be extremely busy, but that didn't appear to be the case from the Riak error logs I looked at. The message queue ranged anywhere from empty to at most 8 messages when the timeouts happened. Any ideas or thoughts as to the source of this? My plan now is just to watch for it to happen again and try and gather more info to help the diagnosis.

reiddraper:
> Hmm, timeout does not seem to be on of the client options accepted for the third parameter of `riakc_pb_socket:start_link`. Am I missing something or were you thinking of a timeout elsewhere?

I was looking at `connect_timeout`, in the `Options` proplist [here](https://github.com/basho/riak-erlang-client/blob/master/src/riakc_pb_socket.erl#L990).

reiddraper:
@kellymclaughlin for the poolboy stalls in Riak, is it possible that it's the same issue we found with Riak CS where poolboy is blocking in the `init` call of starting a worker?

reiddraper:
Also just confirmed this works for connection timeouts:

``` erlang
riakc_pb_socket:start_link("localhost", 8000, [{connect_timeout, 5000}]).
```

reiddraper:
+1 to merge

PR #207, opened 2012-06-05
Add ping resource for Riak CS.

kellymclaughlin:
Add a webmachine resource for health checks of Riak CS. The resource determines success of the ping request by attempting to acquire a Riak protocol buffers connection and then calling the `riakc_pb_socket:ping/2` function via that connection. If the Riak ping is successful a `200` is returned with 'OK' as the response body; otherwise, a `503` error response code is returned.

The resource first attempts to obtain a connection from the normal request pool and if that request times out then a new connection to Riak is tried outside of the connection pool. Failure to make that connection results in an error response from the ping request. The reason for the two step process is to balance between having the result of the ping be meaningful (_i.e._ A failure really does indicate something is wrong) and being wise with system resource usage in the face of many concurrent or frequent ping requests (_e.g._ Exhausting all available sockets on a machine.).

The resource resides at `/riak-cs/ping` and can be tested on a local setup using curl as follows:

```
curl localhost:8080/riak-cs/ping -v
```

### DIFF
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -225,7 +225,7 @@
 %% to determine the PUT buffer size.
 %% ex. 2 would mean BlockSize * 2
 -define(DEFAULT_PUT_BUFFER_FACTOR, 1).
-
+-define(DEFAULT_PING_TIMEOUT, 5000).
 -define(JSON_TYPE, "application/json").
 -define(XML_TYPE, "application/xml").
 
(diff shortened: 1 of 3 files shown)


kellymclaughlin:
Let me know what you think of [this](https://github.com/basho/riak_moss/commit/10b293b0cb80b0dbd59cffda24e130bbb6827d98).

reiddraper:
+1 to merge

PR #208, opened 2012-06-06
key_context record referenced in dt_return call in riak_moss_wm_user

kellymclaughlin:
Remove an inappropriate reference to the `key_context` record in forbidden function in the `riak_moss_wm_user` module.


cmeiklejohn:
+1 this resolves the 500 server error when retrieving user information.

PR #186, opened 2012-06-07
Support user-specified metadata headers

kellymclaughlin:
Allow users to associate arbitrary metadata headers with objects using `x-amz-meta-` headers.

I cherry-picked Marcel's original commit and made a few changes and did some refactoring to reduce code duplication.

The original branch Marcel created is [here](https://github.com/basho/riak_moss/tree/user-metadata-handling). 
## Testing

Headers can be specified with `s3cmd` using the `--add-header=NAME:VALUE` option. 

_e.g._ `--add-header=x-amz-meta-cheese=feta` 

Using the `-d` option when fetching a file dumps the headers so the presence of the metadata can be verified.

### DIFF
--- a/src/riak_moss_wm_utils.erl
+++ b/src/riak_moss_wm_utils.erl
@@ -20,6 +20,8 @@
          validate_auth_header/3,
          deny_access/2,
          extract_name/1,
+         normalize_headers/1,
+         extract_amazon_headers/1,
          extract_user_metadata/1]).
 
 -include("riak_moss.hrl").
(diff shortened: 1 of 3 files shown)


src/riak_moss_wm_utils.erl
```diff
@@ -278,10 +281,62 @@ extract_name(?MOSS_USER{name=Name}) ->
 extract_name(_) ->
     "-unknown-".
 
+extract_amazon_headers(Headers) ->
+    FilterFun =
+        fun({K, V}, Acc) ->
+                case lists:prefix("x-amz-", K) of
+                    true ->
+                        V2 = unicode:characters_to_binary(V, utf8),
```
reiddraper, line 291: As I understand, HTTP headers must be US-ASCII. UTF-8 is a superset of that so it will work fine, but do we have a failing test case where we'd need utf8 to work within HTTP / expected s3 behavior?

reiddraper:
ps. i rebased this off of master to test it. I was getting badrecord errors from the user record before this. Probably because I was using master stanchion?

reiddraper:
+1

PR #149, opened 2012-06-08
Expand User API to be able to disable a user

kellymclaughlin:
Update `riak-cs` and `stanchion` to allow users to be disabled.

With this change a user may use a `PUT` to `/riak-cs/user` to disabled their account. The put must include a document indicating a status of _disabled_. JSON or XML formats are supported for this document. Samples of each are shown below. The `Content-Type` header should also be set appropriately. The admin user may also disable or reenable a user's account via a `PUT` to `/riak-cs/user/<user-key-id>`. Users may not reenable their own account once it is disabled.
### Sample JSON status update document

```
{
  "status" : "enabled"
}
```

The `users` resource for admin users to list all users has also been updated to list only enabled accounts by default. Disabled account can be included in the results by using a query parameter of `?disabled=yes` or `?disabled=true`. 
### Sample XML status update document

```
<UserUpdate>
  <Status>disabled</Status>
</UserUpdate>
```

**NOTE:** This change also includes [this stanchion pr](https://github.com/basho/stanchion/pull/15). The two should be reviewed and tested together.
### Testing

I suggest creating three user account and setting up one as admin. Save update documents as `user.json` and `user.xml` respectively. Set up a _.s3curl_ file with the credentials for the three users. Create a bucket and place some objects in it with one of the non-admin users. Then disable the user. An example of doing this with s3curl is as follows:
`s3curl --id cs --put=user.json --contentType=application/json -- -v --proxy1.0 "localhost:8080" http://s3.amazonaws.com/riak-cs/user`

At this point access to the bucket and objects in that bucket should be disallowed for the user. The next step is to reenable the user using the admin account.
`s3curl --id admin --put=user.json --contentType=application/json -- -v --proxy1.0 "localhost:8080" http://s3.amazonaws.com/riak-cs/user/NO-YQPYEKKUN-KJNPIGO`

Next disable the non-admin user account again and list all users and verify that the disabled user does not show up.
`s3curl --id admin -- --proxy1.0 "localhost:8080" -v http://s3.amazonaws.com/riak-cs/users`

Then use the `disabled` query parameter to display the user accounts including the disabled user.
`s3curl --id admin -- --proxy1.0 "localhost:8080" -v http://s3.amazonaws.com/riak-cs/users?disabled=true`

Next reenable the user and use ACL manipulation to grant access to the some buckets or objects to the third and so far unused user account. Verify the third user can appropriately access said buckets or objects and then disable the user that owns them. The third user should now also be denied access to the disabled user's buckets and objects.

### DIFF
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -12,7 +12,18 @@
           key_secret :: string(),
           canonical_id :: string(),
           buckets=[] :: [moss_bucket()]}).
--type moss_user() :: #moss_user_v1{}.
+
+-record(rcs_user_v2, {
+          name :: string(),
+          display_name :: string(),
+          email :: string(),
+          key_id :: string(),
+          key_secret :: string(),
+          canonical_id :: string(),
+          buckets=[] :: [moss_bucket()],
+          status=enabled :: enabled | disabled}).
+-type moss_user() :: #rcs_user_v2{} | #moss_user_v1{}.
+-type rcs_user() :: #rcs_user_v2{} | #moss_user_v1{}.
 
 -record(moss_bucket, {
           name :: string(),
@@ -189,7 +200,8 @@
 
 -define(ACL, #acl_v2).
 -define(MOSS_BUCKET, #moss_bucket_v1).
--define(MOSS_USER, #moss_user_v1).
+-define(MOSS_USER, #rcs_user_v2).
+-define(RCS_USER, #rcs_user_v2).
 -define(USER_BUCKET, <<"moss.users">>).
 -define(ACCESS_BUCKET, <<"moss.access">>).
 -define(STORAGE_BUCKET, <<"moss.storage">>).
--- a/src/riak_moss_s3_response.erl
+++ b/src/riak_moss_s3_response.erl
@@ -33,6 +33,8 @@ error_message(user_already_exists) ->
     "The specified email address has already been registered. Email addresses must be unique among all users of the system. Please try again with a different email address.";
 error_message(entity_too_large) ->
     "Your proposed upload exceeds the maximum allowed object size.";
+error_message(invalid_user_update) ->
+    "The user update you request was invalid.";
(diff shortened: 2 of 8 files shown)


cmeiklejohn:
@kellymclaughlin, when I use the API to retrieve the user record, shoudn't it return the disabled/enabled status as well?

cmeiklejohn:
Few notes from the testing I've done so far:
- Updating a user record returns a 204 No Content; a PUT object request to Amazon should return 200 to be compliant.
- Status is not returned when you request the list of all users -- this would be great to have included to know which users are disabled vs. enabled.
- Can the default be to return all users, and the parameter used to override?  Is there a reason we're doing it otherwise?
- Same problem with CGI escaping in the URL for the user resource as #209.

cmeiklejohn:
> This begs the question if we should continue to follow the Amazon way in
> all cases even in the User Management API or should we take the liberty to
> draw a box around the storage API and go our own way outside of that box.
> Not sure about the exact right answer here so if you have thoughts I'm
> interested to hear.

Returning status codes that are in line with what Amazon is returning for
object storage and retrieval will allow for greater support across client
implementations of the S3 API.  I feel that until we have a compelling
reason it might make sense to go this route.  Thoughts?

> - Can the default be to return all users, and the parameter used to
>   override?  Is there a reason we're doing it otherwise?
>   So you would prefer default to showing even disabled accounts and only
>   exclude them with an indication via the query parameter?

I think it makes more sense for /users to provide the full unfiltered
listing, and the URL parameter should be used to apply filters such as
disabled=false, etc.  I can see users being confused when geting a
duplicate account error when creating someone and then hitting /users only
to see the user not there.

kellymclaughlin:
@cmeiklejohn Pushed two commits to change listing users to default to all users and check for the status query parameter and also return `200` on successful `PUT` to `/riak-cs/user`.

cmeiklejohn:
+1, thanks @kellymclaughlin

PR #211, opened 2012-06-13
Configuration option to control user creation

kellymclaughlin:
Add configuration option to allow system installers to choose to restrict user creation to admins or permit anyone to create a user directly.

The configuration option, `anonymous_user_creation`, can be placed in the `riak_moss` section of the app.config. The default value is `false` is no value is specified.
## Testing

With `anonymous_user_creation` set to `true`, user creation should behave as it always has up to this point. Any user should be able to issue a POST request using curl with no authentication required.

With `anonymous_user_creation` set to `false`, anonymous POST requests to `/riak-cs/user` should result in a `403` error. Authenticated POST requests from non-admin user accounts should also receive a `403` in response. Finally, only a POST request from the admin user account should be successful in creating the new user account. 

Using _s3curl_ a simple way to issue the POST would look as follows:

```
s3curl --id cs --post newuser.json --contentType application/json -- --proxy1.0 "localhost:8080" -v http://s3.amazonaws.com/riak-cs/user
```

where _newuser.json_ contains the following:

```
{"email":"test1@test.com", "name":"test1"}
```

### DIFF
--- a/src/riak_moss_utils.erl
+++ b/src/riak_moss_utils.erl
@@ -9,7 +9,8 @@
 -module(riak_moss_utils).
 
 %% Public API
--export([binary_to_hexlist/1,
+-export([anonymous_user_creation/0,
+         binary_to_hexlist/1,
          close_riak_connection/1,
          close_riak_connection/2,
          create_bucket/5,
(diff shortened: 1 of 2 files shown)


cmeiklejohn:
@kellymclaughlin I know it's really awful, but could this support PUT as well?  This would allow you to create users with s3cmd and the like.  Yes, it's gross, I know.  Thoughts?

kellymclaughlin:
It is pretty awful, but it would be a pretty minor change to make it work. If you think it would be very useful I can add that in.

cmeiklejohn:
@kellymclaughlin I'm going to turn the anonymous creation on for now to keep moving forward, but let's plan to mumble about it on Monday.

cmeiklejohn:
+1, other than the aforementioned objection.

cmeiklejohn:
+1 merge it!

PR #152, opened 2012-06-15
Expand User API to be able to issue new credentials to a user

kellymclaughlin:
Changes to allow users to change generate a new `key_secret` for their account.  Return updated user document in the response body. Also allow admin users to generate a new `key_secret` for a specified user account.
## Testing

The `key_secret` for a user account can be reissued by a `PUT` to `/riak-cs/user` with the appropriate JSON or XML document. For admin users the `PUT` would be to `/riak-cs/user/<key-id>`. The documents should look as follows:

JSON

```
{"new_key_secret":true}
```

XML

```
<UserUpdate>
  <NewKeySecret>true</NewKeySecret>
</UserUpdate>
```

**NOTE** The `new_key_secret` field may be combined with other user update fields in the same request. Currently, the only other supported field is `status`, but more may be added in the future. Unsupported fields are simply ignored.

### DIFF
--- a/src/riak_moss_s3_response.erl
+++ b/src/riak_moss_s3_response.erl
@@ -78,7 +78,11 @@ status_code(_) -> 503.
 
 
 respond(StatusCode, Body, ReqData, Ctx) ->
-    {{halt, StatusCode}, wrq:set_resp_body(Body, ReqData), Ctx}.
+     UpdReqData = wrq:set_resp_body(Body,
+                                    wrq:set_resp_header("Content-Type",
+                                                        ?XML_TYPE,
+                                                        ReqData)),
+    {{halt, StatusCode}, UpdReqData, Ctx}.
 
 api_error(Error, ReqData, Ctx) when is_atom(Error); is_tuple(Error) ->
     error_response(status_code(Error), error_code(Error), error_message(Error),
--- a/src/riak_moss_utils.erl
+++ b/src/riak_moss_utils.erl
@@ -22,6 +22,7 @@
          get_buckets/1,
          get_env/3,
          get_keys_and_manifests/3,
+         is_admin/1,
          map_keys_and_manifests/3,
          get_object/3,
          get_user/2,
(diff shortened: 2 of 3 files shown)


cmeiklejohn:
@kellymclaughlin I'm running into issues when creating users on this branch.  However, retrieving a user by key_id is working fine.

```
      Expected([201]) <=> Actual(403 Forbidden)
        request => {:connect_timeout=>60, :headers=>{"Content-Type"=>"application/json", "Host"=>"localhost:8080", "Content-Length"=>55}, :instrumentor_name=>"excon", :mock=>false, :read_timeout=>60, :retry_limit=>4, :ssl_ca_file=>"/Users/cmeiklejohn/.rbenv/versions/1.9.3-p194/lib/ruby/gems/1.9.1/gems/excon-0.13.4/data/cacert.pem", :ssl_verify_peer=>true, :write_timeout=>60, :host=>"localhost", :path=>"/user", :port=>"8080", :query=>nil, :scheme=>"http", :expects=>[201], :method=>"POST", :body=>"{\"email\":\"user-1339816031-4@example.com\",\"name\":\"User\"}"}
```

cmeiklejohn:
Confirmed switching back to master resolves the problem.

kellymclaughlin:
Hmm, don't know that I tried creating a user on that branch. I'll check it out and see if I have similar issues.

cmeiklejohn:
thanks @kellymclaughlin

kellymclaughlin:
@cmeiklejohn Oh yes, I forgot to mention that I based this branch on the changes to restrict user creation. Checkout [this](https://github.com/basho/riak_moss/pull/211) pull request for more details. You can either create a user using the admin creds by default of set `anonymous_user_creation` to `true`.

cmeiklejohn:
@kellymclaughlin see comment on other pull request.

cmeiklejohn:
+1, added support to fog as well.

cmeiklejohn:
+1 revisions

PR #219, opened 2012-06-22
Update python integration tests with most recent Riak CS changes

reiddraper:
- Submit user creation as a json body
- Change format of some strings, ie. KeyID -> key_id
- Remove unused urllib2 import


kellymclaughlin:
+1 to merge

PR #147, opened 2012-06-27
GC for deleted and overwritten files

kellymclaughlin:
Changes to implement a garbage collection system for Riak CS. A detailed description how garbage collection works can be found [here](https://github.com/basho/riak_moss/wiki/Object-Chunking-and-Garbage-Collection). 
## Testing

A number of new tests have been written, but test coverage is far from complete. A number of other tests are planned, but due to the size of change set we have decided to move ahead with the pull request and add further tests as we prepare for the coming release. 

The default `leeway` time for an object version's blocks to be reaped is 24 hours and the default interval that the GC daemon wakes up to look for a batch of work is 15 minutes. For testing and manual inspection it maybe easier to set these values lower. Here is an example `app.config` snippet we have been using for testing:

```
%% garbage collection related                                                                                                                                          
{gc_interval, 30},
{leeway_seconds, 30},
{gc_retry_interval, 5},
```

Also the default bucket properties for Riak should be correctly set such that all buckets are created with `allow_mult` set to `true`. Siblings are expected and handled.

The `1.2.1` tag of the protobuffs clients should be used both for Riak CS and Stanchion. There are some incompatibilities that need to be addressed with recent changes to that client, but since our biggest user may upgrade to the next release of Riak CS before upgrading to Riak 1.2.0, we are sticking with the `1.2.1` tag. Reid has made changes to lock the `rebar.config` file to this tag and those changes are merged to `master`, but this branch has not been rebased from `master` since that merge.

There remain a few dialyzer error or warning reports. I found that many of these were related to the fact that the pb client has many bad specs which in turn cause dialyzer reports for Riak CS code. Thankfully Sean has fixed some of these at least on `master` already.


src/riak_moss_manifest_resolution.erl
```diff
@@ -43,8 +43,8 @@ resolve_dicts(A, B) ->
 %% @private
 -spec resolve_manifests(term(), term(), term()) -> term().
 resolve_manifests(_Key, A, B) ->
-    AState = A#lfs_manifest_v2.state,
-    BState = B#lfs_manifest_v2.state,
+    AState = A?MANIFEST.state,
```
slfritchie: https://github.com/basho/riak_moss/pull/224

Simply & shrink code: use integers to help compare states that we know "follow" each other....

slfritchie:
Reminder for Kelly: search & destroy the manifest record `deleted` state.  Perhaps start with `egrep '\bdeleted\b' src/*` ?

... although the `deleted` state does appear to be used inside of `riak_kv_delete_fsm.erl`.  But that state should not "leak" into any version of a manifest record that is stored in KV, correct?

src/riak_cs_delete_fsm.erl
```diff
@@ -0,0 +1,213 @@
...
+         terminate/3,
+         code_change/4]).
+
+-record(state, {bucket :: binary(),
+                key :: binary(),
+                uuid :: binary(),
+                manifest :: lfs_manifest(),
+                riakc_pid :: pid(),
+                delete_blocks_remaining :: ordsets:ordset(integer()),
+                unacked_deletes=ordsets:new(),
```
slfritchie: Same type as `delete_blocks_remaining`?  The current value isn't a valid type.  Ah, it's a valid assignment, but then it's type is only implicit, silly me.

src/riak_cs_delete_fsm.erl
```diff
@@ -0,0 +1,213 @@
...
+%% ====================================================================
+
+-spec handle_receiving_manifest(state()) ->
+    {next_state, atom(), state()}.
+handle_receiving_manifest(State=#state{riakc_pid=RiakcPid,
+                                       manifest=Manifest}) ->
+    {NewManifest, BlocksToDelete} = blocks_to_delete_from_manifest(Manifest),
+
+    NewState = State#state{manifest=NewManifest,
+                           unacked_deletes=[],
```
slfritchie: `ordsets:new()` please, or add the initializer into the `#state` record definition or into `init()`?

src/riak_cs_delete_fsm.erl
```diff
@@ -0,0 +1,213 @@
...
+    {next_state, StateName, State}.
+
+handle_sync_event(_Event, _From, StateName, State) ->
+    {next_state, StateName, State}.
+
+handle_info(_Info, StateName, State) ->
+    {next_state, StateName, State}.
+
+terminate(normal, _StateName, _State) ->
+    gen_fsm:sync_send_event(riak_cs_gc_d, {self(), ok}, 5000),
```
slfritchie: Well, I guess that we're in the `terminate()` func, so if a timeout happens ... then there'd only be a SASL error report (I think) and any code following (none right now but might be something in the future) wouldn't get executed.  Either `catch` this (and the other one on line 117), or adjust the timeout to `infinity`?

src/riak_cs_delete_fsm.erl
```diff
@@ -0,0 +1,213 @@
...
+         code_change/4]).
+
+-record(state, {bucket :: binary(),
+                key :: binary(),
+                uuid :: binary(),
+                manifest :: lfs_manifest(),
+                riakc_pid :: pid(),
+                delete_blocks_remaining :: ordsets:ordset(integer()),
+                unacked_deletes=ordsets:new(),
+                all_delete_workers :: list(pid()),
```
slfritchie: Nobody appears to be freeing/unallocating these workers?  Probably because they're linked to us, and they're going to die when we die.  Except that we can die with 'normal' status, which won't trigger link death .... er, except that the block server workers are setting trap_exit, which means they won't die automatically, and I don't see them handling 'EXIT' messages anywhere in that module.

src/riak_moss_block_server.erl
```diff
@@ -49,6 +50,36 @@ start_link() ->
...
+-spec start_block_servers(pid(), pos_integer()) -> [pid()].
+start_block_servers(RiakcPid, 1) ->
+    {ok, Pid} = start_link(RiakcPid),
+    [Pid];
+start_block_servers(RiakcPid, MaxNumServers) ->
+    case start_link() of
+        {ok, Pid} ->
+            [Pid | start_block_servers(RiakcPid, (MaxNumServers - 1))];
+        {error, normal} ->
+            start_block_servers(RiakcPid, 1)
```
slfritchie, line 80: Hrm, what will prevent the server from returning not-ok again?  The first clause doesn't account for that, so the badmatch will explode in the caller's face.  Is that explosion acceptable?

src/riak_cs_delete_fsm.erl
```diff
@@ -0,0 +1,213 @@
...
+        false ->
+            _ = finish(NewState),
+            {stop, normal, NewState}
+    end.
+
+maybe_delete_blocks(State=#state{free_deleters=[]}) ->
+    State;
+maybe_delete_blocks(State=#state{delete_blocks_remaining=[]}) ->
+    State;
+maybe_delete_blocks(State=#state{bucket=Bucket,
```
slfritchie, line 164: To be picky, all of the [_|_] pattern matches here aren't kosher if the `ordsets` data type were truly opaque.  ... Though to honor opacity, the code (inevitably?) gets more verbose and case'y/if'y, hrm.

src/riak_cs_delete_fsm.erl
```diff
@@ -0,0 +1,213 @@
...
+    NewDeleteBlocksRemaining = ordsets:del_element(BlockID, DeleteBlocksRemaining),
+    _ = lager:debug("Deleting block: ~p ~p ~p ~p", [Bucket, Key, UUID, BlockID]),
+    riak_moss_block_server:delete_block(DeleterPid, Bucket, Key, UUID, BlockID),
+    NewFreeDeleters = ordsets:del_element(DeleterPid, FreeDeleters),
+    maybe_delete_blocks(State#state{unacked_deletes=NewUnackedDeletes,
+                                    free_deleters=NewFreeDeleters,
+                                    delete_blocks_remaining=NewDeleteBlocksRemaining}).
+
+
+-spec finish(state()) -> state().
```
slfritchie: FWIW comment: my eyes are used to seeing `finish` as an FSM state, but this function isn't an FSM state and doesn't have a `gen_fsm` type spec.

kellymclaughlin:
On Jun 28, 2012, at 3:35 PM, Scott Lystig Fritchie wrote:

> Reminder for Kelly: search & destroy the manifest record `deleted` state.  Perhaps start with `egrep '\bdeleted\b' src/*` ?
> 
> ... although the `deleted` state does appear to be used inside of `riak_kv_delete_fsm.erl`.  But that state should not "leak" into any version of a manifest record that is stored in KV, correct?

Exactly right

src/riak_cs_delete_fsm.erl
kellymclaughlin: Perhaps the `finish` tasks would be better placed in `terminate`?

src/riak_cs_delete_fsm.erl
```diff
@@ -0,0 +1,213 @@
...
+finish(State=#state{manifest=?MANIFEST{state=deleted},
+                    bucket=Bucket,
+                    key=Key,
+                    uuid=UUID,
+                    riakc_pid=RiakcPid}) ->
+    {ok, ManiFsmPid} = riak_moss_manifest_fsm:start_link(Bucket, Key, RiakcPid),
```
slfritchie, line 192: Is the `ManiFsmPid` also being leaked by a `normal` exit?

kellymclaughlin, line 192: It is. I'll fix that as well.

src/riak_cs_delete_fsm.erl
slfritchie: That would be fine with me.

slfritchie:
Hrm, the style of several function arguments in `riak_cs_gc.erl` are hard to read: `a()` calls `b()` where `b` has function clauses to discriminate by (what appear to be?) `riakc_pb_socket` KV client operation return values, and sometimes it's `a()` that's doing the KV call (indirectly) or sometimes it's the caller of `a()` that did the KV call?  Why not have `a()` (and `a`'s caller) deal with the KV error cases?

src/riak_cs_gc.erl
```diff
@@ -0,0 +1,177 @@
...
+-compile(export_all).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
+%% export Public API
+-export([gc_interval/0,
+         gc_retry_interval/0,
+         gc_manifests/6,
+         leeway_seconds/0,
+         move_manifests_to_gc_bucket/2,
```
slfritchie: I don't see any external code calling this function, am I missing some test code somewhere?

src/riak_cs_gc.erl
```diff
@@ -0,0 +1,177 @@
...
+            Resolved = twop_set:resolve(SetsToResolve),
+            riakc_obj:update_value(PreviousObject, term_to_binary(Resolved))
+    end,
+
+    %% Create a set from the list of manifests
+    _ = lager:debug("Manifests scheduled for deletion: ~p", [ManifestSet]),
+    riak_moss_utils:put_with_no_meta(RiakcPid, ObjectToWrite).
+
+-spec build_manifest_set(twop_set:twop_set(), [lfs_manifest()]) -> twop_set:twop_set().
+build_manifest_set(Set, []) ->
```
slfritchie: Refactor using `lists:foldl/3`?

src/riak_cs_gc_d.erl
```diff
@@ -0,0 +1,550 @@
...
+%% @doc Status is returned as a 2-tuple of `{State, Details}'.  State
+%% should be `idle', `deleting', or `paused'.  When `idle' the
+%% details (a proplist) will include the schedule, as well as the
+%% times of the last calculation and the next planned calculation.
+%% When `calculating' or `paused' details also the scheduled time of
+%% the active calculation, the number of seconds the process has been
+%% calculating so far, and counts of how many users have been
+%% processed and how many are left.
+status() ->
+    gen_fsm:sync_send_event(?SERVER, status).
```
slfritchie: `infinity` timeout

src/riak_cs_gc.erl
```diff
@@ -0,0 +1,177 @@
...
+get_pending_delete_manifests({error, Reason}) ->
+    _ = lager:warning("Failed to get pending_delete manifests. Reason: ~p",
+                      [Reason]),
+    [].
+
+%% @doc Copy data for a list of manifests to the
+%% `riak-cs-gc' bucket to schedule them for deletion.
+-spec move_manifests_to_gc_bucket([lfs_manifest()], pid()) ->
+    ok | {error, term()}.
+move_manifests_to_gc_bucket(Manifests, RiakcPid) ->
```
slfritchie, line 133: From HipChat:

```
Scott F.
Noob conflict resolution question: if the system decides that it needs
to delete N objects all at the same wall clock second, it's possible
that O(N ^ 2) things will need to be get'ed and put'ed at the Riak
client level, using what's currently in
move_manifests_to_gc_bucket(). Can't we just do naive puts and let the
conflict resolution always be done by the GC daemon? (we'll still have
O(N ^ 2) data xfer, but it'd be "deeper" down at the riak_kv_vnode <->
backend layer.)

(assuming that it's the GC daemon that's the only one that's really
interested in the results of reading a key in the GC bucket)

(if a netsplit happened in the middle of the same second, then after
the split were healed, handoff would create list of yet-unresolved
siblings anyway)

Kelly M.
@scott that's an excellent observation
i think that will work fine
```

I didn't know if there was a hidden reason why to have the resolution
code here at object delete time ... but if there isn't a hidden reason, then
it'd be nice to remove a bit of code here.

slfritchie:
Is it intentional that the GC daemon cannot be paused if there is no active batch running?  It seems useful to me to be able to pause all GC activity to prevent any GC from this instant until I decide to resume.

kellymclaughlin:
The reasoning behind this style is to use more and smaller functions and avoid unnecessary nesting to try and improve the readability of the code. To me handling a return value in a function clause is just like handling it in a case statement, but you can give the function a meaningful name. 

For example the way that riak_cs_gc:gc_manifests is written is very easy to follow I think. What does it do? 1. mark the manifests 2. get pending_delete manifests 3. Move those manifests 4. Mark the manifests again. This is a fairly high-level function and I think it can be argued there is a point of diminishing returns when you get to the point of handling things like pb client responses. Also some of that interaction with the pb client could probably be better organized. Anyway that's some of the reasoning behind the stylistic choices here. I wonder if you find this approach disagreeable or just think that it needs some refinement for more clarity. Thanks.

slfritchie:
I'm a believer in using `timer:send_interval()` in many cases.  It isn't clear if it's possible for calls to `schedule_next()` to be missed in certain circumstances. If you're comfortable with the scheduling, I guess I am too, because I doubt that I'll be able to trace through all of the gen_fsm callbacks to verify that scheduling behaves itself.

src/riak_cs_gc_d.erl
```diff
@@ -0,0 +1,550 @@
...
+
+init(Args) ->
+    Interval = riak_cs_gc:gc_interval(),
+    SchedState = schedule_next(#state{interval=Interval}),
+    _ = check_bucket_props(Args),
+    {ok, idle, SchedState}.
+
+%% Asynchronous events
+
+%% @doc Transitions out of idle are all synchronous events
```
slfritchie, line 129: Heh, except for the `start_batch` timer-tiggered message.

src/riak_cs_gc_d.erl
```diff
@@ -0,0 +1,550 @@
...
+    %% just because the normal request pool is empty; pool
+    %% workers just happen to be literally the socket process,
+    %% so "starting" one here is the same as opening a
+    %% connection, and avoids duplicating the configuration
+    %% lookup code
+    {ok, Riak} = riak_moss_riakc_pool_worker:start_link([]),
+    NewState = start_batch(State#state{riak=Riak}),
+    {next_state, fetching_next_fileset, NewState};
+handle_info(start_batch, InBatch, State) ->
+    _ = lager:error("Unable to start garbage collection batch"
```
slfritchie: Change to info status: the operator may want to know that GC is falling behind, but I don't think it's an error level thing.  Some operators I've worked with in the past tend to take logged event severity literally ... and 'error' is assumed to be quite bad and should never happen under normal conditions.

src/riak_cs_gc_d.erl
```diff
@@ -0,0 +1,550 @@
...
+%% for delete.
+-spec fetch_eligible_manifest_keys(pid(), non_neg_integer()) -> [binary()].
+fetch_eligible_manifest_keys(RiakPid, IntervalStart) ->
+    EndTime = list_to_binary(integer_to_list(IntervalStart)),
+    eligible_manifest_keys(gc_index_query(RiakPid, EndTime)).
+
+eligible_manifest_keys({{ok, BKeys}, _}) ->
+    [Key || [_, Key] <- BKeys];
+eligible_manifest_keys({{error, Reason}, EndTime}) ->
+    _ = lager:warning("Error occurred trying to query from time 0 to ~p"
```
slfritchie, line 357: Hrm, lager message severity review ...  is there anything that the operator can do about this error?  I suppose we haven't really created rigid guidelines of what each severity level should mean.

If this is something that might happen occasionally as a matter of normal operation (including network partitions), then it should probably be info level?

src/riak_cs_gc.erl
```diff
@@ -0,0 +1,177 @@
...
+    Error.
+
+%% @doc Compile a list of `pending_delete' manifests.
+-spec get_pending_delete_manifests({ok, [{binary(), lfs_manifest()}]} |
+                                   {error, term()}) ->
+                                          [{binary(), lfs_manifest()}].
+get_pending_delete_manifests({ok, MarkedManifests}) ->
+    riak_moss_manifest:pending_delete_manifests(MarkedManifests);
+get_pending_delete_manifests({error, Reason}) ->
+    _ = lager:warning("Failed to get pending_delete manifests. Reason: ~p",
```
slfritchie, line 125: Hrm, the severity level here might also be `info`.  What if, for example, we were racing with another GC daemon on a different node?

Such a race begs the question of what if more than one GC daemon is operating in lock-step or nearly lock-step with another daemon on another node?  I've got that situation on my review-to-ponder list....

src/riak_cs_gc_d.erl
```diff
@@ -0,0 +1,550 @@
...
+            %% Get any values from the riak object and resolve them
+            %% into a single set.
+            BinValues = riakc_obj:get_values(RiakObj),
+            Values = [binary_to_term(BinValue) || BinValue <- BinValues],
+            ManifestSet = twop_set:resolve(Values),
+            {ok, ManifestSet, RiakObj};
+        {error, notfound}=Error ->
+            Error;
+        {error, Reason}=Error ->
+            _ = lager:warning("Error occurred trying to read the fileset"
```
slfritchie: info severity?

src/riak_cs_gc_d.erl
```diff
@@ -0,0 +1,550 @@
...
+-spec finish_file_delete(non_neg_integer(),
+                         twop_set:twop_set(),
+                         riakc_obj:riakc_obj(),
+                         pid()) -> ok.
+finish_file_delete(0, _, RiakObj, RiakPid) ->
+    %% Delete the key from the GC bucket
+    _ = riakc_pb_socket:delete_obj(RiakPid, RiakObj),
+    ok;
+finish_file_delete(_, FileSet, RiakObj, RiakPid) ->
+    _ = lager:info("Remaining file keys: ~p", [twop_set:to_list(FileSet)]),
```
slfritchie: debug severity?

src/riak_cs_gc_d.erl
```diff
@@ -0,0 +1,550 @@
...
+
+eligible_manifest_keys({{ok, BKeys}, _}) ->
+    [Key || [_, Key] <- BKeys];
+eligible_manifest_keys({{error, Reason}, EndTime}) ->
+    _ = lager:warning("Error occurred trying to query from time 0 to ~p"
+                      "in gc key index. Reason: ~p",
+                      [EndTime, Reason]),
+    [].
+
+gc_index_query(RiakPid, EndTime) ->
```
slfritchie, line 362: So, here's a crazy idea.  What if CS nodes could know about each other?  That'd be cool, because then we could ask other CS GC procs if they are also doing GC, and if so, what time they started GC, e.g., time T.  Then we could start our query at T or T+1 instead of `?EPOCH_START`.

src/riak_cs_gc_d.erl
```diff
@@ -0,0 +1,550 @@
...
+    {next_state, fetching_next_fileset, State#state{batch=RestKeys,
+                                                    batch_count=1+BatchCount}};
+initiating_file_delete(continue, #state{current_files=[NextManifest | _RestManifests],
+                                        riak=RiakPid}=State) ->
+    %% Use an instance of `riak_cs_delete_fsm' to handle the
+    %% deletion of the file blocks.
+    %% Don't worry about delete_fsm failures. Manifests are
+    %% rescheduled after a certain time.
+    Args = [RiakPid, NextManifest, []],
+    {ok, Pid} = riak_cs_delete_fsm_sup:start_delete_fsm(node(), Args),
```
slfritchie, line 196: I had a crazy time trying to figure out how this delete FSM's reply is sent to us.  I finally found the answer, here's a comment that I'm proposing adding here:

```
%% The delete FSM is hard-coded to send a sync event to our registered
%% name upon terminate(), so we do not have to pass our pid to it
%% in order to get a reply.
```

That FSM isn't trapping exits, hrm.  So if it dies, we'll wait forever waiting for the reply, which doesn't feel quite right to me.

src/riak_moss_block_server.erl
reiddraper, line 80: Notice the `start_link` call used in the top function clause also passes in a RiakcPid into start link, which is used. So there won't be an error trying to create a new connection.

src/riak_cs_delete_fsm.erl
reiddraper, line 164: Yeah, no doubt we've completely given up on the opacity. I'd be more inclined to argue for opacity if we didn't already have to distinguish between `dict` and `orddict` with the module name we use.

src/riak_cs_gc_d.erl
kellymclaughlin, line 362: I don't think it's a crazy idea. From the beginning of the CS endeavor I have thought about whether or not the CS nodes should be aware of each other. As in this case there are some real advantages, but also some disadvantages such as the increased complexity to support clustering of the nodes. I would like to think more about this and maybe have some discussions around it as time and brain power permits. I am not sure it's appropriate to implement it for this release, but I am guessing that's not what you're suggesting either?

src/riak_cs_gc_d.erl
reiddraper: fixed in 5180b21ac750cb06e6cd821b886bc7bd4cd2e83a

src/riak_cs_gc_d.erl
reiddraper: pushed fa0ec43ebab89d86abe3ecada9b80f6b4a5f03a5

src/riak_cs_gc.erl
reiddraper: pushed 1867b8b443288c6d39cda9a6df328e2a7f5eb56c

src/riak_cs_gc_d.erl
slfritchie, line 362: Well, I was thinking of this release, because it wouldn't be very hard.  Except for the "knowing other nodes & cookies" stuff.  :-)

I was thinking of something like an `rpc:multicall()` with a short (e.g. 3-10 second) timeout.  Take all successful answers and calculate T as the largest answer.  Overlapping queries & processing isn't supposed to hurt, so if you get zero answers, no big deal, use T = ?EPOCH_START.

src/riak_cs_gc.erl
reiddraper: fixed in bc996a4720694849dcbd193ea405df54240a4171

slfritchie:
After a mumble chat with Sean, Reid, and others ... the `twop_set.erl` implementation of a 2-phase CRDT is intentionally biased to favor delete operations.  Some comments (module level?) would be a good idea to make this (now implicit) bias explicit.

**added module docstring in 77bae8d221ec7a7912f8e47417ba95db05fb727a**

src/riak_cs_gc_d.erl
```diff
@@ -0,0 +1,550 @@
...
+waiting_file_delete(_, _From, State) ->
+    {reply, ok, waiting_file_delete, State}.
+
+paused({Pid, ok}, _From, State=#state{delete_fsm_pid=Pid,
+                                      pause_state=waiting_file_delete,
+                                      current_files=[CurrentManifest | RestManifests],
+                                      current_fileset=FileSet}) ->
+    UpdFileSet = twop_set:del_element(CurrentManifest, FileSet),
+    UpdState = State#state{delete_fsm_pid=undefined,
+                           pause_state=initiating_file_delete,
```
slfritchie: I'm puzzled by a couple of things here (for both ok and error processing of the delete FSM reply), when comparing the `paused` state to the `waiting_file_delete` state:
1. Why is `delete_fsm_pid` set to `undefined` in one and not the other?  Error?
2. Why is `pause_state` overwritten in one and not the other?  Error?

seancribbs:
My comments about the anomaly in the `twop_set`: https://gist.github.com/2175ecdb42e900cc6563

I think if you go forward with the version that violates the precondition, it would be best to name it something else, otherwise the anomaly will resurface unexpectedly when using the type in the normal way.

slfritchie:
Please review and, if acceptable, merge any/all on my behalf: https://github.com/basho/riak_moss/pull/225

src/riak_moss_lfs_utils.erl
```diff
@@ -198,9 +188,24 @@ remove_write_block(Manifest, Chunk) ->
...
+    Manifest?MANIFEST{write_blocks_remaining=Updated,
+                             state=ManiState,
+                             last_block_written_time=os:timestamp()}.
+
+%% @doc Remove a chunk from the `delete_blocks_remaining'
+%% field of `Manifest'
+remove_delete_block(Manifest, Chunk) ->
+    Remaining = Manifest?MANIFEST.delete_blocks_remaining,
+    Updated = ordsets:del_element(Chunk, Remaining),
+    ManiState = case Updated of
```
slfritchie, line 200: Ok, ok, I'll give up on ordsets opacity.

src/riak_moss_manifest.erl
```diff
@@ -39,47 +50,184 @@ active_manifest(Manifests) ->
...
-                    Manifests)
-    end.
+            FoldFun =
+                fun ({UUID, Elem}, Acc) ->
+                        case Elem of
+                            Active ->
+                                Acc;
+                            Elem=?MANIFEST{state=active} ->
+                                [UUID | Acc];
+                            Elem=?MANIFEST{state=writing,
```
slfritchie: This case clause could be eliminated if the `leeway_elapsed()` function could do the correct thing when changed to take both lbwt and wst and use the latter if the former is `undefined`?

src/riak_moss_manifest.erl
```diff
@@ -39,47 +50,184 @@ active_manifest(Manifests) ->
...
+    orddict:orddict(binary(), lfs_manifest()).
+prune(Manifests, Time) ->
+    [KV || {_K, V}=KV <- Manifests, not (needs_pruning(V, Time))].
+
+upgrade_wrapped_manifests(ListofOrdDicts) ->
+    DictMapFun = fun (_Key, Value) -> upgrade_manifest(Value) end,
+    MapFun = fun (Value) -> orddict:map(DictMapFun, Value) end,
+    lists:map(MapFun, ListofOrdDicts).
+
+%% TODO:
```
slfritchie: Doesn't this work?

```
-spec upgrade_manifest(#lfs_manifest_v1{} | #lfs_manifest_v2{} | #lfs_manifest_v3{}) -> ...
```

src/riak_moss_manifest_fsm.erl
```diff
@@ -75,25 +71,65 @@
...
+                    {error, notfound}
+            end;
+        {error, notfound}=NotFound ->
+            NotFound
+    end.
+
+get_specific_manifest(Pid, UUID) ->
+    case gen_fsm:sync_send_event(Pid, get_manifests, infinity) of
+        {ok, Manifests} ->
+            try orddict:fetch(UUID, Manifests) of
```
slfritchie: Hrm, it's mebbe a line of code shorter to use `orddict:find()` instead of `fetch()`.  But since we really aren't expecting to find the key 100% of the time, `find()` fits the API's intent better, I think.

src/riak_cs_gc_d.erl
reiddraper: 688d53a7b1cdf8fd46c41cd795276bc172189454

src/riak_cs_gc_d.erl
reiddraper: 211072a7c01181ac6d83570b8c8bece4a095b9b3

src/riak_cs_gc_d.erl
reiddraper, line 362: I think the simpler thing for this release would be to just have a singleton GC daemon. Using disterl this late in the game seems dangerous to me.

kellymclaughlin:
The code actually used timer:send_interval here until very recently, but I didn't see how using timer:send_interval was any different from timer:send_after as far as efficiency and since we will potentially have many concurrent processes using these timers I heeded the warning provided here: http://www.erlang.org/doc/efficiency_guide/commoncaveats.html#id2262042 and went with erlang:send_after. I'd be happy to change it back if my concern is misplaced here though.

kellymclaughlin:
On Jun 28, 2012, at 4:49 PM, Scott Lystig Fritchie wrote:

> Is it intentional that the GC daemon cannot be paused if there is no active batch running?  It seems useful to me to be able to pause all GC activity to prevent any GC from this instant until I decide to resume.

Yes I can see how this is confusing. Currently, you can pause a long running batch of files that are in the process of being gc'd, but pausing otherwise has no effect. I think a more useful thing is to extend that to be able to pause the daemon regardless of the state and that the daemon just picks up the batch at resume time if one was in progress. Also missing is a way to set the gc interval time via the console tool. I will make both changes.

src/riak_cs_gc_d.erl
kellymclaughlin, line 362: Another option is having all gc daemons read from and write to a `LAST_GC_START` key in some known bucket when the daemon is preparing to do a run. The reading daemon could just use the greatest time value of any siblings and write back a new time. This would not eliminate all duplicate work, but should serve to minimize it some, especially if paired with some randomization of the scheduling of the first interval for the case where many nodes are started within a close time proximity of one another.

src/riak_moss_manifest.erl
kellymclaughlin: Addressed by [e1c2a0f](https://github.com/basho/riak_moss/commit/e1c2a0f49b1c650fb54bc6e491890ffbe25bc17a)

src/riak_moss_manifest_fsm.erl
kellymclaughlin: Addressed by [614bb29](https://github.com/basho/riak_moss/commit/614bb2900aae4e92e76532eeeb2e24292ef5dc82)

kellymclaughlin:
Pushed commits to change `pause` so that the daemon can be paused in any state and so that the interval can be set via the `riak-cs-gc` command line utility. It works such that if the daemon is paused while in the `idle' state it records how much time is remaining on the interval timer and resets a timer with that remainder of time once the daemon is resumed. Discussion, arguments, or throwing of rotten produce welcomed.

slfritchie:
> The code actually used timer:send_interval here until very recently, [...]

I don't believe that we're using the `timer` module very much in Core or KV or CS.  I've been bitten by poor `timer` performance in Hibari, but that was caused by lots of `send_interval()` churn, process churn, and processes dying.

Since we're using it sparingly, the advantage of `timer:send_interval()` is pretty big: not needing to constantly calculate the sleep time & call whatever function to send the next reminder message.

slfritchie:
> Another option is having all gc daemons read from and write to a LAST_GC_START key [...]

I like that one, +1 if you've got the time & energy to add it.  Otherwise, it can wait for another day via an as-yet-uncreated kanban card?

slfritchie:
Kelly, no rotten produce for the pause changes, other than it'd conflict a small bit with commit b9dd932, which is part of https://github.com/basho/riak_moss/pull/225 ... have you or Reid had a chance to peek at it yet?

src/riak_cs_gc.erl
slfritchie, line 133: For the record:

> I would say yes, you can.  You'll pay a painful price of t2b'd dicts until we sort out object representation, but it should work.
> 
> It should go like this, once fowarded to a node in the preflist
> - put FSM calls vnode put with the coord flag on the local vnode
> - local vnode will get the object from the KV store, see that the incoming object is a sibling or ancestor and merge/increment the vclock, merging in the sibling.
> - put FSM sends requests to remote vnodes and stores.
> 
> At some point in the architectural future I'd like to investigate DVVs better and see if some kind of multi-instance-of-keys backend is possible so we could just write fragments and resolve at read time instead (you can still have the option to resolve at write).
> 
> Does that help?
> 
> Jon

Via HipChat, there was more discussion to clarify the CS use case.  Russell was a bit in doubt, perhaps, but Sean still thinks it's a bad idea unless otherwise backed by some real measurements.

So, here's what I'm going to do:
- The current code doesn't appear to have any obvious flaw, except perhaps suboptimal performance.
- I'll add a kanban card to measuring the difference and, if it's worthwhile, change the implementation.
  - https://github.com/basho/riak_moss/issues/226

kellymclaughlin:
On Jul 2, 2012, at 2:39 PM, Scott Lystig Fritchie wrote:

> Kelly, no rotten produce for the pause changes, other than it'd conflict a small bit with commit b9dd932, which is part of https://github.com/basho/riak_moss/pull/225 ... have you or Reid had a chance to peek at it yet?

Looking now.

slfritchie:
OK, I've another PR related to this one that I'd like some looks at:

https://github.com/basho/riak_moss/pull/228

Main highlights:
- Rename riak_moss_manifest.erl -> riak_cs_manifest_utils.erl
- Rename some functions in riak_cs_manifest_utils.erl to better reflect the data those functions actually operate on.
- Some small Dialyzer fixes

test/twop_set_eqc.erl
```diff
@@ -0,0 +1,276 @@
...
+                   Dels = twop_set:dels(Res),
+                   ?WHENFAIL(
+                      begin
+                          ?debugFmt("Adds: ~p~n", [sets:to_list(Adds)]),
+                          ?debugFmt("Deletes: ~p~n", [sets:to_list(Dels)]),
+                          ?debugFmt("Expected Adds: ~p~n", [sets:to_list(ExpectedAdds)]),
+                          ?debugFmt("Expected Deletes: ~p~n", [sets:to_list(ExpectedDels)])
+                      end,
+                      conjunction(
+                        [
```
slfritchie, line 107: I'm a bit puzzled by the 6 tests here.  Are they doing anything more than these two replacement tests?

```
                     {addX, equals(lists:sort(sets:to_list(Adds)),
                                   lists:sort(sets:to_list(ExpectedAdds)))},
                     {delX, equals(lists:sort(sets:to_list(Dels)),
                                   lists:sort(sets:to_list(ExpectedDels)))}
```

slfritchie:
Assuming that we can get everything merged together as discussed above and playing nicely, +1 to merge, many thanks to all.

test/twop_set_eqc.erl
kellymclaughlin, line 107: I remember there being a reason, but as I look at it now I can't imagine what it was.

kellymclaughlin, line 107: Addressed by [4a88e56](https://github.com/basho/riak_moss/commit/4a88e56f05f17cae51549775ecc686f687fa3444)

slfritchie:
Oops, I forgot to put this PR into this PR: https://github.com/basho/riak_moss/pull/231

Fixes for type errors in riak_moss_wm_key:forbidden/2 and riak_moss_wm_usage:forbidden/2.

include/riak_moss.hrl
```diff
@@ -216,11 +252,13 @@
 -define(MD_ACL, "X-Moss-Acl").
```
macshonle: Should this be <<"X-Moss-Acl">>?

SLF: This is a tricky thing.  The answer is probably no.  Given the string() vs. binary() flip-flops that is used for Riak's internal object representation of something like that metadata dictionary key and the external representation, I'd expect many things to break by blindly changing that constant to a binary() type.  But you can try it out and see what happens?

kellymclaughlin: Answer is yes based on the changes to `riak_pb` I made today. The handling of the `X-User-Meta` headers was a little messed up and that's why it is a string. It's already a binary on master now (that was in the commit I tacked onto your change) so I'll rebase this branch before merge and pull that change in.

PR #221, opened 2012-06-28
Config option for gc epoch time

kellymclaughlin:
This is to address cases where a very large number of objects have accumulated that need to be garbage collected such that it could potentially overwhelm the system.  The epoch start can be adjusted so that the backlog of objects can be handled in smaller batches and hopefully avoid any system degradation.


src/riak_cs_gc.erl
```diff
@@ -64,6 +65,18 @@ gc_retry_interval() ->
             RetryInterval
     end.
 
+%% @doc Return the start of GC epoch represented as a binary.
+%% This is the time that the GC daemon uses to  begin collecting keys
+%% from the `riak-cs-gc' bucket.
+-spec epoch_start() -> non_neg_integer().
```
reiddraper: should be `binary()`, not `non_neg_integer()`.

reiddraper:
+1 to merge

kellymclaughlin:
This branch is usurped by `moss221-gc-epoch-config-rebase`.

PR #224, opened 2012-06-28
Gh147 gc manifest resolution1

slfritchie:
Simply & shrink code: use integers to help compare states that we know "follow" each other in the simple chain of logic that is the manifest resolution process

Also, remove the no-longer-valid 'deleted' state.


reiddraper:
+1

PR #228, opened 2012-07-03
Gh147 gc manifest1

slfritchie:
Here's another PR that's based on the `gh147-gc` branch.  Main highlights:
- Rename riak_moss_manifest.erl -> riak_cs_manifest_utils.erl
- Rename some functions in riak_cs_manifest_utils.erl to better reflect the data those functions actually operate on.
- Some small Dialyzer fixes

### DIFF
--- a/src/riak_cs_gc.erl
+++ b/src/riak_cs_gc.erl
@@ -119,7 +119,7 @@ mark_manifests({error, _Reason}=Error, _, _, _) ->
                                    {error, term()}) ->
                                           [cs_uuid_and_manifest()].
 get_pending_delete_manifests({ok, MarkedManifests}) ->
-    riak_cs_manifest_utils:filter_pending_delete_manifests(MarkedManifests);
+    riak_cs_manifest_utils:filter_pending_delete_uuid_manifests(MarkedManifests);
 get_pending_delete_manifests({error, Reason}) ->
     _ = lager:warning("Failed to get pending_delete manifests. Reason: ~p",
                       [Reason]),
--- a/src/riak_cs_gc.erl
+++ b/src/riak_cs_gc.erl
@@ -15,7 +15,8 @@
 -endif.
 
 %% export Public API
--export([gc_interval/0,
+-export([decode_and_merge_siblings/2,
+         gc_interval/0,
          gc_retry_interval/0,
          gc_manifests/6,
          leeway_seconds/0,
@@ -142,18 +143,7 @@ move_manifests_to_gc_bucket(Manifests, RiakcPid) ->
             %% so resolve all the siblings and add the
             %% new set in as well. Write this
             %% value back to riak
-            %%
-            %% NOTE: If a prior GC daemon deleted this key as part of
-            %%       it work, then our get above can return siblings
-            %%       where one siblings is the tombstone.  In that case,
-            %%       we assume that that tombstone has a value <<>>,
-            %%       and we will skip looking in the value's metadata
-            %%       dict for the tombstone.
-            DecodedPrevious = [binary_to_term(V) ||
-                                  V <- riakc_obj:get_values(PreviousObject),
-                                  V /= <<>>],
-            SetsToResolve = [ManifestSet | DecodedPrevious],
-            Resolved = twop_set:resolve(SetsToResolve),
+            Resolved = decode_and_merge_siblings(PreviousObject, ManifestSet),
             riakc_obj:update_value(PreviousObject, term_to_binary(Resolved))
     end,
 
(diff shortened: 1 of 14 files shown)


src/riak_cs_gc.erl
```diff
@@ -169,6 +167,19 @@ generate_key() ->
...
         timestamp() + leeway_seconds())).
 
+%% @doc Given a list of riakc_obj-flavored object (with potentially
+%%      many siblings and perhaps a tombstone), decode and merge them.
+-spec decode_and_merge_siblings(riakc_obj:riakc_obj(), twop_set:twop_set()) ->
+      twop_set:twop_set().
+decode_and_merge_siblings(Obj, OtherManifestSets) ->
+    Some = [binary_to_term(V) || {MD, V} <- riakc_obj:get_contents(Obj),
+                                 not has_tombstone(MD),
+                                 V /= <<>>],
```
kellymclaughlin, line 173: Since we tracked down why the `X-Riak-Deleted` header was not showing up do we still need this? I guess it doesn't hurt as a safety valve or if someone has not updated the `riakc` dep.

src/riak_cs_gc.erl
```diff
@@ -169,6 +167,19 @@ generate_key() ->
...
+    twop_set:resolve([OtherManifestSets | Some]).
+
+has_tombstone(MD) ->
```
kellymclaughlin, line 176: Once [this](https://github.com/basho/riak-erlang-client/pull/63) pr is merged for the riak erlang client and we update `rebar.config` to point to the `1.2` branch or a yet-to-be-created tag for `riakc` I think we can remove this function and replace its use with a call to `riakc_obj:content_delete/1`.

kellymclaughlin:
+1 to merge

slfritchie:
Rebased:

```
% git push -f origin gh147-gc-manifest1
Counting objects: 73, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (50/50), done.
Writing objects: 100% (57/57), 11.40 KiB, done.
Total 57 (delta 43), reused 8 (delta 7)
To git@github.com:basho/riak_moss.git
 + 6fda6c6...306c119 gh147-gc-manifest1 -> gh147-gc-manifest1 (forced update)
```

slfritchie:
Hrm.  When I use the following `basho_bench` config to clobber the same file a few hundred times in 1 minute:

```
%% Sample config file for Riak CS basho_bench benchmarking

{mode, max}.
{duration, 1}.
{concurrent, 10}.

%% Replace these with your own paths
{source_dir, "/Users/fritchie/b/src/riak_moss/bench"}.
{code_paths, ["deps/ibrowse", "/Users/fritchie/b/src/riak_moss",
              "/Users/fritchie/b/src/riak_moss/deps/stanchion"]}.
{driver, basho_bench_driver_moss}.

%% Replace this with a user you have created.
%% Instructions to create a user are here:
%% https://github.com/basho/riak_moss/wiki/Creating-a-User
{moss_access_key, "NP3ZEHK_H9MBSHEP2XWS"}.
{moss_secret_key, "Nnc4WfGomD0XBAviP0fbr-yOiKRrhZH5KDX_Nw=="}.
{moss_bucket, "test"}.
%{moss_disconnect_frequency, 5}.
{moss_raw_ip, "s3.amazonaws.com"}. % DO NOT CHANGE
{moss_raw_port, 80}.               % DO NOT CHANGE
%% Replace these with your HTTP proxy's location (i.e. Riak CS)
{moss_http_proxy_host, "localhost"}.
{moss_http_proxy_port, 8080}.
{moss_request_timeout, 999999000}.

{key_generator, {uniform_int, 1}}.
{value_generator_source_size, 12100105}. %% Must be larger than val gen max size.
{value_generator, {fixed_bin, 1024}}.

{operations, [{insert, 1}]}.
```

... I see the following interesting things:
- `basho_bench` says that the object was overwritten 2974 times.
- After `basho_bench` has finished, the query `[{B, length(element(2,C:list_keys(B)))} || <<"0b:",_/binary>> = B <- (element(2, C:list_buckets()))].` says that there are are 2984 keys in the folder's bucket.  Hrm, perhaps basho_bench lied and didn't quite stop?  Or perhaps each worker proc failed to report one operation?  
- Then I kick off a manual GC: `riak_cs_gc_d:manual_batch([]).`
- When the GC is finished, I see a number that is quite a bit larger than 2974 or 2984:
  
  21:13:51.704 [info] Finished garbage collection: 32 seconds, 62 batch_count, 0 batch_skips, 5593 manif_count, 5593 block_count
- Then I use Riak & `list_keys` again to see how many keys remain in the block bucket: there are 222 undeleted blocks.

So, there appears to be a leak somewhere, when the same file is repeatedly clobbered.  I'll continue investigating.

slfritchie:
Weird.  With some brute-force tracing (io:format()) added, I see a small number of objects written with some UUID `u` where I see the initial manifest written and the final manifest written (it's a 1KB file, so no other manifest updates are needed), but that  `OverwrittenUUIDs` never has `u` as a member (in `riak_moss_manifest:get_and_update()`).

Giving up for the night, zzzzz.

reiddraper:
@slfritchie sanity check, do you have allow_mult true set in riak?

reiddraper:
I've had [weird troubles](https://gist.github.com/16ac8cd3132e62fdbf8c#comments) with basho_bench, but I did write a python script to overwrite the same key 1001 times. Here was the result of running a manual batch:

```
(riak-cs@127.0.0.1)1> riak_cs_gc_d:manual_batch([]).
ok
(riak-cs@127.0.0.1)2> 10:35:16.536 [info] Finished garbage collection: 3 seconds, 35 batch_count, 0 batch_skips, 1000 manif_count, 1000 block_count
```

slfritchie:
My comment of 13 hours ago is invalid: pilot error, due to {allow_mult,false}.  Many thanks for the sanity check, @reiddraper.

slfritchie:
I'm going to merge this branch to `gh147-gc`.  This branch has been rebased, so the GH's auto-close probably won't work.

PR #229, opened 2012-07-03
GC Daemon refactoring

kellymclaughlin:
Some functional refactoring and fixed a state problem found by the good old eqc test when a batch is canceled when the fsm is in the `paused` state. The `pause_state` record field needed to be updated in that case to be `idle`.


reiddraper:
+1 from me, I like this style

slfritchie:
Ditto

PR #231, opened 2012-07-07
Gh147 not-really-gc-related type fixes

slfritchie:
Here are fixes for a couple of type problems that I found while fiddling with Meck evil.  Both are the result of `finish_request()` callbacks being given the wrong context record type.  The result is a `function_clause` error, which then cascades to Poolboy with an error message like this to the console:

```
00:10:56.204 [error] Supervisor poolboy_sup had child riak_moss_riakc_pool_worker started with {riak_moss_riakc_pool_worker,start_link,undefined} at <0.223.0> exit with reason killed in context child_terminated
```


kellymclaughlin:
The inner context stuff in `riak_moss_wm_key` is a bit ugly. I'd like to do some refactoring to get rid of that, but now is probably not the time. Anyhow, +1 to merge on this.

PR #234, opened 2012-07-09
Update required VSN to version with hmac_init.

cmeiklejohn:
Update OTP VSN to R14B03 as a minimum, as it's the first release to support crypto:hmac_init.

@kellymclaughlin


kellymclaughlin:
+1 to merge

PR #237, opened 2012-07-11
Ms pb cs updates

macshonle:
This is for issue: https://github.com/basho/riak_moss/issues/232


kellymclaughlin:
I pushed a small commit to adjust for a fix that was made to `riak_pb` to properly handle the `X-Riak-Meta` headers. With that change and the update that is now merged to `master` of `riak_pb` all the tests pass for me so +1 to merge. Thanks.

PR #239, opened 2012-07-12
Bucket listing incompatability between 1.0 and master

kellymclaughlin:
Bucket names were stored as binaries in the bucket records in 1.0, but we have since changed them to be strings. Need to account for this when listing a user's buckets.


reiddraper:
+1 to merge

PR #240, opened 2012-07-13
Default dtrace_support to false.

kellymclaughlin:
Sadly most of our installs will probably be without dtrace support so default this to `false` to be safe.


slfritchie:
+1, thanks!

PR #249, opened 2012-07-23
Create 3-node devrel command

reiddraper:
Change devrel command to create 3
Riak CS nodes that are meant to talk
to (one-to-one relationship) with a
Riak devrel cluster. The HTTP ports for
Riak CS change to 807[1-3].

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,9 +48,11 @@ stage : rel
 	$(foreach dep,$(wildcard deps/*), rm -rf rel/riak-cs/lib/$(shell basename $(dep))-* && ln -sf $(abspath $(dep)) rel/riak-cs/lib;)
 	$(foreach app,$(wildcard apps/*), rm -rf rel/riak-cs/lib/$(shell basename $(app))-* && ln -sf $(abspath $(app)) rel/riak-cs/lib;)
 
-devrel:
+devrel: dev1 dev2 dev3
+
+dev1 dev2 dev3:
 	mkdir -p dev
-	(cd rel && ../rebar generate skip_deps=true target_dir=../dev/$(PKG_NAME) overlay_vars=vars/dev_vars.config)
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -48,9 +48,11 @@ stage : rel
 	$(foreach dep,$(wildcard deps/*), rm -rf rel/riak-cs/lib/$(shell basename $(dep))-* && ln -sf $(abspath $(dep)) rel/riak-cs/lib;)
 	$(foreach app,$(wildcard apps/*), rm -rf rel/riak-cs/lib/$(shell basename $(app))-* && ln -sf $(abspath $(app)) rel/riak-cs/lib;)
 
-devrel:
+devrel: dev1 dev2 dev3
+
+dev1 dev2 dev3:
 	mkdir -p dev
-	(cd rel && ../rebar generate skip_deps=true target_dir=../dev/$(PKG_NAME) overlay_vars=vars/dev_vars.config)
+	(cd rel && ../rebar generate skip_deps=true target_dir=../dev/$@ overlay_vars=vars/$@_vars.config)
```
kellymclaughlin, line 55: Get rid of the `skip_deps=true` here. It causes the beam files for the dependencies not to be copied into the release and the release fails to start. For some odd reason `skip_deps` is required for the `rel` target due to some error with `stanchion` and the resulting release works. Probably something to do with the directory changing for the `devrel` target.

kellymclaughlin:
Why not just have 4 nodes created instead of 3 to match the 4 created by riak?

kellymclaughlin:
It would be nice if we could have a `stagedevrel` target also. I tend to use `stage` and `stagedevrel` much more frequently than `rel` or `devrel`.

kellymclaughlin:
Looks great! +1 to merge.

PR #250, opened 2012-07-23
Use vclock when issuing block deletes

reiddraper:
Do a head (get w/ [head] option) request to get the
vector clock for a block before issuing the delete.
Use the vclock when issuing the subsequent block
delete.


kellymclaughlin:
+1 to merge

PR #252, opened 2012-07-24
Handle tombstones in manifest fsm

kellymclaughlin:
```
(riak-cs@127.0.0.1)36> 17:41:04.327 [error] gen_fsm <0.17924.129> in state waiting_command terminated with reason: bad argument in call to erlang:binary_to_term(<<>>) in lists:map/2 line 1173
```

### DIFF
--- a/src/riak_cs_gc.erl
+++ b/src/riak_cs_gc.erl
@@ -181,13 +181,9 @@ generate_key() ->
 -spec decode_and_merge_siblings(riakc_obj:riakc_obj(), twop_set:twop_set()) ->
       twop_set:twop_set().
 decode_and_merge_siblings(Obj, OtherManifestSets) ->
-    Some = [binary_to_term(V) || {MD, V} <- riakc_obj:get_contents(Obj),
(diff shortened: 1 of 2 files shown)


slfritchie:
+1 to merge, thanks!

shino:
@kellymclaughlin May I ask a question for this old issue?

I don't understand why empty binary check at https://github.com/basho/riak_cs/pull/252/files#L1R446 is added.
Does not `dict:is_key(<<"X-Riak-Deleted">>, MD)` suffice in order to confirm an object is in tombstone state?
I want to understand the change to fix the issue https://github.com/basho/riak_cs/issues/524.

kellymclaughlin:
I believe the reason is that the `X-Riak-Deleted` has not always been reliably included so the extra check is to handle those cases. I think that problem has since been corrected so we can probably look at only using the check for `X-Riak-Deleted` in the future.

shino:
@kellymclaughlin Thanks for reply. I understand it as a historical reason.

PR #251, opened 2012-07-25
Monitor delete fsm from gc daemon

reiddraper:
We need to monitor the delete FSMs the GC daemon starts, because if they crash we get stuck forever in the `waiting_file_delete` state, waiting for a response from the FSM. I believe we should treat this case exactly the same as if the delete_fsm replies back to the gc daemon with an error.

In the particular case I found, the delete fsm is dying because (actually in the manifest fsm):

```
(riak-cs@127.0.0.1)36> 17:41:04.327 [error] gen_fsm <0.17924.129> in state waiting_command terminated with reason: bad argument in call to erlang:binary_to_term(<<>>) in lists:map/2 line 1173
```

### DIFF
--- a/src/riak_cs_gc_d.erl
+++ b/src/riak_cs_gc_d.erl
@@ -194,6 +194,11 @@ initiating_file_delete(continue, #state{current_files=[Manifest | _RestManifests
     %% name upon terminate(), so we do not have to pass our pid to it
     %% in order to get a reply.
     {ok, Pid} = riak_cs_delete_fsm_sup:start_delete_fsm(node(), Args),
+
+    %% Link to the delete fsm, so that if it dies,
+    %% we go down too. If the link fails, we should
+    %% also receive an 'EXIT' message.
+    catch link(Pid),
     {next_state, waiting_file_delete, State#state{delete_fsm_pid = Pid}};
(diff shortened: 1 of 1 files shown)


src/riak_cs_gc_d.erl
```diff
@@ -194,6 +194,11 @@ initiating_file_delete(continue, #state{current_files=[Manifest | _RestManifests
     %% name upon terminate(), so we do not have to pass our pid to it
     %% in order to get a reply.
     {ok, Pid} = riak_cs_delete_fsm_sup:start_delete_fsm(node(), Args),
+
+    %% Link to the delete fsm, so that if it dies,
+    %% we go down too. If the link fails, we should
+    %% also receive an 'EXIT' message.
+    catch link(Pid),
```
kellymclaughlin: I don't think you actually need `catch` here. If the delete fsm crashes between the time when it is started and the daemon calls link, it may actually provide more useful crash information if the failure is not caught. It also causes the crash to happen as early as possible. I don't think having the `catch` really matters for any of the other failure scenarios so I think removing it is ok.

kellymclaughlin:
+1 to merge

PR #256, opened 2012-08-01
make target for client tests

reiddraper:
run all of (or maybe selectively) the client integration tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ ERLANG_BIN       = $(shell dirname $(shell which erl))
 REBAR           ?= $(BASE_DIR)/rebar
 OVERLAY_VARS    ?=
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -9,7 +9,7 @@ ERLANG_BIN       = $(shell dirname $(shell which erl))
 REBAR           ?= $(BASE_DIR)/rebar
 OVERLAY_VARS    ?=
 
-.PHONY: rel stagedevrel deps test
+.PHONY: rel stagedevrel deps test test-python test-clojure test-client
```
kellymclaughlin: You don't need to add these as phony targets since there is no directory named `test-*`. That's the only reason `test` is listed here.

kellymclaughlin:
+1 to merge

PR #258, opened 2012-08-03
Secondary block delete attempt when primary constraints fail

kellymclaughlin:
This goal of this change is to minimize the effect of the data resurrection condition when deleting object from Riak. It also has the effect of serving as a way to force read-repair on blocks that were deleted while one or more primaries were unavailable.

### DIFF
--- a/src/riak_cs_gc_d.erl
+++ b/src/riak_cs_gc_d.erl
@@ -524,7 +524,7 @@ handle_common_sync_reply({MsgBase, _}, Common, State) when is_atom(MsgBase) ->
 %% Refactor TODO:
 %%   1. delete_fsm_pid=undefined is desirable in both ok & error cases?
 %%   2. It's correct to *not* change pause_state?
-handle_delete_fsm_reply({ok, NumDeleted},
+handle_delete_fsm_reply({ok, {TotalBlocks, TotalBlocks}},
                         #state{current_files=[CurrentManifest | RestManifests],
                                current_fileset=FileSet,
                                block_count=BlockCount} = State) ->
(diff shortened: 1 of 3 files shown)


src/riak_cs_gc_d.erl
```diff
@@ -533,6 +533,13 @@ handle_delete_fsm_reply({ok, NumDeleted},
     State#state{delete_fsm_pid=undefined,
                 current_fileset=UpdFileSet,
                 current_files=RestManifests,
+                block_count=BlockCount+TotalBlocks};
+handle_delete_fsm_reply({ok, {NumDeleted, _TotalBlocks}},
+                        #state{current_files=[_CurrentManifest | RestManifests],
+                               block_count=BlockCount} = State) ->
+    gen_fsm:send_event(self(), continue),
```
reiddraper, line 540: Not something I'm really thinking about for this review, but I'm wondering if sending events to yourself is a bit of a code smell. Would it be better to just call the helper handle_$WHATEVER fun that would correspond to the helper that was called in the gen_fsm callback for receiving `continue`?

kellymclaughlin, line 540: I think explicitly calling a callback is ugly as well, but I don't know really which is worse. I have seen both methods used. Ideally `gen_fsm` would have a builtin way to do an _active_ state transition as a different result type, but alas it does not. So I guess something we may want to evaluate. I guess perhaps directly calling the callback has the advantage of avoiding any other messages being processed that may have snuck on the the message queue ahead of the `continue`.

reiddraper:
+1 to merge, nice work

src/riak_cs_gc_d.erl
reiddraper, line 540: Not sure if I was clear about this, but I thinking calling a helper function, not one of the gen_fsm callbacks. If we're doing the "small functions thing anyway", the gen_fsm callback likely just proxies to another helper fun immediately anyway.

It might end up looking something like this:

``` erlang
handle_delete_fsm_reply({error, _}, #state{current_files=[_ | RestManifests]} = State) ->
    ContState = continue(State),
    ContState#state{delete_fsm_pid=undefined,
                current_files=RestManifests}.
```

PR #267, opened 2012-08-10
Move and rename boto test module.

kellymclaughlin:
- Rename `s3api_verification_test.py` to `boto_test.py`.
- Move `boto_test.py` to `client_tests/python`.
- Update Makefile targets.

I made a separate `test-python` Makefile target mostly as a placeholder for the time when we may have s3cmd tests or other python library tests.


reiddraper:
+1

PR #268, opened 2012-08-10
Upgrade clojure tests to leiningen2

reiddraper:
Instructions for upgrading are here:
https://github.com/technomancy/leiningen/wiki/Upgrading

make everything should remain the same from the perspective
of running the make target, 'make test-clojure'


kellymclaughlin:
wfm. +1 to merge.

PR #272, opened 2012-08-17
Don't expose tracebacks to users on 500 error

reiddraper:
The error is still logged to error.log, so it is not hidden.


kellymclaughlin:
Works like a charm. +1 to merge.

PR #266, opened 2012-08-23
Make a test using the erlcloud library

kellymclaughlin:
Add an client test for the `erlcloud` library. Also add a rebar plugin and do some file reorganization to separate out the running of client tests, integration test, and normal eunit tests.


reiddraper:
Woot, nice work! At some point we should include a wiki page or some other note about running dnsmasq while running test-erlang.

+1 to merge

PR #279, opened 2012-08-29
Don't require folsom to be restarted when the riak_cs app is restarted

kellymclaughlin:
Detect if `folsom` metrics already exist and delete and recreate is so. This allow the `riak_cs` application to be restarted without requiring `folsom` to also be restarted which is beneficial for testing.

I'm not certain the metrics need to be deleted and recreated, but I don't see that it hurts to do it for good measure.


reiddraper:
+1

PR #283, opened 2012-09-05
Refactor wm_service to use riak_moss_wm_utils:find_and_auth_user

kellymclaughlin:
Change `riak_moss_wm_service` module to use `riak_moss_wm_utils:find_and_auth_user`.  Also add a boolean parameter to `find_and_auth_user` to indicate if anonymous requests should be accepted or rejected.


reiddraper:
+1

PR #274, opened 2012-09-06
Issue with Updates on User Objects

kellymclaughlin:
From a client perspective: Updates on user objects (like changing status) don't work consistently.
From an internal perspective: It looks like siblings get created which then don't get merged/resolved correctly.

Vector clocks were not being used for user account updates and this resulted in sibling creation and unexpected behavior from the user perspective. This change reworks how user record updates are handled so that now instead of storing only the vector clock in the webmachine resource context and having to reconstruct an entire new object for an update, we store the entire object in the context and make the appropriate updates to that object prior to saving.

There is still a chance that even with these changes that a user object could have siblings, but this would require multiple admin users modifying the same account at the same time. Currently, the code to retrieve and update the user records can handle the siblings, but there is no meaningful semantic I can think of with the current user record to intelligently pick which sibling should _win_. In the future we may want to add a modified timestamp to the user record as a means to make this decision a little more intelligently. For now I think the likelihood of this happening is minimal enough to wait on that.

### DIFF
--- a/include/riak_moss.hrl
+++ b/include/riak_moss.hrl
@@ -43,7 +43,7 @@
 -record(context, {start_time :: erlang:timestamp(),
                   auth_bypass :: atom(),
                   user :: moss_user(),
-                  user_vclock :: term(),
+                  user_object :: riakc_obj:riakc_obj(),
                   bucket :: binary(),
                   requested_perm :: acl_perm(),
                   riakc_pid :: pid()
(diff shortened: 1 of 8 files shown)


reiddraper:
Should this be going through stanchion? Or are we getting our serialization by assuming only one admin is changing a particular user record at a time?

reiddraper:
Hipchat convo re: ^^, kelly and I decided to not use stanchion for now, but to consider it for the future, given the timeline for this release

reiddraper:
+1 to merge

PR #284, opened 2012-09-07
Added documentation for the 'anonymous_user_creation' in the app.config ...

novabyte:
A user was following "[Configuring Riak CS](http://wiki.basho.com/Configuring-Riak-CS.html)" and couldn't find the `{anonymous_user_creation, true}` config property.

I've added this to the app.config as `{anonymous_user_creation, false}` with some code documentation.


reiddraper:
+1 to merge

PR #287, opened 2012-09-12
Add basic support for riak-test

bookshelfdave:
Please see riak_test/bin/README for basic info on how to run this.
### Notes
-  riak_cs devrel's need to be fixed if you want to remove the rtdevrel make targets and `rel/vars/rtdev[1-6]_vars.config` files. At that point, `/tmp/rt/current/cs/rtdev*` targets need to be changed in the code. The current devrel would not allow me to copy ./dev/\* to another directory and run, as paths were hardcoded as `../`.
- node_package needs to be pinned at 1.0.0. As of right now, node_package master looks for `/lib/env.sh`.


reiddraper:
nice work, +1 to merge

PR #288, opened 2012-09-14
test may timeout in weak machine: test_n_chunks_builer/1 takes O(n) time...

kuenishi:
... while its timeout was fixed - 30 sec


reiddraper:
+1

PR #291, opened 2012-09-17
Fix some of the issues found by dialyzer

reiddraper:
I've been running dialyzer like so:

`dialyzer -Wno_return -Wunmatched_returns --plt ~/.riak-cs_dialyzer_plt ebin | fgrep -v -f ./dialyzer.ignore-warnings`

notice I'm ignoring deps for now. Each commit should be an explanation of the error dialyzer found.


slfritchie:
+1 to merge, thanks!

PR #292, opened 2012-09-18
Fix sbin scripts and reltool misconfiguration

jaredmorrow:
Additional shell scripts with variable substitutions existed in the `priv/sbin` directory that I missed when updating node_package directories and scripts.   I updated this to fall in line with the runner script `sbin/riak-cs`.

Also, in an a horrible attempt to clean up after riak-cs 1.1, I removed the empty directory `data/riak-cs` from `reltool.config` and that broke packaging.  Going forward I would like to have that configurable in node_package, but for now putting that directory back is the better fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,7 @@ stagedevrel: dev1 dev2 dev3 dev4
 rtdevrel: rtdev1 rtdev2 rtdev3 rtdev4 rtdev5 rtdev6
 
 rtdev1 rtdev2 rtdev3 rtdev4 rtdev5 rtdev6:
+	$(if $(RT_TARGET_CS),,$(error "RT_TARGET_CS var not set, see riak_test/ directory for details"))
 	mkdir -p $(RT_TARGET_CS)/$@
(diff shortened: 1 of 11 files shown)


bookshelfdave:
tested on Ubuntu 12.04
+1

bookshelfdave:
Although for the riak_test + CS, I get this error:

[0]R15B01:[jem-fix-sbin-scripts]brawl:~/test-releases/riak_cs$ /tmp/rt/current/cs/rtdev3/sbin/riak-cs console
/tmp/rt/current/cs/rtdev3/sbin/riak-cs: 8: .: Can't open {{target_dir}}/sbin/../lib/env.sh

jaredmorrow:
The rtdevrel*_vars.config files didn't get updated along with the other
vars.config files.  They were on a parallel branch at the time of my merge.
 I'll update them on this branch to bring it all in at once.

bookshelfdave:
looks good, merge away

Thanks Jared.

PR #293, opened 2012-09-21
refactor gc moving

reiddraper:


reiddraper:
So, in hindsight I probably shouldn't have based this work off of the [dialyzer fixes branch](https://github.com/basho/riak_cs/pull/291), but alas.

reiddraper:
The main goal of this PR is to implement that change described in fce4c7a19135333547842ca497b924d9126be985

include/riak_cs.hrl
```diff
@@ -31,7 +31,7 @@
           acl :: acl()}).
 
 -record(moss_bucket_v1, {
-          name :: string(),
+          name :: string() | binary(),
```
slfritchie, line 34: Hrm, why the type change here?  The name in the __user_v_ records is still a string().

slfritchie:
Everything else seems sane.  No new tests, but since this PR is supposed to make writing other tests easier, I can't complain.  +1 to merge, thanks!

include/riak_cs.hrl
reiddraper, line 34: I believe this was because previously there were some records stored with binaries. @kellymclaughlin you recall that being true?

PR #296, opened 2012-09-28
403 instead of 404 returned when trying to list nonexistent bucket

kellymclaughlin:
Adds a check when a valid user account attempts to list a bucket so that a `404` is returned if the bucket does not exist in the system. If the bucket does exist the system continues to return a `403`. Also adds a `check_bucket_exists` helper function to the `riak_cs_utils` module.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -22,18 +22,18 @@
 ]}.
 
 {deps, [
-        {node_package, ".*", {git, "git://github.com/basho/node_package", {branch, "master"}}},
+        {node_package, "1.1.*", {git, "git://github.com/basho/node_package", {tag, "1.1.0"}}},
         {webmachine, ".*", {git, "git://github.com/basho/webmachine", {branch, "wrd-content-length-with-streaming"}}},
-        {riakc, ".*", {git, "git://github.com/basho/riak-erlang-client", {branch, "master"}}},
-        {lager, ".*", {git, "git://github.com/basho/lager", {branch, "master"}}},
-        {eper, ".*", {git, "git://github.com/basho/eper.git", "master"}},
-        {sext, ".*", {git, "git://github.com/esl/sext", "master"}},
+        {riakc, "1.3.1", {git, "git://github.com/basho/riak-erlang-client", {tag, "1.3.1"}}},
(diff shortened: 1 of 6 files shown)


slfritchie:
I'm a bit puzzled here.  If I have an environment where:
- Bucket 'test' is owned by the current user & is readable.
- Bucket 'test2' is owned by another user and has been set unreadable via `s3cmd setacl --acl-private s3://test2` by that other user
- Bucket 'testX' does not exist.

If I use the 'master' branch, I see:

```
% s3cmd --debug ls s3://test |& grep S3Error
% s3cmd --debug ls s3://test2 | & grep S3Error
DEBUG: S3Error: 403 (Forbidden)
% s3cmd --debug ls s3://testX | & grep S3Error
DEBUG: S3Error: 404 (Object Not Found)
```

If I use the 'cs296-nonexistent-bucket-404' branch, I see the exact same thing.

kellymclaughlin:
I left out a key piece of information in the original problem description. This problem manifests when a bucket is created by a user and then subsequently deleted. Buckets that have never existed do not exhibit the problem.

slfritchie:
+1 to the fix, -1 to the regression test (or, rather, lack of one).

Here's a rough sketch of what I'd done to create.

Setup:

```
s3cmd -c ~/.s3cfg.sbb-number1 mb s3://test
s3cmd -c ~/.s3cfg.sbb-number2 mb s3://test2
s3cmd -c ~/.s3cfg.sbb-number2 setacl --acl-private s3://test2
s3cmd -c ~/.s3cfg.sbb-number2 rb s3://test2
```

Master's behavior (bad):

```
% s3cmd -c ~/.s3cfg.sbb-number1 --debug ls s3://test |& grep S3Error
% s3cmd -c ~/.s3cfg.sbb-number1 --debug ls s3://test2 | & grep S3Error
DEBUG: S3Error: 403 (Forbidden)
% s3cmd -c ~/.s3cfg.sbb-number1 --debug ls s3://testX | & grep S3Error
DEBUG: S3Error: 404 (Object Not Found)
```

Whereas cs296-nonexistent-bucket-404's behavior yields 404 for both the 2nd & 3rd commands.

reiddraper:
+1 to merge

PR #298, opened 2012-10-03
Add prereduce to storage calculation

ian-plosker:
In my trivial test, the storage calculation for a bucket with 100K objects was reduced from 7s to 3s on a 6 node 256 partition cluster on physical hardware.


beerriot:
This _might_ make the calculation for small buckets slower by a couple of milliseconds, due to the overhead in the extra pipe stage for the prereduce. But, for large buckets, it should be all improvement.

Even a 10k-object bucket on a single Riak node on my laptop dropped from 3s to 2s with this patch.

+1 to merge whenever it fits in your release schedule, CS team. This should have no upgrade implications, as no APIs change. It could even be rolled out as a hot patch to the module on the CS node, if needed.

kellymclaughlin:
Merging to 1.2 via the `igp-prereduce-1.2` branch

PR #299, opened 2012-10-16
Replace basho_bench_log:log() -> lager:log() calls

slfritchie:
Take care of some bitrot that Seth Thomas noticed a couple of weeks ago: basho_bench's internal logging convention changed, so the drivers outside of the basho_bench tree need to change as well.


reiddraper:
+1 to merge

PR #303, opened 2012-10-19
Don't do bucket list for HEAD or ?versioning or ?location request.

reiddraper:
Some clients use HEAD or the `versioning` or `location`
subresource to check to see if a bucket exists. Previously,
they would cause an expensive bucket list behind the scenes.

There is still a known issue where anonymous users will
receive 403 on a request when they should get 404.

The response for versioning returns that versioning has not been turned
on (we don't support it yet). And the response for location
returns the 'classic' location.


kellymclaughlin:
+1 to merge

PR #306, opened 2012-10-22
Reduce phase for bucket listing

kellymclaughlin:
Pipe all bucket listing results through a reduce phase to take advantage of the builtin backpressure. This is just a temporary change to prevent large buckets from causing riak nodes to crash until proper sink backpressure is added to riak_pipe.

### DIFF
--- a/src/riak_cs_utils.erl
+++ b/src/riak_cs_utils.erl
@@ -26,6 +26,7 @@
          has_tombstone/1,
          is_admin/1,
          map_keys_and_manifests/3,
+         reduce_keys_and_manifests/2,
          get_object/3,
          get_manifests/3,
          get_user/2,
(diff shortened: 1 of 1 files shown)


dreverri:
Would there be any benefit to using pre-reduce as well?
http://docs.basho.com/riak/latest/references/appendices/MapReduce-Implementation/#Configuration%20Tuning%20for%20Reduce%20Phases

kellymclaughlin:
@dreverri Not in the case of bucket listings I don't think because we are just using the reduce for the queuing and backpressure, but we have merged that for storage calculations: https://github.com/basho/riak_cs/commit/6eb55fcd081abb926d541b4ea76062362cffb1c9

slfritchie:
+1, thanks!

PR #307, opened 2012-10-24
Gh236 stop waiting value

slfritchie:
Sorry, no test, but according to OTP recommended usage of gen_fsm, all states should have StateName/2 and StateName/3 functions ... so I've added a couple of other missing callback functions.


reiddraper:
+1 to merge

PR #313, opened 2012-10-26
Delete manifests with a vector clock

reiddraper:
Previously we deleted just by BKey. We should've been
using the vector clock from the object.


slfritchie:
+1 to merge, thanks!

PR #322, opened 2012-11-08
Adjust the timeouts on the riak_cs_get_fsm_test.

kellymclaughlin:
Increase the timeouts for the get_fsm test to be more forgiving on slower or busy systems.


reiddraper:
+1

PR #323, opened 2012-11-09
Never send stop message to undefined pid in terminate

reiddraper:
Add helper functions to repsective modules to allow
termination of servers, when the value may not have
been initialized yet.


kellymclaughlin:
+1 to merge

PR #326, opened 2012-11-13
Clean up and fix edge case in GC moving

reiddraper:
Previously the manifests marked as scheduled
delete and moved to GC bucket would be any
in the `pending_delete` state. This was for
two reasons. One, `handle_mark_as_pending_delete`
would grab the `pending_delete` manifests based
on observing all the manifests, not just the
specific UUIDs passed in. Two, `pending_delete_manifest/2`
assumed that manifests were updated as their blocks were deleted.
This was based on an outdated view of GC. Now, the manifests
are copied into the GC bucket the GC daemon views and manipulates
them from this bucket.

This change makes it so that the manifests that are moved
to the GC bucket are:
1. the UUIDs passed into the GC function to begin with
2. any `pending_delete` manifests that appear to have been abandoned.
   This means manifests that were put into the `pending_delete` state
   but were never (or likely never) successfully moved to the GC bucket
   and marked as `scheduled_delete`.

### DIFF
--- a/src/riak_cs_manifest_utils.erl
+++ b/src/riak_cs_manifest_utils.erl
@@ -83,30 +83,38 @@ overwritten_UUIDs_no_active_fold({UUID, _}, Acc) ->
     fun(({binary(), lfs_manifest()}, [binary()]) -> [binary()]).
 overwritten_UUIDs_active_fold_helper(Active) ->
     fun({UUID, Manifest}, Acc) ->
-            case {UUID, Manifest} of
-                {_UUID, Active} ->
-                    Acc;
-                {_UUID, ?MANIFEST{state=active}} ->
-                    [UUID | Acc];
-                {_UUID, ?MANIFEST{state=writing,
-                                  last_block_written_time=undefined,
-                                  write_start_time=WST}} ->
-                    acc_leeway_helper(UUID, Acc, WST);
-                {_UUID, ?MANIFEST{state=writing,
-                                  last_block_written_time=LBWT}} ->
-                    acc_leeway_helper(UUID, Acc, LBWT);
-                _Else ->
-                    Acc
-            end
+            update_acc(UUID, Manifest, Acc, Active =:= Manifest)
     end.
 
--spec acc_leeway_helper(binary(), list(), undefined | erlang:timestamp()) ->
-    list().
+-spec update_acc(binary(), lfs_manifest(), [binary()], boolean()) ->
+    [binary()].
+update_acc(_UUID, _Manifest, Acc, true) ->
+    Acc;
+update_acc(UUID, ?MANIFEST{state=active}, Acc, false) ->
+    [UUID | Acc];
+update_acc(UUID, Manifest=?MANIFEST{state=writing}, Acc, _) ->
+    LBWT = Manifest?MANIFEST.last_block_written_time,
+    WST = Manifest?MANIFEST.write_start_time,
+    acc_leeway_helper(UUID, Acc, LBWT, WST);
+update_acc(_, _, Acc, _) ->
+   Acc.
+
+-spec acc_leeway_helper(binary(), [binary()],
+                        undefined | erlang:timestamp(),
+                        undefined | erlang:timestamp()) ->
+    [binary()].
+acc_leeway_helper(UUID, Acc, undefined, WST) ->
+    acc_leeway_helper(UUID, Acc, WST);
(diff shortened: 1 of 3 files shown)


reiddraper:
Going to push another change to this branch, please hold up on review for the time being.

reiddraper:
Ok now ready

src/riak_cs_manifest_utils.erl
```diff
@@ -62,28 +62,55 @@ active_and_writing_manifests(Dict) ->
...
+overwritten_UUIDs_no_active_fold({_, ?MANIFEST{state=State}}, Acc)
+        when State =:= writing ->
+    Acc;
+overwritten_UUIDs_no_active_fold({UUID, _}, Acc) ->
+    [UUID | Acc].
+
+-spec overwritten_UUIDs_active_fold_helper(lfs_manifest()) ->
+    fun(({binary(), lfs_manifest()}, [binary()]) -> [binary()]).
+overwritten_UUIDs_active_fold_helper(Active) ->
+    fun ({UUID, Manifest}, Acc) ->
```
kellymclaughlin: This is the line i meant to comment on. Extra space after `fun`.

src/riak_cs_manifest_utils.erl
```diff
@@ -62,28 +62,55 @@ active_and_writing_manifests(Dict) ->
...
+            case {UUID, Manifest} of
```
kellymclaughlin: Is it worth refactoring a little more to break out this `case` statement into some smaller functions? It's a little ugly, but not completely so. On the other hand I do like getting rid of ugly `case` statements. What do you think?

reiddraper: So, the reason I did this is because I needed to close over  the `Active` variable and pattern match on it. And (for better or worse) you can't return a lambda that uses the closed over variables in its head pattern matches. [This](https://gist.github.com/fa63bacf9a1a5c2475ce) was the first way I tried to do it, but it gets a compile error (actually a warning about shadowing `Active`). Is that what you had in mind? If there's another way to avoid the case statements I'd be all for it.

src/riak_cs_manifest_utils.erl
```diff
@@ -62,28 +62,55 @@ active_and_writing_manifests(Dict) ->
...
+                                       [binary()]) -> [binary()].
+overwritten_UUIDs_no_active_fold({_, ?MANIFEST{state=State}}, Acc)
+        when State =:= writing ->
+    Acc;
+overwritten_UUIDs_no_active_fold({UUID, _}, Acc) ->
+    [UUID | Acc].
+
+-spec overwritten_UUIDs_active_fold_helper(lfs_manifest()) ->
+    fun(({binary(), lfs_manifest()}, [binary()]) -> [binary()]).
+overwritten_UUIDs_active_fold_helper(Active) ->
```
kellymclaughlin, line 84: So what do you think of replacing `overwritten_UUIDs_active_fold_helper` and adding additional functions similar to this:

```
-spec overwritten_UUIDs_active_fold_helper(lfs_manifest()) ->
    fun(({binary(), lfs_manifest()}, [binary()]) -> [binary()]).
overwritten_UUIDs_active_fold_helper(Active) ->
    fun({UUID, Manifest}, Acc) ->
            update_acc(UUID, Manifest, Acc, Active =:= Manifest)
    end.

update_acc(_UUID, _Manifest, Acc, true) ->
    Acc;
update_acc(UUID, ?MANIFEST{state=active}, Acc, false) ->
    [UUID | Acc];
update_acc(UUID, Manifest=?MANIFEST{state=writing}, Acc, _) ->
    LBWT = Manifest?MANIFEST.last_block_written_time,
    WST = Manifest?MANIFEST.write_start_time,
    acc_leeway_helper(UUID, Acc, LBWT, WST);
update_acc(_, _, Acc, _) -> 
   Acc.

acc_leeway_helper(UUID, Acc, undefined, WST) ->
    acc_leeway_helper(UUID, Acc, WST);
acc_leeway_helper(UUID, Acc, LBWT, _) ->
    acc_leeway_helper(UUID, Acc, LBWT).
```

---

EDIT: fixed in f4838e05e1cacdb37fa8d2a70a97dc20cee15be6

---

reiddraper, line 84: Yes! I very much like it. Will push a change soon with this suggestion.

kellymclaughlin:
Looks good! +1 to merge.

PR #330, opened 2012-11-19
Better handling of anonymous GET Service requests

kellymclaughlin:
Anonymous service-level requests should be denied to avoid unnecessary error messages trying to list the buckets owned by an undefined user.


reiddraper:
woot, before I got HTTP 500, now I get HTTP 403 with `curl -v localhost:8080/`. +1 to merge

PR #332, opened 2012-11-20
Don't use repository snapshots

reiddraper:
This specifies specific revisions and means you
can use the tests offline as long as your maven
cache has been populated (ie. you've run the tests
at least once before)


reiddraper:
Before this PR, `make test-clojure` would spit out a bunch of text about retrieving repos from maven. Now, once you've run the tests at least once, `make test-clojure` should no longer require a net connection and should use only the specified repos.

cmeiklejohn:
:+1:

PR #341, opened 2012-11-28
reorganizing dtrace usage in wm_common & sub resources (for cs310)

jrwest:
Clean up use of dtrace in cs310-wm-resource-reorg branch. 
- move dt_entry & friends to common module
- most calls to dt_entry only in wm_common (a few in sub resources for functions specific to that resource)
- most sub resources do not call riak_cs_wm_dtrace or call one of dt_{service,bucket,object}_{entry,return}


jrwest:
@kellymclaughlin let me know what you think. feel free to merge, if not ill do it w/ a +1. @slfritchie maybe you want to take a look as well?

jrwest:
oh and here is some example output: https://gist.github.com/2c85a52bacdaa9349b32

kellymclaughlin:
We're not limited to using the dtrace function calls only in wm resources. Maybe we should rename `riak_cs_wm_dtrace` to  just `riak_cs_dtrace`. I know there are a few uses of it in `riak_cs_block_server` currently and could be others in the future.

kellymclaughlin:
Other than my last comment, I'm +1 to merging this in.

jrwest:
re: `riak_cs_dtrace`. Good point, I kind of ran into that when trying to move the dtrace op code defines (e.g. `?DT_WM_OP`) out of `riak_cs.hrl`. We have a `riak_cs_dtrace` already, ill just move whats in `riak_cs_wm_dtrace` there, rename a few things and get rid of the webmachine specific dtrace module altogether.

jrwest:
@kellymclaughlin made some changes as suggested. let me know what you think and if the +1 still holds i'll merge it in later today

slfritchie:
I'd like to finish looking at it, should be done later today.

jrwest:
@slfritchie cool, will wait for the +1 from you as well. no rush on getting it merged in.

src/riak_cs_wm_common.erl
```diff
@@ -74,21 +76,27 @@ service_available(Pool, RD, Ctx) ->
...
+                                      valid_entity_length,
+                                      [RD, Ctx],
+                                      ExportsFun(valid_entity_length)),
+    riak_cs_dtrace:dt_wm_return_bool({?MODULE, Mod}, <<"valid_entity_length">>, Valid),
+    R.
 
 forbidden(RD, Ctx=#context{auth_module=AuthMod, submodule=Mod, riakc_pid=RiakPid}) ->
-    dt_entry(Mod, <<"forbidden">>),
+    riak_cs_dtrace:dt_wm_entry({?MODULE, Mod}, <<"forbidden">>),
     {UserKey, AuthData} = AuthMod:identify(RD, Ctx),
```
slfritchie, line 100: Would it be useful to add UserKey and/or AuthData in the probe immediately above?

jrwest, line 100: i'd say yes but i'll defer to you and @kellymclaughlin since you guys know more about how we plan to use these.

kellymclaughlin, line 100: Yeah seems like a good idea to me.

jrwest, line 100: k will add that. @slfritchie I also noticed you have a commented out dtrace op specific to auth, any interest in putting that in and using it in `authenticate/2` and `authorize/2`?

jrwest:
pushed changes that adds `UserKey` to forbidden dtrace entry call (skipped `AuthData` since its just some signed auth data most likely but open to adding that as well if people think it will help). Also fixed an issue with the return call in `authorize/2`.

slfritchie:
Hrm.  Aside from giving every single probe a look-over to see what args might be useful to include in the probe, there's the larger matter of correlation, which I think is the remaining problem.  Also, has the reorganization broken anything substantial wrt Erlang function tracing (for those platforms where DTrace & SystemTap aren't available.)  I haven't finished either of those yet.

"Correlation" meaning being able to correlate related probes to one another.  That correlation thingie needs to be an argument passed into:
1. The first probe in a chain of probes, and then rely on the DTrace `pid` or `tid` variables to correlate later probes, or
2. All probes contain the correlation key.

Option 1 won't work: `pid` is the same for the entire virtual machine, and the thread id `tid` will change whenever the VM's work stealing kicks in to migrate an Erlang process from one scheduler Pthread to another.

So, we need a correlation key in every probe.  The common thread between these probes is the HTTP session that triggered them.  I suppose then we could use any of these as a key:
- HTTP socket port term
- HTTP socket `inet:get_peerinfo()` or whatever func is needed to return the socket's peer's IP address & TCP port number.
- Some unique counter value selected at the beginning of the WM's processing.
- Some random number (from a big key space) selected at the beginning of the WM's processing (and know that correlation bugs could occur sometimes).

Pros & cons of the first two:
- Pro: they contain more info that could also be helpful for debugging (esp. the peer's IP addr)
- Con: they would require a string, and we only have 4 string args to play with.

Pros & cons of the last two:
- Pro: Assuming we fit them into a 62 bit integer (64 bits - 2 bits for tagging), they're perfect as a numeric argument to an Erlang-triggered probe
- Con: integers convey (to humans) less info than a string.

We can store the correlation integer (as a string) or a string using the Erlang pid's DTrace label (via `dyntrace:put_tag/1`).  Or we can put the correlation integer/string as one of the 4 integers/strings in the regular probe args.  I _believe_ that the same Erlang pid is going to be firing all of these probes, so using `dyntrace:put_tag/1` is feasible?  If it is indeed feasible, is there any other use of that tag that that pid might be an even better use?  I can't think of one right now....

jrwest:
this PR, right now, is just trying to make sure there was no info-loss (minus the previous comment, I didn't add any strings/ints to the probes) between the master and the cs310 branch (many of the probes were broken/duplicated/removed during the re-org). Not opposed to addressing the correlation issue here by any means but would it be better to work on that in a separate branch targeted at cs310?

re: erlang function tracing, im not sure what the re-org would have specifically affected (besides maybe making things a little more complicated to trace at first) but maybe there is something I am not aware of when it comes to erlang function tracing and dtrace interacting?

kellymclaughlin:
what about using either a string or integer representation of the erlang pid of the wm process handling the request as the means for correlation?

kellymclaughlin:
+1 to work on a different branch for things that are improvements on the way the probes worked prior to this reorganization.

jrwest:
re: process pid as string, would we want correlation across the webmachine process to something like the fetch object probes (that exist in imagination only right now iirc), which may be in another process?

slfritchie:
My fingers/coffee-lacking brain omitted the Erlang pid of the WM proc, sorry.  It's also a good string type choice, especially if there's only one Erlang pid that's firing these probes.

As app complexity grows, there's no way that we can choose a single tag that is useful to all parts of the app.  In cases were we:

a. Cross a pid boundary
b. Cross a task boundary (which has its own set of related DTrace probes that need correlation)

... then I'm not aware of a good solution other than adding/modifying a probe's args that can correlate both kinds of correlation tags.  Then a D script, if it needs to make a correlation across both kinds of tags can do so.

```
probe_typea1(TagA, blah blah...)
probe_typea2(TagA, blah blah...)
probe_typeb1(TagB, TagA, blah blah)
probe_typeb2(TagB, blah blah)
...
```

kellymclaughlin:
My vote would be use the wm pid for the wm probing for now and let's look at options for correlation as we build out our probing. I don't want to go too far afield with the scope of this work.

slfritchie:
OK, then at the very first WM callback, please tag via `dyntrace:put_tag/1` using `self()` in string form.

kellymclaughlin:
Unless @slfritchie has other input, I'd say merge it in.

slfritchie:
+1 to merge, thanks!

Issue #310, opened 2012-10-25
Webmachine resource reorg


kellymclaughlin:
- Implement URL rewriting proposed [here](https://github.com/basho/riak_cs/wiki/UrlMapping)
- Reorganize existing webmachine resources

Issue #300, opened 2012-10-16
research supporting multipart upload


reiddraper:
create a multipart upload plan

slfritchie:
Reid's strawman:

https://gist.github.com/f700f4c5831ddeba588a

Scott's strawman:

https://gist.github.com/a3e588e4d524dbed2249

slfritchie:
We've agreed to implement Reid's strawman, evolving as necessary to address shortcomings.

@reiddraper Feel free to use this card as you wish, I suppose, closing in favor of new ones or keeping it open.

slfritchie:
The `gh300-multipart-upload` branch has now finished the first of three milestones.
1. DONE: all basic multipart upload functions with Erlang API is done
2. Add WebMachine wiring for HTTP API to step 1's functions
3. Modify to use appropriate strong consistency mechanisms

https://github.com/basho/riak_cs/tree/gh300-multipart-upload

slfritchie:
Item 2, add WebMachine wiring, is now finished and ready for a first review:

https://github.com/basho/riak_cs/pull/369

slfritchie:
PR #369 was closed via merge today, yay!

PR #352, opened 2012-12-05
Webmachine Resource Reorg

jrwest:
implements the URL rewriting proposed here: https://github.com/basho/riak_cs/wiki/UrlMapping and refactoring of webmachine resources. https://github.com/basho/riak_cs/issues/310. Common functionality should now live in `riak_cs_wm_common`. All bucket/object routes serve `wm_common` as the resource, which delegates to "subresources".

Additionally attempts to make auth more pluggable by requiring `riak_cs_auth` behaviours to implement two methods, `identify` and `authenticate`. Authorization is handled by `wm_common` or one of the "subresources" it calls. 

Where things went:
- `wm_service` -> `wm_buckets`
- `wm_bucket` -> `wm_bucket`, `wm_bucket_acl`, `wm_bucket_location`, `wm_bucket_versioning`, `wm_objects`
- `wm_key` -> `wm_object`, `wm_object_acl`


kuenishi:
This PR looks good to me and works fine. If I may, I'd like to send +1

But this is huge patch, so we'd better have a few more +1.

reiddraper:
The old user resource (`/user`) appears to be broken. What do we think about keeping it for backward compatibility of tooling?

reiddraper:
There is a little bit of duplication in calls to `riak_cs_wm_common:resource_call/4`. Usage seems to be like this:

``` erlang
resource_call(Mod, SomeAtom, [Rd, Ctx], ExportsFun(SomeAtom))
```

is there ever a time where the 2nd arg (`SomeAtom`) is not the unary argument to `ExportsFun`? If not, the first simplification would be

``` erlang
resource_call(Mod, SomeAtom, [Rd, Ctx], ExportsFun)
```

and since `ExportsFun` is in the `Ctx`, it could even just be:

``` erlang
resource_call(Mod, SomeAtom, Rd, Ctx)
```

thoughts?

jrwest:
@reiddraper re: `/user` when I discussed it with @kellymclaughlin we decided to drop it (hence the changes to the test suite). Open to revisiting.

reiddraper:
yeah, just not sure if we have any customers or home-grown tooling that is using it at the moment. Personally I'd vote to keep it for one release and deprecate it in the release notes. After it's deprecated, we could even start doing a redirect.

jrwest:
@reiddraper re: `resource_call/4`, good point, probably don't need the whole context (and `RD` and `Ctx` are not always the two arguments -- if any -- passed in) so how about the first version?

``` erlang
resource_call(Mod, SomeAtom, [Rd, Ctx], ExportsFun)
```

jrwest:
@reiddraper as I understood it, it was already deprecated but I could have misunderstood. I'll defer to you and @kellymclaughlin here since you guys are more familiar with the status of things. 

EDIT: as the convo points out, to fix it, we can handle `"/user"` specially in the `rewrite` module I believe. 

here is our original convo fwiw:

```
Kelly McLaughlin
we've deprecated /user in favor of /riak-cs/user so we could just rip that out and only use the latter
Kelly McLaughlin
may still have an issue, but fwiw
Jordan West
yea i have a feeling if i use /riak-cs/user it will work
Jordan West
about to try that
Jordan West
yea /riak-cs/user works
Jordan West
and it does think /user is a bucket
Jordan West
i suppose we can just pattern match on "/user" and similar in the rewrite module
Kelly McLaughlin
i'd just drop the rule for /user and call it done
Jordan West
ok /riak-cs/user it is
```

kellymclaughlin:
My understanding was we did deprecate `/user` as of the last release. We don't really mention any deprecations in our release notes and until last week we weren't actually publishing our release notes anywhere so nobody would even know that, but I did make that publicly known at the last webinar fwiw. I guess it wouldn't hurt to leave it in for another release, but I'd also like to have [this](https://github.com/basho/riak_cs/issues/265) as part of the next release (or soonish thereafter) which would free us to rework our admin API and not worry about taking up useful CS bucket names.

src/riak_cs_wm_object.erl
```diff
@@ -0,0 +1,330 @@
...
+                    UpdCtx = Ctx#context{local_context=UpdLocalCtx},
+                    {[{ContentType, produce_body}], RD, UpdCtx}
+            end;
+       true ->
+            %% TODO this shouldn't ever be called, it's just to
+            %% appease webmachine
+            {[{"text/plain", produce_body}], RD, Ctx}
+    end.
+
+-spec produce_body(#wm_reqdata{}, #context{}) -> {iolist()|binary(), #wm_reqdata{}, #context{}}.
```
reiddraper: dialyzer complains about this spec, likely because of us returning `{{known_length_stream, ...`

src/riak_cs_wm_bucket_acl.erl
```diff
@@ -0,0 +1,157 @@
...
+-include("riak_cs.hrl").
+-include_lib("webmachine/include/webmachine.hrl").
+
+
+%% @doc Get the list of methods this resource supports.
+-spec allowed_methods() -> [atom()].
+allowed_methods() ->
+    ['GET', 'PUT'].
+
+-spec content_types_provided(#wm_reqdata{}, #context{}) -> [{string(), atom()}].
```
reiddraper: should be closer to

``` erlang
-spec content_types_provided(#wm_reqdata{}, #context{}) -> {[{string(), atom()}], #wm_reqdata, #context}.
```

src/riak_cs_wm_bucket_location.erl
```diff
@@ -0,0 +1,71 @@
...
+
+-include("riak_cs.hrl").
+-include_lib("webmachine/include/webmachine.hrl").
+
+%% @doc Get the list of methods this resource supports.
+-spec allowed_methods() -> [atom()].
+allowed_methods() ->
+    ['GET'].
+
+-spec content_types_provided(#wm_reqdata{}, #context{}) -> [{string(), atom()}].
```
reiddraper: same wrong spec as other comment

src/riak_cs_wm_bucket_versioning.erl
```diff
@@ -0,0 +1,75 @@
...
+-include("riak_cs.hrl").
+-include_lib("webmachine/include/webmachine.hrl").
+
+%% @doc Get the list of methods this resource supports.
+-spec allowed_methods() -> [atom()].
+allowed_methods() ->
+    %% TODO: PUT?
+    ['GET'].
+
+-spec content_types_provided(#wm_reqdata{}, #context{}) -> [{string(), atom()}].
```
reiddraper: same other other `content_types_provided` wrong spec

reiddraper:
Will stop updating comments on all of them, but it seems like most of the spec's for `content_types_provided` are wrong (I'm sicking dialyzer on you ;))

reiddraper:
> My understanding was we did deprecate /user as of the last release.

I'm +1 for leaving it as it is then, no need to keep /user

src/riak_cs_utils.erl
```diff
@@ -162,6 +163,17 @@ create_user(Name, Email) ->
             Error
     end.
 
+%% @doc Get the active version of Riak CS to use in checks to
+%% determine if new features should be enabled.
+-spec cs_version() -> pos_integer().
+cs_version() ->
+    cs_version(application:get_env(riak_cs, cs_version)).
```
reiddraper, line 170: should we start setting this as of this PR?

src/riak_cs_web.erl
```diff
@@ -10,26 +10,78 @@
...
-stats_props(AuthBypass) ->
-    [{auth_bypass, AuthBypass}].
+-spec one_three_resources(undefined | pos_integer()) -> [dispatch_rule()].
+one_three_resources(undefined) ->
+    [];
+one_three_resources(Version) when Version < 010300 ->
+    [];
+one_three_resources(_Version) ->
+    [
+    ].
```
reiddraper: `[]`

reiddraper:
Other than the above comments, this looks really great. Definitely appreciate the simplified resources. Have you guys put any thought into squashing some of the commits once we're ready to merge?

jrwest:
pushed some changes to address above comments. @reiddraper re: squashing some commits, we probably should but I haven't put much thought into what they squash down to yet.

src/riak_cs_wm_object.erl
reiddraper: this one is still unaddressed

reiddraper:
84423c467e19ecbc4ea6935067582930ec5be83f looks great. I put one more comment about a spec in wm_object:147 that still needs updating, but looking good

reiddraper:
(ps. i'm removing `deps/*/ebin` from `make dialyzer` to get it to run)

reiddraper:
other than whatever squashing @jrwest and @kellymclaughlin want to do, I'm +1 to merge. Excellent work!

slfritchie:
Just to let y'all know that I'll be getting pretty antsy starting tomorrow ... Please merge this & the other WM reorg PR quickly, I'm ready to start doing the WM work for multipart uploads.

https://github.com/basho/riak_cs/issues/300#issuecomment-11320721

jrwest:
should have updated this, talked offline. We are going to start squashing the branch as @reiddraper laid out. @kellymclaughlin has round one.

reiddraper:
@slfritchie I'd say go ahead and wire it up to the current wm modules then. IMHO we should be doing as little possible in the webmachine callbacks anyway, so reworking to the new org should be trivial. Just my 2c.

jrwest:
rebased and merged. closing.

PR #338, opened 2012-12-10
riak_test support scripts

kellymclaughlin:
- Replace existing `riak_test` support scripts to build releases of `riak_cs` for running `riak_test` suites.
- Update the existing `riak_test` support module `rtcs.erl`.
- Update the replication test `repl_test.erl`.
- Update the `README` and convert to markdown format.


kellymclaughlin:
Related `stanchion` pr is [here](https://github.com/basho/stanchion/pull/33)

kellymclaughlin:
~~For the test summary to be printed properly, you'll need the branch of `riak_test` referenced by [this](https://github.com/basho/riak_test/pull/123) pr.~~ Merged to master.

riak_test/bin/rtdev-setup-releases.sh
```diff
@@ -0,0 +1,18 @@
...
+# using rtdev-mixed.config settings. Should be run inside a directory
+# that contains devrels for prior Riak CS releases. Easy way to create this
+# is to use the rtdev-build-releases.sh script
+
+rm -rf /tmp/rtcs
+mkdir /tmp/rtcs
+for rel in */dev; do
+    vsn=$(dirname "$rel")
+    mkdir "/tmp/rtcs/$vsn"
+    cp -a "$rel" "/tmp/rtcs/$vsn"
```
jrwest: apparently Solaris doesn't support `-a` should use `-p -P -R` instead like riak test? https://github.com/basho/riak_test/commit/765779f830c6c91054729f0704f12b5f29de61d3

.plugins/rebar_test_plugin.erl
```diff
@@ -54,6 +56,18 @@ int_test_run(Config, AppFile) ->
         _ -> test_run(int_test, Config, AppFile)
     end.
 
+riak_test_clean(Config, AppFile) ->
+    case should_i_run(Config) of
+        false -> ok;
+        _ -> test_clean(client_test, Config, AppFile)
```
jrwest: think the first argument should be `riak_test` instead of `client_test`. Also, any reason there isn't a make target for this as well?

Makefile
```diff
@@ -22,6 +22,9 @@ compile-client-test:
 compile-int-test:
 	@./rebar int_test_compile
 
+compile-riak-test:
```
jrwest: if you run `make clean` and `rebar riak_test_clean` (assuming changes made in comment above) then `make compile-riak-test` will crash rebar. Making this target depend on `compile` or `all` stops it from happening but not sure its worth changing, just something i came across.

riak_test/src/rtcs.erl
```diff
@@ -75,31 +86,35 @@ deploy_nodes(NumNodes, InitialConfig) ->
...
                             Config)),
-                    update_cs_config(rt:config(csroot), N,
+                    update_cs_config(rt:config(?CS_CURRENT), N,
                         proplists:get_value(cs, Config)),
-                    update_stanchion_config(rt:config(csroot),
+                    update_stanchion_config(rt:config(?STANCHION_CURRENT),
                         proplists:get_value(stanchion, Config));
                 ({{_CSNode, RiakNode}, Config}) ->
                     N = rtdev:node_id(RiakNode),
+                    lager:info("Config2: ~p~n~p", [Config, N]),
```
jrwest, line 98: these were actually really useful when debugging the issue w/ building a `devrel` vs. `stagedevrel` but for normal test runs they seem verbose. What do you think about trying to hide these so they are only printed when running `riak_test` with `-v`? 

EDIT: by "these", i mean this one and the similar log message a few lines above (L89).

jrwest:
was able to get everything up and running and `repl_test` is passing (see note below). Left a couple comments but it looks good. Perhaps someone else should also review the `repl_test` specifics, though, since I'm still getting familiar with repl.

note: if you plan to run the `repl_test` i had to rebase onto (or merge) master to since `velvet` is now in a separate repo.

riak_test/src/rtcs.erl
kellymclaughlin, line 98: From what I can tell `riak_test` only uses lager debug level during test execution and the `verbose` option is only used to determine what to output for the test summary. So I do not think it would actually be possible with the way `riak_test` currently works.

jrwest, line 98: ah ok, in that case i'd say leave it since its useful for debugging

jrwest:
+1

PR #354, opened 2012-12-12
Made webmachine log directory configurable through parameter http_logdir

cdahlqvist:
I received a client request to make the Riak-CS http log directory configurable (ZenDesk ticket 3434). I updated riak_cs_sup.erl to look for environment variable 'http_logdir' (seemed to follow convention used in riak_core). Is this an acceptable solution?

The client is using Riak-CS 1.2.1 and have requested a patch.


reiddraper:
+1 to merge.

PR #305, opened 2012-12-14
riak_test regression test debt

kellymclaughlin:
Add `riak_test` regression tests for the following issues:
- [403 instead of 404 returned when trying to list nonexistent bucket](https://github.com/basho/riak_cs/issues/296)
- [No response body in 404 to the bucket that have never been created once](https://github.com/basho/riak_cs/issues/347)

_Note:_ At the time of opening this pull request `cs347_regression_test` is known to fail due to the following open issues:
- ~~[Resource element content in error document is not S3 API resource](https://github.com/basho/riak_cs/issues/359)~~ Merged

This pr is against `master`, but is dependent on [this](https://github.com/basho/riak_cs/issues/338) pr.


riak_test/tests/cs347_regression_test.erl
```diff
@@ -0,0 +1,55 @@
...
+    rt:wait_until_ring_converged(RiakNodes),
+
+    %% STFU sasl
+    application:load(sasl),
+    application:set_env(sasl, sasl_error_logger, false),
+    erlcloud:start(),
+
+    FirstNode = hd(RiakNodes),
+
+    {AccessKeyId, SecretAccessKey} = rtcs:create_user(FirstNode, 1),
```
jrwest: what do you think about refactoring the similar setup code between the two regression tests into functions in `rtcs`, similar to `rt:build_cluster` (https://github.com/basho/riak_test/blob/master/src/rt.erl#L733)? Maybe a separate function to setup `erlcloud` and turn off SASL error logging?

kellymclaughlin: I dig it, changes coming up

riak_test/tests/cs347_regression_test.erl
```diff
@@ -0,0 +1,55 @@
...
+%% @doc Regression test for `riak_cs' <a href="https://github.com/basho/riak_cs/issues/347">
+%% issue 347</a>. The issue description is: No response body in 404 to the
+%% bucket that have never been created once.
+
+-export([confirm/0]).
+-include_lib("eunit/include/eunit.hrl").
+
+-define(TEST_BUCKET, "riak_test_bucket").
+%% "Object Not Found" is the non-standard error message generated by inets
+-define(NOT_FOUND, {aws_error,{http_error,404,"Object Not Found","<?xml version=\"1.0\" encoding=\"UTF-8\"?><Error><Code>NoSuchBucket</Code><Message>The specified bucket does not exist.</Message><Resource>/riak_test_bucket</Resource><RequestId></RequestId></Error>"}}).
```
jrwest: i tend to shy away from exact string matching of full response bodies. maybe not worth it since this is pretty simple, well-formed/documented XML but any reason not to parse it, and check, for example, that `Resource` actually matches the requested path, vs. hardcoding it, etc?

riak_test/tests/repl_test.erl
```diff
@@ -154,141 +93,131 @@ confirm() ->
...
 
-    erlcloud_s3:put_object("riak_test_bucket", "object_three", Object3),
+    erlcloud_s3:put_object(?TEST_BUCKET, "object_three", Object3, U1C1Config),
 
     lager:info("disconnect the clusters"),
     LeaderB = rpc:call(hd(BNodes), riak_repl_leader, leader_node, []),
-    replication:del_site(LeaderB, "site1"),
+    repl_helpers:del_site(LeaderB, "site1"),
+
+    timer:sleep(5000),
```
bookshelfdave, line 133: An RPC call to riak_repl_ring:get_site/2 might work better than a sleep

bookshelfdave:
@kellymclaughlin repl tests look good, very readable now :-)

riak_test/tests/cs347_regression_test.erl
kellymclaughlin: Good idea, pushed change with that.

jrwest:
+1

PR #359, opened 2012-12-14
Resource element content in error document is not S3 API resource

kellymclaughlin:
After the change to rewrite the S3 URLs the rewritten path is being used as the content for the Resource element in S3 error responses.  Change to using the original resource with any trailing slashes removed instead.

The `cs347_regression_test` failure should be fixed by this change.


jrwest:
+1

PR #346, opened 2012-12-17
riak-cs-gc does not accept "set-interval infinity"

kellymclaughlin:
around here
https://github.com/basho/riak_cs/blob/master/src/riak_cs_gc_console.erl#L53

### DIFF
--- a/src/riak_cs_gc_console.erl
+++ b/src/riak_cs_gc_console.erl
@@ -74,6 +74,9 @@ resume() ->
 
 set_interval(undefined) ->
     output("Error: No interval value specified");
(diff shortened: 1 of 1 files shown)


kuenishi:
worked fine to me. +1

```
$ sbin/riak-cs-gc set-interval infinity 
The garbage collection interval was updated.
$ sbin/riak-cs-gc status
There is no garbage collection in progress
  The current garbage collection interval is: infinity (i.e. gc is disabled)
  Last run started at: undefined
  Next run scheduled for: 00421218T051733Z
```

Issue #344, opened 2012-11-29
IP filtering by bucket policy


kuenishi:

michellep-basho:
adding 1.3 milestone for tracking purposes

PR #364, opened 2012-12-19
add bucket policy for IP address filtering (issue #344)

kuenishi:
rebar.config refers to stanchion's bucket_policy branch, this will be fixed as 'master' after stanchion's issue # 35 get fixed

### DIFF
--- a/src/riak_cs_s3_policy.erl
+++ b/src/riak_cs_s3_policy.erl
@@ -1,0 +1,709 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2007-2012 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc policy utility functions
+
+-module(riak_cs_s3_policy).
+
+-behaviour(riak_cs_policy).
+
+-include("riak_cs.hrl").
+-include("s3_api.hrl").
+-include_lib("webmachine/include/wm_reqdata.hrl").
+
+-ifdef(TEST).
+
+-include_lib("eunit/include/eunit.hrl").
+
+-endif.
+
+%% Public API
+-export([
+         eval/2,
+         check_policy/2,
+         reqdata_to_access/3,
+         policy_from_json/1,
+         policy_to_json_term/1,
+         supported_object_action/0,
+         supported_bucket_action/0,
+         log_supported_actions/0
+        ]).
+
+-type policy1() :: ?POLICY{}.
+
+-export_type([policy1/0]).
+
+-define(AMZ_DEFAULT_VERSION, <<"2008-10-17">>).
+
+%% ===================================================================
+%% Public API
+%% ===================================================================
+
+-spec eval(access(), policy() | undefined | binary() )-> boolean().
+eval(_, undefined) -> undefined;
+eval(Access, JSON) when is_binary(JSON) ->
+    eval(Access, policy_from_json(JSON));
+eval(Access, ?POLICY{version=V, statement=Stmts}) ->
+    case V of
+        undefined ->        aggregate_evaluation(Access, Stmts);
+        ?AMZ_DEFAULT_VERSION -> aggregate_evaluation(Access, Stmts);
+        _ -> false
+    end.
+
+aggregate_evaluation(_, []) -> undefined;
+aggregate_evaluation(Access, [Stmt|Stmts]) ->
+    case eval_statement(Access, Stmt) of
+        undefined -> aggregate_evaluation(Access, Stmts);
+        true  -> true;
+        false -> false
+    end.
+
+
+% @doc  semantic validation of policy
+-spec check_policy(access(), policy()) -> ok | {error, atom()}.
+check_policy(#access_v1{bucket=B} = _Access,
+             Policy) ->
+    
+    case check_all_resources(B, Policy) of
+        false -> {error, malformed_policy_resource};
+        true ->
+            case check_principals(Policy?POLICY.statement) of
+                false -> {error, malformed_policy_principal};
+                true -> 
+                    case check_actions(Policy?POLICY.statement) of
+                        false -> {error, malformed_policy_action};
+                        true -> ok
+                    end
+            end
+    end.
+
+% @doc confirm if forbidden action included in policy
+% s3:CreateBucket and s3:ListAllMyBuckets are prohibited at S3
+check_actions([]) -> true;
+check_actions([Stmt|Stmts]) ->
+    case Stmt#statement.action of
+        '*' -> check_actions(Stmts);
+        Actions ->
+            case not lists:member('s3:CreateBucket', Actions) of
+                true ->
+                    case not lists:member('s3:ListAllMyBuckets', Actions) of
+                        true -> check_actions(Stmts);
+                        false -> false
+                    end;
+                false -> false
+            end
+    end.
+
+check_principals([]) -> false;
+check_principals([Stmt|Stmts]) ->
+    case check_principal(Stmt#statement.principal) of
+        true -> true;
+        false -> check_principals(Stmts)
+    end.    
+
+check_principal('*') ->
+    true;
+check_principal([]) ->
+    false;
+check_principal([{canonical_id, _Id}|_]) -> %% TODO: do we check all canonical ids exist?
+    true;
+check_principal([{aws, '*'}|_]) ->
+    true;
+check_principal([_|T]) ->
+    check_principal(T).
+
+% @doc check if the policy is set to proper bucket by checking arn
+check_all_resources(BucketBin, ?POLICY{statement=Stmts} = _Policy) ->
+    CheckFun = fun(Stmt) ->
+                       check_all_resources(BucketBin, Stmt)
+               end,
+    lists:all(CheckFun, Stmts);
+check_all_resources(BucketBin, #statement{resource=Resources} = _Stmt) ->
+    CheckFun = fun(Resource) ->
+                       check_all_resources(BucketBin, Resource)
+               end,
+    lists:all(CheckFun, Resources);
+check_all_resources(BucketBin, #arn_v1{path=Path} = _Resource) ->
+    [B|_] = string:tokens(Path, "/"),
+    B =:= binary_to_list(BucketBin).
+
+-spec reqdata_to_access(#wm_reqdata{}, Target::atom(), ID::binary()) -> {ok, policy()}.
+reqdata_to_access(RD, Target, ID) ->
+    Key = case wrq:path_info(object, RD) of
+              undefined -> undefined;
+              RawKey -> mochiweb_util:unquote(mochiweb_util:unquote(RawKey))
+          end,
+    #access_v1{
+       method = wrq:method(RD), target = Target,
+       id = ID, req = RD,
+       bucket = list_to_binary(wrq:path_info(bucket, RD)),
+       key    = Key
+      }.
+
+-spec policy_from_json(JSON::binary()) -> policy().
+policy_from_json(JSON)->
+    {struct, Pairs} = mochijson2:decode(JSON),
+    Version = proplists:get_value(<<"Version">>, Pairs),
+    ID      = proplists:get_value(<<"Id">>, Pairs),
+    Stmts0  = proplists:get_value(<<"Statement">>, Pairs),
+    Stmts = lists:map(fun({struct,S})->statement_from_pairs(S, #statement{}) end, Stmts0),
+    ?POLICY{id=ID, version=Version, statement=Stmts}.
+
+-spec policy_to_json_term(policy()) -> JSON::binary().
+policy_to_json_term( ?POLICY{ version = V,
+                                 id = ID, statement = Stmts0} ) when
+      V =:= ?AMZ_DEFAULT_VERSION orelse V =:= undefined ->
+    Stmts = lists:map(fun statement_to_pairs/1, Stmts0),
+    % hope no unicode included
+    Policy0 = [{"Id", ID}, {"Statement",Stmts}],
+    Policy = case V of
+                 undefined ->  Policy0;
+                 ?AMZ_DEFAULT_VERSION -> [{"Version", V}|Policy0]
+             end,
+    list_to_binary(mochijson2:encode(Policy)).
+
+-spec supported_object_action() -> [s3_object_action()].
+supported_object_action() -> ?SUPPORTED_OBJECT_ACTION.
+
+-spec supported_bucket_action() -> [s3_bucket_action()].
+supported_bucket_action() -> ?SUPPORTED_BUCKET_ACTION.
+
+% @doc put required atoms into atom table
+% to make policy json parser safer by using erlang:binary_to_existing_atom/2.
+-spec log_supported_actions() -> ok.
+log_supported_actions()->
+    _ = lager:info("supported object actions: ~p",
+                   [lists:map(fun atom_to_list/1, supported_object_action())]),
+    _ = lager:info("supported bucket actions: ~p",
+                   [lists:map(fun atom_to_list/1, supported_bucket_action())]),
+    ok.
+
+%% ===================================================================
+%% internal API
+
+-spec resource_matches(binary(), binary(), #statement{}) -> boolean().
+resource_matches(_, _, #statement{resource='*'} = _Stmt ) -> true;
+resource_matches(BucketBin, KeyBin, #statement{resource=Resources}) ->
+    Bucket = binary_to_list(BucketBin),
+    % @TODO in case of using unicode object keys
+    Path0 = case KeyBin of
+                undefined -> Bucket;
+                Key when is_list(KeyBin) ->
+                    binary_to_list(BucketBin) ++ "/" ++ Key;
+                _ when is_binary(KeyBin) ->
+                    binary_to_list(<<BucketBin/binary, "/", KeyBin/binary>>)
+            end,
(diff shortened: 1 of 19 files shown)


kellymclaughlin:
You may have noticed that we broke the stanchion client into a separate repo since that was the only module from stanchion that CS needed and to help with some planned refactoring. Another reason is to ensure that CS and Stanchion remain as loosely coupled as possible. Just making sure you are aware of that in case you want to rebase to account for it sooner rather than later.

kellymclaughlin:
I see you are calling the `stanchion_policy` module directly so my last comment might be confusing. For things that we need to use in both CS and Stanchion I think we should move to an approach of pushing those things into separate repos and using them as separate dependencies. In this case I would probably move toward creating an `s3_bucket_policy` repo set up very much like the new `velvet` repo is (_i.e._ It just builds the source files and is not concerned with being a proper OTP app or building releases) and put as much of the bucket policy code there and have it as a dependency for CS and Stanchion. We want to do the same thing with the large file manifest code that is now copied in CS and Stanchion among other parts. What do you think of this approach?

kuenishi:
I agree and really like that way of thinking - but my suggestion is a bit different. I completely removed s3 specific codes from stanchion and now every s3 specific is included in riak_cs PR.
And, for best refactoring, the separated repository would include all S3 specific codes ( including riak_cs_s3_auth, s3_response, and so on) . Its name would be something like `riak_cs_s3` or so.

Other common codes shared-by-copying by CS and Stanchion (like Large file manifest) - I don't have better idea than yours. It would be good to separate these code into more fine-grained repositories, but let me mention another smaller problem: it may make us confusing when making PRs. If we're creating another repo for refactoring, for keeping dependency graph simple, velvet's code may be all included in riak_cs. Because riak_cs is the only repo that depends on velvet.

kuenishi:
related PRs:

https://github.com/basho/stanchion/pull/35
https://github.com/basho/velvet/pull/1

reiddraper:
`<<"2008-10-17">>` appears a few times in this PR, maybe we should `-define ...` it in one of the `.hrl` files?

kuenishi:
@reiddraper It's AWS specific constant, so defining it in riak_cs_s3_policy.erl would be good.

src/riak_cs_s3_policy.erl
```diff
@@ -0,0 +1,695 @@
...
+                                 id = ID, statement = Stmts0} ) when
+      V =:= ?AMZ_DEFAULT_VERSION orelse V =:= undefined ->
+    Stmts = lists:map(fun statement_to_pairs/1, Stmts0),
+    % hope no unicode included
+    Policy0 = [{"Id", ID}, {"Statement",Stmts}],
+    Policy = case V of
+                 undefined ->  Policy0;
+                 ?AMZ_DEFAULT_VERSION -> [{"Version", V}|Policy0]
+             end,
+    list_to_binary(lists:flatten(mochijson2:encode(Policy))).
```
kellymclaughlin: I would recommend replacing `lists:flatten` with `iolist_to_binary` here.

src/riak_cs_s3_policy.erl
```diff
@@ -0,0 +1,695 @@
...
+                Key when is_list(KeyBin) ->
+                    binary_to_list(BucketBin) ++ "/" ++ Key;
+                _ when is_binary(KeyBin) ->
+                    binary_to_list(<<BucketBin/binary, "/", KeyBin/binary>>)
+            end,
+    lists:any(fun(#arn_v1{path="*"}) ->    true;
+                 (#arn_v1{path=Path}) ->
+                      case Path of
+                          Bucket -> true;
+                                                % only prefix matching
```
kellymclaughlin: It's not clear what this comment pertains to since it is floating out at the end of an empty line.

src/riak_cs_s3_policy.erl
```diff
@@ -0,0 +1,695 @@
...
+    statement_from_pairs(T, Stmt#statement{effect=allow});
+statement_from_pairs([{<<"Effect">>,<<"Deny">>}|T], Stmt) ->
+    statement_from_pairs(T, Stmt#statement{effect=deny});
+
+statement_from_pairs([{<<"Principal">>,P}  |T], Stmt) ->
+    Principal = parse_principal(P),
+    statement_from_pairs(T, Stmt#statement{principal=Principal});
+
+statement_from_pairs([{<<"Action">>,As}    |T], Stmt) ->
+    Atoms = lists:map(fun binary_to_action/1, As),
```
kellymclaughlin, line 432: Actions are not required to be lists so this line fails for me when using a valid action like: `"Action": "s3:*"`.

src/riak_cs_wm_objects.erl
```diff
@@ -42,35 +43,64 @@ authorize(RD, #context{user=User,
...
+
+                                undefined ->
+                                    riak_cs_wm_utils:deny_access(AccessRD, PermCtx);
+                                false ->
+                                    %% no facility for logging bad access
+                                    %% against unknown actors
+                                    riak_cs_wm_utils:deny_access(AccessRD, PermCtx)
+                            end;
+
+                        _ ->
```
kellymclaughlin: Need to also evaluate the policy for this case since it is where the user making the request is not the owner or a user with an ACL grant, but they may still be allowed access by the policy.

src/riak_cs_s3_policy.erl
```diff
@@ -0,0 +1,695 @@
...
+    
+
+
+% functions to eval:
+-spec eval_statement(access(), policy()) -> boolean() | undefined.
+eval_statement(#access_v1{method=M, target=T, req=Req, bucket=B, key=K} = _Access,
+               #statement{effect=E, condition_block=Conds, action=As} = Stmt) ->
+    {ok, A} = make_action(M, T),
+    ResourceMatch = resource_matches(B, K, Stmt),
+    IsRelated = lists:member(A, As) orelse lists:member('*', As),
```
kellymclaughlin, line 221: This check is insufficient since the Actions in the policy may not be a list and there are other valid wildcard conditions, though in our case just additionally checking `'s3:*'` is probably sufficient.

src/riak_cs_s3_policy.erl
```diff
@@ -0,0 +1,695 @@
...
+eval_numeric_lt(_, _) -> throw(not_supported).
+eval_numeric_lte(_, _) -> throw(not_supported).
+
+eval_date_eq(_, _) -> throw(not_supported).
+eval_date_lt(_, _) -> throw(not_supported).
+eval_date_lte(_, _) -> throw(not_supported).
+
+-spec eval_ip_address(#wm_reqdata{}, [{'aws:SourceIp', binary()}]) -> boolean().
+eval_ip_address(Req, Conds) ->
+    {Peer, _} = parse_ip(Req#wm_reqdata.peer),
```
reiddraper, line 331: If riak-cs is behind a load balancer or reverse proxy, we should be looking for the IP address from a header, like [X-Forwarded-For](http://en.wikipedia.org/wiki/X-Forwarded-For). How about two app.config settings

``` erlang
use_remote_addr_for_ip=true :: boolean().
remote_addr_header :: string().
```

thoughts?

kellymclaughlin:
I'm wondering how we should handle `Principal` specification. Amazon uses a JSON object as the value where `AWS` is the key and a single ARN or a list of ARNs is set as the object value. The ARN uses a user's AWS account number which is not a concept we currently have in Riak CS. You may alternatively specify a user's canonical id, but support for that is not part of this change set yet.

A few options occur to me on how to handle this:
1. Add an account concept that can be a container for multiple users. This to me would seem to put our schedule for delivery in serious jeopardy.
2. Use the ARN scheme as-is, but just use `RCS` instead of `AWS` for the key and the user's `key_id` instead of account number.
3. Only accept the `CanonicalUser` specification for `Principal`. The only odd thing here is that S3 translates the canonical id to an account id and stores the policy using a normal ARN. It ismore efficient for us to store the user's `key_id` for direct lookup so this essentially becomes an extension of the previous option.
4. Use a custom scheme similar to ARN, but that is more tailored for our needs.

Anyone have thoughts on these options or any other ideas?

src/riak_cs_s3_policy.erl
kuenishi: Do you mean replacing like this? I'll do that at next commit.

```
    iolist_to_binary(lists:flatten(mochijson2:encode(Policy))).
```

src/riak_cs_s3_policy.erl
kuenishi: By this comment I wanted to mean that:
only supports prefix matching like /path/to/there which matches /path/to/\* and not supporting partial matching like /path/to/over/there -> /path/*/there

... but this logic includes some bugs that matches /path/to/there with /path/*/here. I'll fix it.

src/riak_cs_s3_policy.erl
kuenishi, line 432: Can I cut this out to another ticket? This is out of what is required till next feature, and it is not bug but just not-yet-implemented  thing.

src/riak_cs_s3_policy.erl
kuenishi, line 221: Same as above, can I cut this out to another new ticket?

src/riak_cs_s3_policy.erl
kuenishi, line 331: Agreed. I just didn't hit on that. But, I prefer more fine-grained task granularity, so I'd like to work on this as another issue after this long-running branch merged.

kuenishi:
> You may alternatively specify a user's canonical id, but support for that is not part of this change set yet.

@kelly I thought I had included support for canonical ID at riak_cs_s3_policy:check_principals/2 ... And I thought it worked.

And I like option #3 with using key_id looks more suitable to us and our customers' use case. So far it will be enough for us, and to think of KeyStone integration, there would be much smarter solution to make it more loosely-coupled.

src/riak_cs_wm_objects.erl
kuenishi: Oh, I need fix this. Let me update the PR.

src/riak_cs_s3_policy.erl
kellymclaughlin, line 432: Yep it's fine with me if you want to make a new ticket for this. I like finer grained tasks as well.

src/riak_cs_s3_policy.erl
kellymclaughlin, line 221: +1 to new ticket for this also

src/riak_cs_s3_policy.erl
kellymclaughlin: I think the comment is fine, but it is the preceding whitespace on that line I was referring to. It makes it look like the comment is hanging out in space with no real attachment. I get this problem occasionally in emacs when using auto-ident because it treats `%` comments differently from `%%` comments. Not sure if you're also an emacs user and if it could be the problem here too.

src/riak_cs_s3_policy.erl
kellymclaughlin: Actually replacing `list_to_binary(lists:flatten())` with `iolist_to_binary` was what I meant. `mochijson2:encode` returns an iolist type so that should be sufficient and I think more efficient that `lists:flatten`.

src/riak_cs_s3_policy.erl
```diff
@@ -0,0 +1,695 @@
...
+
+
+% @doc  semantic validation of policy
+-spec check_policy(access(), policy()) -> ok | {error, atom()}.
+check_policy(#access_v1{bucket=B,
+                        id=Id} = _Access,
+             Policy) ->
+    
+    case check_all_resources(B, Policy) of
+        false -> {error, malformedpolicy_resource};
```
kellymclaughlin: This is a subjective observation, but I find myself wanting all cases of `malformedpolicy` to instead be `malformed_policy`.

src/riak_cs_s3_policy.erl
```diff
@@ -0,0 +1,695 @@
...
+parse_principal({struct, List}) when is_list(List) ->
+    parse_principals(List, []).
+
+
+parse_principals([], Principal) -> Principal;
+parse_principals([{<<"AWS">>,[<<"*">>]}|TL], Principal) ->
+    parse_principals(TL, [{aws, '*'}|Principal]);
+parse_principals([{<<"AWS">>,<<"*">>}|TL], Principal) ->
+    parse_principals(TL, [{aws, '*'}|Principal]);
+parse_principals([{<<"AWS">>,CanonicalIds}|TL], Principal) ->
```
kellymclaughlin: I retested using the `CanonicalUser` principal specification and I think `<<"AWS">>` here should be replaced with `<<"CanonicalUser">>`.

src/riak_cs_s3_policy.erl
```diff
@@ -0,0 +1,695 @@
...
+    case check_principal(Id, Stmt#statement.principal) of
+        true -> true;
+        false -> check_principals(Id, Stmts)
+    end.    
+
+check_principal(_, '*') ->
+    true;
+check_principal(_, []) ->
+    false;
+check_principal(Id, [{canonical_id, Id}|_]) ->
```
kellymclaughlin: This line also causes issue when specifying `CanonicalUser` for a principal since it is only valid if the canonical id of the bucket owner is the same as that of the entity specified by the principal which should rarely happen in practice.

src/riak_cs_s3_policy.erl
reiddraper: +1

kuenishi:
PRs updated, and he're others related to this PR. https://github.com/basho/stanchion/pull/35 https://github.com/basho/velvet/pull/1

I'll update rebar.config specifying the bucket_policy branch and squash the whole dirty commit logs after +1 given.

src/riak_cs_s3_policy.erl
kellymclaughlin: I think this must still be addressed before things will work properly with the `CanonicalUser`, at least in most practical use cases.

kuenishi: Yes, as I re-thought how it should be, and my suggestion is not to support CanonicalUser principal but just "AWS" principal with "*" (all). I'll ask my customers do they really need this feature again, with the details of this problem including things Kelly mentioned. If then don't care about this, I'll leave this as unimplemented.

src/riak_cs_wm_objects.erl
```diff
@@ -42,35 +43,71 @@ authorize(RD, #context{user=User,
...
+                            %% log bad requests against the actors
+                            %% that make them
+                            AccessRD = riak_cs_access_logger:set_user(User, RD),
```
kellymclaughlin: `AccessRD` has already been assigned and the comparison fails here. We should probably assign `Access` and `AccessRD` inside the individual case clauses for the `User` case statement at [L73](https://github.com/basho/riak_cs/pull/364/files#L16R73).

kellymclaughlin:
Once granting access for all actions (`["*"]`) to all users with the `Principal` value of `{"AWS", ["*"]}` I cannot get the location information (which is used by `s3cmd info`) for the bucket using another user because `riak_cs_wm_bucket_location` has not been changed to check the bucket policy.

kellymclaughlin:
I wanted to also share the policy that I have been testing with in case you want to verify your fixes with what I am using. I have a bucket simply called `bucket`.

```
{
  "Id": "Policy1357229988414",
  "Statement": [
    {
      "Sid": "Stmt1357229986236",
      "Action": ["*"],
      "Effect": "Allow",
      "Resource": ["arn:aws:s3:::bucket", "arn:aws:s3:::bucket/*"],
      "Principal": {
        "AWS": ["*"]
      }
    }
  ]
}
```

kuenishi:
@kelly I did a rather large refactor and fixed that problem by making a new helper function `riak_cs_wm_utils:bucket_access_authorize_helper/4` to handle bucket access call from four `authorise/2` .

src/riak_cs_wm_objects.erl
kuenishi: this is fixed by latest commit, which is also included in `riak_cs_wm_utils:bucket_access_authorise_helper/4` . If you know a much better name of this function, please let me know.

kellymclaughlin:
I very much like that refactoring. Thanks!

kellymclaughlin:
`s3cmd info` is working now

kellymclaughlin:
I see two more modules that need to be changed to check the policy in the case where the user requesting access is not the owner, but may fall under the policy: `riak_cs_wm_object` and `riak_cs_wm_object_acl`.

kellymclaughlin:
I think we can use your `bucket_access_authorize_helper` refactoring to eventually remove the `authorize` function from the resource submodules and have one call in `riak_cs_wm_common`. That has been on my `TODO` list, but I think it is more appropriate as a separate pull request since we are close to finishing this one.

src/riak_cs_s3_policy.erl
reiddraper, line 331: Works for me, just gotta make sure we don't forget about it.

kuenishi:
@kelly I did the same styled refactoring to `riak_cs_wm_object` and `riak_cs_wm_object_acl` with fixing the problem you mentioned.

BTW I still don't understand your idea of removing `authorize` function. It should be separate discussion, though, I prefer current style of helper function style, to implicitly-inherited callback style.

kellymclaughlin:
Latest changes look good. All policy enforcement seems to work as expected and automated tests pass. I haven't finished a `riak_test` module yet, but I think we should proceed and implement that test separately. 

+1 to merge.

kuenishi:
squashed.

PR #367, opened 2012-12-20
URL rewriting fails for buckets containing '.'

kellymclaughlin:
Properly extract bucket names that contain periods from the host header

### DIFF
--- a/src/riak_cs_s3_rewrite.erl
+++ b/src/riak_cs_s3_rewrite.erl
@@ -15,33 +15,35 @@
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
+-compile(export_all).
 -endif.
 
 -type subresource() :: {string(), string()}.
 -type subresources() :: [subresource()].
 
 %% @doc Function to rewrite headers prior to processing by webmachine.
--spec rewrite(atom(), atom(), {integer(), integer()}, gb_tree(), string()) -> 
+-spec rewrite(atom(), atom(), {integer(), integer()}, gb_tree(), string()) ->
                      {gb_tree(), string()}.
 rewrite(Method, _Scheme, _Vsn, Headers, RawPath) ->
     riak_cs_dtrace:dt_wm_entry(?MODULE, <<"rewrite">>),
     Host = mochiweb_headers:get_value("host", Headers),
     HostBucket = bucket_from_host(Host),
-    {Path, QueryString, _} = mochiweb_util:urlsplit_path(RawPath),    
+    {Path, QueryString, _} = mochiweb_util:urlsplit_path(RawPath),
     RewrittenPath = rewrite_path(Method, Path, QueryString, HostBucket),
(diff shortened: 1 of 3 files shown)


src/riak_cs_s3_rewrite.erl
```diff
@@ -15,33 +15,35 @@
...
     Host = mochiweb_headers:get_value("host", Headers),
     HostBucket = bucket_from_host(Host),
-    {Path, QueryString, _} = mochiweb_util:urlsplit_path(RawPath),    
+    {Path, QueryString, _} = mochiweb_util:urlsplit_path(RawPath),
     RewrittenPath = rewrite_path(Method, Path, QueryString, HostBucket),
-    RewrittenHeaders = mochiweb_headers:default(?RCS_REWRITE_HEADER, 
+    RewrittenHeaders = mochiweb_headers:default(?RCS_REWRITE_HEADER,
                                                 rcs_rewrite_header(RawPath, HostBucket),
                                                 Headers),
+    lager:info("Hdrs: ~p Path: ~p", [RewrittenHeaders, RewrittenPath]),
```
slfritchie: This lager call looks like a debugging leftover?

src/riak_cs_s3_rewrite.erl
```diff
@@ -85,18 +87,20 @@ bucket_from_host(HostHeader) ->
     riak_cs_dtrace:dt_wm_entry(?MODULE, <<"bucket_from_host">>),
     HostNoPort = hd(string:tokens(HostHeader, ":")),
     {ok, RootHost} = application:get_env(riak_cs, cs_root_host),
-    extract_bucket_from_host(string:tokens(HostNoPort, "."),
-                             string:tokens(RootHost, ".")).
+    extract_bucket_from_host(HostNoPort,
+                             string:rstr(HostNoPort, RootHost)).
```
slfritchie: Please look over commit e67288c for a bugfix here: use `rstr()` instead of `str()`.

test/riak_cs_s3_rewrite_eqc.erl
```diff
@@ -0,0 +1,84 @@
...
+%% ====================================================================
+%% EQC Properties
+%% ====================================================================
+
+prop_extract_bucket_from_host() ->
+    ?FORALL({Bucket, BaseHost},{riak_cs_gen:bucket_or_blank(), base_host()},
+            begin
+                BucketStr = binary_to_list(Bucket),
+                Host = compose_host(BucketStr, BaseHost),
+                BaseHostIdx = string:str(Host, BaseHost),
```
slfritchie: Hrm, another `string:str/2` leaked out ... how about commit 5d64db7 to rearrange things?

slfritchie:
I changed a bit of the testing code but didn't do anything about the lager call.  If the test changes are OK, and if the lager call should be removed, then +1, thanks!

kellymclaughlin:
For some reason I thought using the root host in the bucket names was verboten, but evidently it is not. Thanks for the updates. I've removed the lager output and going to rebase to squash that commit and then merge.

PR #369, opened 2012-12-21
Multipart upload (rebase1)

slfritchie:
This is the user/protocol-facing part of the implementation.  Items missing:
- riak_test automation
- Strong consistency server-side work
- Finish multi-cluster server-side work


slfritchie:
Comments for reviewers while I'm out for the rest of this week (and Monday 07 Jan) ... though I will be checking email at least once/day, so I'll be able to answer questions.  If it'd be easier/quicker to chat via Mumble or Skype, I can probably make some time to do that on Friday 04 Jan or Mon 07 Jan.

I wanted to make the first implementation independent of WebMachine, because doing the WM wiring/glue/plumbing is damn annoying to my little peabrain.  So all work started using functions in `riak_cs_mp_utils`.  The functions with `test_` prefix aren't really EUnit tests, but I did use them (mostly interactively at the shell) constantly during development.  After a compile and reload cycle, I'd do something like this to run one or more tests:

```
{ok = riak_cs_mp_utils:test_0(), ok = riak_cs_mp_utils:test_1(), ok = riak_cs_mp_utils:test_comb_parts()}.
```

In a mostly-TDD-like manner, the `riak-cs_mp_utils:test_0()` test came first.  Then `test_1()`, and the `test_comb_parts()` later when I wanted some extra assurance that the code that figured out which MP parts were and weren't included in the final object when an MP upload was finished.

I did put some examples in the commit messages, also.  Here is an update of the examples from the first commit of this branch, https://github.com/basho/riak_cs/commit/b5133fa16dd9da62a541f79b1058f5c938fc8579 ... Run the `riak_cs_mp_utils:test_0()` function to create (via brute-force) the S3 user & bucket that this snippet is based upon.

```
> ok = riak_cs_mp_utils:test_0().
ok

> riak_cs_mp_utils:initiate_multipart_upload(<<"test">>, <<"mp0">>, <<"text/plain">>, {"foobar", "18983ba0e16e18a2b103ca16b84fad93d12a2fbed1c88048931fb91b0b844ad3", "J2IP6WGUQ_FNGIAN9AFI"}, []).
{ok,<<135,161,89,47,120,49,76,168,184,224,77,15,2,240,18,
      209>>}

> rr("../../include/riak_cs.hrl").
> riak_cs_mp_utils:list_multipart_uploads(<<"test">>, {"foobar", "18983ba0e16e18a2b103ca16b84fad93d12a2fbed1c88048931fb91b0b844ad3", "J2IP6WGUQ_FNGIAN9AFI"}, []).                                
{ok,{[#multipart_descr_v1{key = <<"mp0">>,
                          upload_id = <<128,28,9,15,181,46,79,64,170,130,30,6,
                                        15,151,191,68>>,
                          owner_display = "foobar",
                          owner_key_id = "J2IP6WGUQ_FNGIAN9AFI",
                          storage_class = regular,
                          initiated = "2013-01-02T23:39:03.000Z"}],
     []}}
```

Over on the Riak console, I would use this to delete everything related to GC _and_ to the <<"test">> bucket, whenever I wanted to clear the state of whatever my test_\* functions may have jumbled.

```
> {ok, C} = riak:local_client().
> C:delete(B, K) || B <- [<<48,111,58,9,143,107,205,70,33,211,115,202,222,78,131,38,39,180,246>>, <<48,98,58,9,143,107,205,70,33,211,115,202,222,78,131,38,39,180,246>>, <<"riak-cs-gc">>], K <- element(2, C:list_keys(B))].
```

slfritchie:
The get, put, and delete FSMs are mostly unchanged.  The biggest changes:
- get: changing the block naming scheme from `UUID:binary()` to `{UUID:binary(), ChunkSequenceNumber:integer()}`, so that the FSM could tell the difference between chunk seq number 1 of UUID X vs. seq number 1 of UUID Y.  Then the prefetching got a bit weird with the previous chunk ordering code, so I replaced it with a `queue()`.  That allowed me to collapse some nearly-identical code in `waiting_chunks()` into a common func.
- put: We don't want the FSM itself to create a new manifest record when it's done put'ing an MP part upload.  We just want to reuse most of the machinery of the put FSM.  The final gluing together of the parts into a single object is outside of the scope of the put FSM.  If `mani_pid` is `undefined`, then we're doing an MP part upload, so then the  `maybe_*` wrapper functions end up a no-ops.
- delete: very few changes, again due to the block naming scheme change from `binary()` to `{binary(), integer()}`.

slfritchie:
GC-wise, most of the changes were fixes for Dialyzer complaints.  `riak_cs_gc_d:gc_active_manifests()` was the only function that had substantial changes, and that was only a handful of lines.

slfritchie:
`git diff b5133fa16dd9d..HEAD src/*manif*` shows that there was some substantial manifest resolution-related stuff, but it was mostly straightforward, I hope.

The resolution code needed to deal with the multipart-related stuff in the `?MANIFEST.props` list.  There are two main things in that prop list that resolution cares about:
- `multipart`, which contains a `?MULTIPART_MANIFEST` record if the object is a multipart thingie.  The `?MANIFEST.state` part of the main manifest record remains the same: `writing`, `active`, `pending_delete`, etc.
- `multipart_clean`: If this property is present, then all of the processing of which parts (if any) were not included in the final object have been scheduled for immediate deleting.  (Well, the GC polling delay factor still applies, but there's no need for the leeway delay when deleting unused parts.)

In an ideal world, the presence of `multipart_clean` in `?MANIFEST.props` is a hack to void creating a new manifest state, e.g., `writing` -> `active` -> `active2` (new state) -> `pending_delete` -> etcetc.

slfritchie:
The `riak_cs_wm_common` module had a few more common WebMachine callbacks added.

I can't claim that all of the WM stuff that I added in modules like `riak_cs_wm_bucket_uploads.erl` are sweetly elegant.  Feel free to pick apart all of my WM hackery with all the savagery that it deserves.  My defense is that using `s3cmd` and `s3curl` both work.  :-)

A lot of the other WM-related modules were touched, but those edits were almost/all? to clean up Dialyzer warnings.

slfritchie:
For `riak_cs.hrl`, I moved some `-define()` items for current record versions to a place higher in the file to avoid Dialyzer grumpiness.

Not much needed to be added to `#key_context`, yay.

slfritchie:
Commentary on `riak_cs_mp_utils.erl`....

All of the places where a `?MANIFEST` record is written to Riak needed to have the 2I stuff added.  `riak_cs_mp_utils:calc_multipart_2i_dict/3` does the work of figuring out what 2I items need to be created.  And then code in both the GC daemon and the manifest FSM needed calls added to modify the Riak object that they wanted to write to add 2I items to the object.

I use some `try` statements with some very specific `catch error:{badmatch, {Special, _}}` where the `Special` thing is an atom that helps figure out what kind of error to return.  I'm not certain how common that technique is elsewhere in Riak land, but it seems useful and mostly concise?

Hrm, `clean_multipart_unused_parts/2` might deadlock/pause for a helluvalong time because it tries to allocate another Riak client PID?  I might have to add the same `nopid` wrapper trick to allow the caller to have the option of passing in a pre-allocated Riak client pid.

Speaking of passing in Riak client pids, if there's a version of an exported function that doesn't include a Riak client PID, then the atom `nopid` is used in its place ... if the caller passes in a valid Riak client PID, we do not want to accidentally close that client when we're done.  `wrap_riak_connection()` and `wrap_close_riak_connection()` are used for figuring out if/what to open & close, along with the `?PID()` macro which helps shorten calls to `get_riakc_pid()`.

The code in `make_content_types_accepted()` appeared often enough in my WM callback modules that I moved it here.

`new_manifest()` creates a `?MANIFEST` record with all of the proper multipart goodness in the `props` list.  The `?MANIFEST.content_length` will be incremented whenever a MP part is added to the object: that `content_length` is supposed to help inform usage stats in the existing usage calculations, so that the storage space of parts is not ignored.

`upload_part_1blob/2` isn't really needed for the S3 API or for the WM glue, but it was tremendously useful as a scaffolding/helper function for the `test_0()` and `test_1()` functions.

The filtering code in `list_multipart_uploads2()` is ugly.  IIRC filtering very similar to that is also necessary for full API support for the S3 "list objects" operation?

Misc testing function stuff:
- `test_create_users()` creates a couple of users via raw Riak `put` ops ... and will need changing if any of the underlying Erlang data types change.  Also, their ACL/user ID stuff is decoupled from the functions `test_user1()`, `test_user1_secret()`, et al.
- The test funcs always use the S3 bucket "test" and object name "mp0".
- An exercise similar to the Erlang code in `test_0()` and `test_1()` ought to be recreated using riak_test?

src/riak_cs_acl.erl
```diff
@@ -52,6 +52,8 @@ anonymous_bucket_access(Bucket, RequestedAccess, RiakPid) ->
         {error, Reason} ->
             %% @TODO Think about bubbling this error up and providing
             %% feedback to requester.
+            %% lager:error("Anonymous bucket access hacked, ignoring ~p", [Reason]),
```
kellymclaughlin: Are these two lines intentionally here or just some debug cruft that should be removed?

src/riak_cs_gc.erl
```diff
@@ -29,13 +30,23 @@
...
+            case riak_cs_mp_utils:clean_multipart_unused_parts(M) of
+                same ->
+                    ActiveUUIDs = [M?MANIFEST.uuid],
+                    GCManiResponse = gc_specific_manifests(ActiveUUIDs,
+                                                           RiakObject,
+                                                           RiakcPid),
+                    return_active_uuids_from_gc_response(GCManiResponse,
+                                                         ActiveUUIDs);
+                updated ->
+                    io:format("Updated!\n"),
```
kellymclaughlin: Debug output

src/riak_cs_mp_utils.erl
```diff
@@ -0,0 +1,893 @@
...
+
+-ifdef(TEST).
+-compile(export_all).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
+-define(PID(WrappedPid), get_riakc_pid(WrappedPid)).
+
+%% export Public API
+-compile(export_all).                           % SLF DEBUGGING ONLY!
```
kellymclaughlin: Just noting a reminder to remove this prior to merge.

kellymclaughlin:
Verified all unit, integration, and client tests pass

kellymclaughlin:
All current riak_test modules pass also. Next step is to try and provide an answer to this question:

> An exercise similar to the Erlang code in test_0() and test_1() ought to be recreated using riak_test?

kellymclaughlin:
I rebased yet again from master to pull in the latest round of `riak_test` scaffolding changes. Branch is [gh300-multipart-upload-rebase2](https://github.com/basho/riak_cs/tree/gh300-multipart-upload-rebase2). Also added my `riak_test` module in for inspection. It can still be expanded with more test cases, but it's a good start I think. Also running the test requires using [this](https://github.com/basho/erlcloud/tree/klm-mp-uploads) branch of erlcloud until the changes are merged to `master`.

slfritchie:
Kelly, line 102 of riak_cs_wm_stats.erl is broken, and I'm not sure how to fix it.  As the "git annotate" alleged author of that line, I'm hoping that you have an idea?  Dialyzer caught this:

```
riak_cs_wm_stats.erl:102: Call to missing or unexported function riak_cs_wm_utils:validate_auth_header/3
```

Otherwise ... I've addressed (I hope) your comments above and also fixed some additional dialyzer warnings that have popped up.  Those fixes are in commit 56ab09da9 on the `gh300-multipart-upload-rebase2` branch ... we might as well finish review there?

kellymclaughlin:
`validate_auth_header/3` became `validate_auth_header/4`. It needs `Ctx` as the 4th paramter.

I'll take a look at the other changes on the `rebase2` branch and we can continue to work from there.

src/riak_cs_mp_utils.erl
```diff
@@ -0,0 +1,893 @@
...
+            same;
+        MpM ->
+            case {proplists:get_value(multipart_clean, Props, false),
+                  MpM?MULTIPART_MANIFEST.cleanup_parts} of
+                {false, []} ->
+                    same;
+                {false, PartsToDelete} ->
+                    try
+                        {m_cmup1, {ok, RiakcPid}} =
+                            {m_cmup1, riak_cs_utils:riak_connection()},
```
kellymclaughlin: Why not just have the riak pid passed as a parameter to this function? That would avoid the overhead of having to establish another connection and simplify this function by removing this `try/catch` block. Another good reason is avoiding the connection pool deadlocking issues we had once upon a time.

slfritchie: Hrm, looks like `clean_multipart_unused_parts/1` was overlooked in my round of adding a wrapped Riak client pid through all the major API functions.  Good catch.

src/riak_cs_get_fsm.erl
```diff
@@ -412,40 +323,36 @@ prepare(#state{bucket=Bucket,
...
         {ok, Manifest} ->
             _ = lager:debug("Manifest: ~p", [Manifest]),
-            State#state{manifest=Manifest,
-                        mani_fsm_pid=ManiPid};
+            case riak_cs_mp_utils:clean_multipart_unused_parts(Manifest) of
+                same ->
+                    State#state{manifest=Manifest,
+                                mani_fsm_pid=ManiPid};
+                updated ->
+                    riak_cs_manifest_fsm:stop(ManiPid),
```
kellymclaughlin, line 324: Would it be reasonable to have `clean_multipart_unused_parts` return an updated manifest so we can carry on without having to shutdown the fsm and call `prepare` again. It would also eliminate another roundtrip to Riak.

src/riak_cs_get_fsm.erl
```diff
@@ -39,30 +39,29 @@
...
                 bucket :: term(),
                 caller :: reference(),
                 key :: term(),
-                block_buffer=[] :: [{pos_integer(), term()}],
+                got_blocks=orddict:new() :: orddict:orddict(),
                 manifest :: term(),
-                manifest_uuid :: term(),
-                blocks_left :: set(),
+                blocks_order :: [block_name()],
+                blocks_intransit=queue:new() :: queue(),
```
kellymclaughlin, line 59: Can this just be a list instead of a queue?

src/riak_cs_mp_utils.erl
```diff
@@ -0,0 +1,893 @@
...
+    {?MD_INDEX, lists:usort(lists:append(L_2i))}.
+
+abort_multipart_upload(Bucket, Key, UploadId, Caller) ->
+    abort_multipart_upload(Bucket, Key, UploadId, Caller, nopid).
+
+abort_multipart_upload(Bucket, Key, UploadId, Caller, RiakcPidUnW) ->
+    do_part_common(abort, Bucket, Key, UploadId, Caller, [], RiakcPidUnW).
+
+clean_multipart_unused_parts(?MANIFEST{bkey=BKey, props=Props} = Manifest) ->
+    case proplists:get_value(multipart, Props, false) of
```
kellymclaughlin, line 74: When `riak_cs_put_fsm` call `riak_cs_lfs_utils:new_manifest` it sets the `props` field of the new manifest to `undefined` (see [here](https://github.com/basho/riak_cs/blob/master/src/riak_cs_put_fsm.erl#L352)). Because of this fact we will have to account for `props` being undefined here so existing manifests cause this line to barf. We should change `riak_cs_put_fsm` to make `props` an empty list instead also.

src/riak_cs_lfs_utils.erl
```diff
@@ -92,9 +92,30 @@ initial_blocks(ContentLength, BlockSize) ->
...
+initial_blocks(ContentLength, SafeBlockSize, UUID) ->
+    Bs = initial_blocks(ContentLength, SafeBlockSize),
+    [{UUID, B} || B <- Bs].
+
+block_sequences_for_manifest(?MANIFEST{uuid=UUID,
+                                       content_length=ContentLength,
+                                       props=Props}=Manifest) ->
     SafeBlockSize = safe_block_size_from_manifest(Manifest),
-    initial_blocks(ContentLength, SafeBlockSize).
+    case proplists:get_value(multipart, Props) of
```
kellymclaughlin, line 105: Same deal here with `props` field being `undefined`. Added this function clause to fix:

```
block_sequences_for_manifest(?MANIFEST{props=undefined}=Manifest) ->    
    block_sequences_for_manifest(Manifest?MANIFEST{props=[]});
```

src/riak_cs_get_fsm.erl
```diff
@@ -213,136 +206,60 @@ waiting_continue_or_stop(Event, From, State) ->
...
+                    %% Must use gen_fsm:reply/2 here!  We are shared
+                    %% with an async event func and must return next_state.
+                    gen_fsm:reply(From, {chunk, Block}),
+                    {sent, State#state{got_blocks=orddict:erase(NextBlock, Got),
+                                       blocks_intransit=UpdIntransit}};
+                error ->
+                    {not_sent, State#state{from=From}}
+            end
+    end.
+%% waiting_chunks(get_next_chunk, From, #state{from=PreviousFrom}=State) when PreviousFrom =/= undefined ->
```
kellymclaughlin: This can be deleted?

src/riak_cs_get_fsm.erl
```diff
@@ -39,30 +39,29 @@
...
                 key :: term(),
-                block_buffer=[] :: [{pos_integer(), term()}],
+                got_blocks=orddict:new() :: orddict:orddict(),
                 manifest :: term(),
-                manifest_uuid :: term(),
-                blocks_left :: set(),
+                blocks_order :: [block_name()],
+                blocks_intransit=queue:new() :: queue(),
                 test=false :: boolean(),
-                next_block=0 :: 0 | pos_integer(),
```
kellymclaughlin, line 61: Getting rid of this field and only using a queue for returning blocks restricts us to only fetching a single block at a time since it relies on the ordering of the block server responses to ensure the blocks are returned in the proper order. Using a `fetch_concurrency` of > 1 works with this branch, but there is a race condition where the blocks can returned out of order. I think it would be worth having the ability to set that to > 1 and be assured the blocks would be returned in order.

src/riak_cs_web.erl
```diff
@@ -51,10 +51,13 @@ base_resources() ->
      {["buckets", bucket, "acl"], riak_cs_wm_common, props(riak_cs_wm_bucket_acl)},
      {["buckets", bucket, "location"], riak_cs_wm_common, props(riak_cs_wm_bucket_location)},
      {["buckets", bucket, "versioning"], riak_cs_wm_common, props(riak_cs_wm_bucket_versioning)},
+     {["buckets", bucket, "uploads"], riak_cs_wm_common, props(riak_cs_wm_bucket_uploads)},
      %% TODO: bucket policy
      %% %% Object resources
      {["buckets", bucket, "objects", object], riak_cs_wm_common, props(riak_cs_wm_object)},
-     {["buckets", bucket, "objects", object, "acl"], riak_cs_wm_common, props(riak_cs_wm_object_acl)}
+     {["buckets", bucket, "objects", object, "acl"], riak_cs_wm_common, props(riak_cs_wm_object_acl)},
+     {["buckets", bucket, "objects", object, "uploads", uploadId], riak_cs_wm_common, props(riak_cs_wm_objects_uploads2)},
```
kellymclaughlin: `riak_cs_wm_objects_uploads` and `riak_cs_wm_objects_uploads2` as module names do not really give any clues as to the distinction. Since the upload only pertains to a single object I would suggest using the singular `object` versus `objects`. How about using 'riak_cs_wm_object_upload' and `riak_cs_wm_object_upload_part` respectively? That at least gives a little more obvious distinction between the two even if it is minimal.

slfritchie:
Hooray, the riak_cs #376 issues appears to address my concern about the `riak_cs_wm_stats:validate_auth_header/3` vs. '/4' arity change.

src/riak_cs_mp_utils.erl
slfritchie: Yay, this one seems easier ... the two callers are `riak_cs_gc.erl` and `riak_cs_get_fsm.erl`.  Both are cases where we notice that a multipart manifest hasn't yet made the pseudo-state transition from active to active + `multipart_clean` property in the manifest's props list.  In both cases, the caller already has a Riak client pid, so I'll pass that one along as a new arg.

I don't see a need to have a allocate-your-own-riak-client-pid version, like most of the other major API functions in `riak_cs_mp_utils.erl` do.

src/riak_cs_get_fsm.erl
slfritchie, line 324: I was being conservative with the implementation ... getting caught in the active-but-not-yet-`multipart_clean`-property state should be quite rare: the client would have to omit a part on commit or overwrite a part (usually doesn't happen), and then there'd have to be a CS or Riak bug/failure to interfere with prepping the omitted/overwritten part(s) for GC.  I guess I'd rather return to Riak in this case to allow opportunities to witness other updates of the manifest(s) and merge 'em.

src/riak_cs_get_fsm.erl
slfritchie, line 59: Yes, I suppose ... but the simplicity of the current `queue` use fell out of an original, more complex implementation that benefitted from using `queue`.  Now, it's simple ... but even in this current use, I think the O(1) amortized cost of using `queue` is worthwhile.

src/riak_cs_mp_utils.erl
slfritchie, line 74: Hrm, in commit 845f2810 at line 378, that call to `riak_cs_lfs_utils:new_manifest/11` is using `[]` for the props already.

src/riak_cs_lfs_utils.erl
slfritchie, line 105: Hrm, with `new_manifest()` being called with the /9 arity or with /11 arity + [] for props, the clause above shouldn't be necessary?  Also, I'm not seeing Dialyzer making any complaints.

src/riak_cs_get_fsm.erl
slfritchie: Yup, deleted.

src/riak_cs_mp_utils.erl
kellymclaughlin, line 74: Yeah, but I was testing fetching a file that was stored with the `master` version so for backward-compatibility reasons we'll have to check for it.

src/riak_cs_lfs_utils.erl
kellymclaughlin, line 105: This is another case where I ran into troubles with files stored with _older_ versions of the code.

src/riak_cs_get_fsm.erl
slfritchie, line 61: Hrm, I thought I'd run an experiment where I had the block servers add a random sleep before replying, and the S3 client never witnessed any out-of-order-reassembly.

slfritchie, line 61: I'm pretty sure that `perhaps_send_to_user()` is taking care of replies that arrive out of order.

https://github.com/basho/riak_cs/blob/gh300-multipart-upload-rebase2/src/riak_cs_get_fsm.erl#L218

The `blocks_intransit` queue contains the order that replies must be sent back to the client.  As chunks arrive, they're added to the unordered `got_blocks` orddict.

src/riak_cs_web.erl
slfritchie: Renamed as suggested.

slfritchie:
Kelly, are there any other bits that need to be addressed in review?

kellymclaughlin:
No, nothing outstanding to be addressed. The one last thing I want to do today is verify the functioning of the various options for listing multipart uploads.

kellymclaughlin:
Ok I lied, one thing: https://github.com/basho/riak_cs/commit/d0e30589b38c16106a842dc8db1b2f353a2484ad#commitcomment-2521509

src/riak_cs_wm_bucket_uploads.erl
```diff
@@ -0,0 +1,152 @@
...
+            %% appease webmachine
+            {[{"text/plain", unused_callback}], RD, Ctx}
+    end.
+
+-spec content_types_accepted(#wm_reqdata{}, #context{}) -> {[{string(), atom()}], #wm_reqdata{}, #context{}}.
+content_types_accepted(RD, Ctx) ->
+    riak_cs_mp_utils:make_content_types_accepted(RD, Ctx).
+
+make_list_mp_uploads_opts(RD) ->
+    %% Weird: wrq:req_qs(RD) returns [], so that any call to
```
kellymclaughlin: _Danger Will Robinson, failure of test coverage!_ This is a bug in the s3 rewriting module. I have made a pull request to address the issue and also expand the test coverage from a branch based off the `rebase2` version of this branch. If it looks ok I would say merge my pr branch into the `rebase2` branch and let's change this to use `wrq:req_qs`. [Here](https://github.com/basho/riak_cs/pull/387) is the pr.

src/riak_cs_web.erl
kellymclaughlin: I still see `riak_cs_wm_objects_uploads2` in this module on the `rebase2` branch.

src/riak_cs_manifest_fsm.erl
```diff
@@ -307,14 +322,34 @@ update_from_previous_read(RiakcPid, RiakObject,
...
+    riak_cs_utils:put(RiakcPid, NewRiakObject).
+
+update_md_with_multipart_2i(RiakObject, WrappedManifests, Bucket, Key) ->
+    MD0 = try
+              %% During testing, it's handy to delete Riak keys in the
+              %% S3 bucket, e.g., cleaning up from a previous test.
+              %% Let's not trip over tombstones here.
+              case ([MD || {MD, V} <- riakc_obj:get_contents(RiakObject),
+                           V /= <<>>]) of
+                  [X] -> X;
```
kellymclaughlin: I found an issue when running the updates to the `riak_test` module where there are metadata siblings (_i.e._ `MD` does not match any of the case clauses here). This happens when I initiate an upload for a key, abort the upload prior to uploading any parts, then initiate another upload. If I take the naive route of adding a clause to just use the first of the metadata siblings the second initiation succeeds.

src/riak_cs_manifest_fsm.erl
```diff
@@ -307,14 +322,34 @@ update_from_previous_read(RiakcPid, RiakObject,
...
+                  []  -> throw(no_metadata)
+              end
+          catch throw:no_metadata ->
+                  dict:new()
+          end,
+    {K_i, V_i} = riak_cs_mp_utils:calc_multipart_2i_dict(
+                   [M || {_, M} <- WrappedManifests], Bucket, Key),
+    MD = dict:store(K_i, V_i, MD0),
```
kellymclaughlin, line 356: Won't this overwrite any existing `index` metadata that may exist? Should we pass in any existing index info and append to it?

kellymclaughlin:
I find I'm having trouble tracking what has been fixed and what has not so I'm going to take the radical step of renaming the `rebase1` branch to `rebase1-a` and renaming `rebase2` branch to `rebase1` and doing a force push so that the pull request will pick it up and I can map changes to comments.

src/riak_cs_wm_bucket_uploads.erl
slfritchie: I've merged that PR (commit 55416a0), will push in a moment.  Everything seems well enough with ad hoc s3cmd and s3curl tests.  I'm still not functioning yet in riak_test land, so I don't know what problems that it might see.

src/riak_cs_manifest_fsm.erl
slfritchie, line 356: I think that this is kosher, but having a Riak 2I experienced eye take a look at it would be good.  AFAIK, we're calculating 100% of all the 2I stuff that we need from all of the `WrappedManifests`, there isn't anything else to preserve?

slfritchie:
Here's a PR for another repo that needs some quick review & approval: https://github.com/basho/riak-erlang-client/pull/83

src/riak_cs_gc.erl
```diff
@@ -159,24 +169,30 @@ mark_as_scheduled_delete(UUIDsToMark, RiakObject, RiakcPid) ->
...
--spec mark_manifests(riakc_obj:riak_object(), [binary()], fun(), pid()) ->
-                    {ok, riakc_obj:riak_object()} | {error, term()}.
+-spec mark_manifests(riakc_obj:riakc_obj(), [binary()], fun(), pid()) ->
+                    {ok, riakc_obj:riakc_obj()} | {error, term()}.
 mark_manifests(RiakObject, UUIDsToMark, ManiFunction, RiakcPid) ->
     Manifests = riak_cs_utils:manifests_from_riak_object(RiakObject),
     Marked = ManiFunction(Manifests, UUIDsToMark),
-    UpdObj = riakc_obj:update_value(RiakObject, term_to_binary(Marked)),
+    UpdObj0 = riakc_obj:update_value(RiakObject, term_to_binary(Marked)),
+    UpdObj = riak_cs_manifest_fsm:update_md_with_multipart_2i(
```
kellymclaughlin, line 184: This is the cause of the problem when trying to do an `initiate/abort/initiate` cycle. It writes the index entries with the hashed bucket name where every other call to `update_md_with_multipart_2i` uses the plaintext bucket name. I do not see why we need to update the index information again here at all. If it is a multipart manifest then the information is already present and if not, we should not add it. Am I missing something or can we just omit this step?

src/riak_cs_mp_utils.erl
```diff
@@ -0,0 +1,901 @@
...
+                    undefined,
+                    [],
+                    P_BlockSize,
+                    no_acl_yet)} ||
+                    ?PART_MANIFEST{part_id=P_UUID,
+                                   content_length=ContentLength,
+                                   block_size=P_BlockSize} <- PartsToDelete],
+    ok = riak_cs_gc:move_manifests_to_gc_bucket(PartDelMs, RiakcPid, false).
+
+%% The intent of the wrap_* functions is to make this module's code
```
kellymclaughlin, line 693: Is it useful to have 2 ways? Why not just force the Riak pid to be passed in and simplify the code by not having to wrap?

kellymclaughlin:
I was having trouble with the prefix and delimiter options when listing uploads, but [these](https://gist.github.com/da8ce60e0cd008107fad) patches along with the test module updates of the last commit have it passing for me now. I borrowed much of the code in this patch from the work I did to implement the similar option support for the new bucket listing work.

src/riak_cs_mp_utils.erl
slfritchie, line 693: It was really handy to have the Riak client pid allocation hidden inside the API when I was doing my initial development & testing.  I think of it as having an easy/lazy option similar to using s3curl for doing individual multipart operations rather than s3cmd doing it all form me.  {shrug}

src/riak_cs_gc.erl
slfritchie, line 184: Fixed by commit 2454371: threading through Bucket & Key args for `update_md_with_multipart_2i()` sanity.

slfritchie:
TODO item: Kelly or anyone else, could you please quickly review the riak-erlang-client diffs in this PR: https://github.com/basho/riak-erlang-client/pull/83 ... it helps clean up some Dialyzer warnings.

slfritchie:
TODO item: Here's another Dialyzer-related set changes for quick review, a PR for velvet: https://github.com/basho/velvet/pull/3

slfritchie:
TODO item: Here's another Dialyzer-related set of changes for quick review, a PR for webmachien: https://github.com/basho/webmachine/pull/119

kellymclaughlin:
+1'd those 3 dialyzer prs

kellymclaughlin:
The `riak_test` is still failing due to the prefix and delimiter changes I mentioned a few days ago.

src/riak_cs_wm_object_upload.erl
```diff
@@ -0,0 +1,212 @@
...
+
+%% @doc Final step of {@link forbidden/2}: Authentication succeeded,
+%% now perform ACL check to verify access permission.
+-spec check_permission(atom(), #wm_reqdata{}, #context{}, lfs_manifest() | notfound) ->
+                              {boolean() | {halt, non_neg_integer()}, #wm_reqdata{}, #context{}}.
+check_permission('GET', RD, Ctx, notfound) ->
+    {{halt, 404}, riak_cs_access_logger:set_user(Ctx#context.user, RD), Ctx};
+check_permission('HEAD', RD, Ctx, notfound) ->
+    {{halt, 404}, riak_cs_access_logger:set_user(Ctx#context.user, RD), Ctx};
+check_permission(_, RD, Ctx=#context{requested_perm=RequestedAccess,local_context=LocalCtx}, Mfst) ->
```
kellymclaughlin, line 60: I think we should change all of the mp upload resources to use kota's nice `bucket_access_authorize_helper` and `object_access_authorize_helper` functions in `riak_cs_wm_utils`, but this change is so large maybe it would be better to make that a separate pull request once this is merged. Thoughts?

src/riak_cs_mp_utils.erl
```diff
@@ -0,0 +1,902 @@
...
+
+list_multipart_uploads(Bucket, Caller, Opts) ->
+    list_multipart_uploads(Bucket, Caller, Opts, nopid).
+
+list_multipart_uploads(Bucket, {_Display, _Canon, CallerKeyId} = Caller,
+                       Opts, RiakcPidUnW) ->
+    case wrap_riak_connection(RiakcPidUnW) of
+        {ok, RiakcPid} ->
+            try
+                BucketOwnerP = is_caller_bucket_owner(?PID(RiakcPid),
```
kellymclaughlin, line 173: I still think this access checking is not needed here and if it is then we should fix the normal access checking so that it is enforced before we reach this point, but maybe this is another thing that makes sense to do as a separate pr once this large and long-running pr has been merged?

src/riak_cs_wm_object_upload.erl
slfritchie, line 60: I would be in favor of a separate task/kanban card/PR for such a refactoring.  This review has already taken quite a while. :-)

src/riak_cs_mp_utils.erl
slfritchie, line 173: +1 to waiting ... it could very well be that this is not the only redundant case of ACL checking.  This is a holdover to implementing the features that the HTTP API needed before doing the WebMachine stuff for the HTTP bits.

Having said that, there may be value in keeping the ACL checking in place: if we want to use these functions outside of WebMachine, then their enforcement function would be valuable.  Refactor to skip the enforcement check in the case that the caller (e.g. WebMachine resource) has done the ACL checking work?

kellymclaughlin:
`prefix` and `delimiter` for listing uploads does not yet work properly, but going to give the go ahead for the merge so we can address the remaining issues as separate (and smaller) pieces of work. 

+1 to merge

PR #376, opened 2013-01-22
missing riak_cs_wm_utils:validate_auth_hader/4 when /stats called

kuenishi:
got 500, looked like a bug.


src/riak_cs_wm_stats.erl
```diff
@@ -109,6 +109,7 @@ forbidden(RD, #ctx{auth_bypass = AuthBypass, riakc_pid = RiakPid} = Ctx) ->
                                                 [], [<<"false">>, Admin]),
                     {false, RD, Ctx};
                 _ ->
+                    %% admin account is not allowd -> 403
```
kellymclaughlin: Minor comment typo: allowed

kellymclaughlin: Comment should actually be _non-admin_ is not allowed

kellymclaughlin:
A couple of things to address in the added comment, but +1 to merge after that is fixed.

PR #379, opened 2013-01-24
Keylist then mapred squashed1 rebased1

reiddraper:
# List Objects Overhaul

This PR makes several major changes to the way object listings are done in
Riak CS. First, let's review how S3 expects object listings to work:
## How it should work...

Keys (and metadata) are returned, sorted, in batches of `max_keys`.
Results start at key `marker`, exclusive. Two other parameters, `prefix`
and `delimiter` are described
[here](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html).
Since the `max_keys` maximum is 1000, users that want to list all of the
keys in a bucket with more than 1000 keys will make multiple
requests, and use the `marker` parameter to paginate through the results.
## How it used to work
- `max_keys` was ignored, we'd return _all_ the keys in the bucket
  every time
- Keylistings for buckets over ~ ten's of thousands of keys would
  be taxing on the cluster, and/or would time out entirely.
- The basic process was to use keylisting w/ key filters as the input
  to a map-reduce job. The biggest issue was that instead of doing this
  in batches of 1000, we'd do the entire bucket at once.
## How it works with this PR

This pull-request, among other things, now uses an FSM to return
the results in batches of `max_keys`. A visual of the FSM is shown below

FSM diagram

The `dot` code for the FSM is:

```
digraph graphname {
     prepare -> waiting_list_keys [label="list all keys in bucket"];
     waiting_list_keys -> waiting_map_reduce [label="first map reduce"];
     waiting_map_reduce -> waiting_map_reduce [label="need more results"];
     waiting_map_reduce -> done [label="have enough results"]
 }
```

The FSM will still list all the keys in the bucket (we don't have much
choice here yet, until we have paginated 2i `$key` or list keys queries.
However, once the key-listing results are in, they're sent to the Riak CS
process, instead of directly used as inputs to a map/reduce job. Then,
they're sorted, and some subset of them are used to make a map/reduce
request. The subset is `max_keys` \* 1.5. (TODO: this should be configurable).
We do a map/reduce request on _more_ than just `max_keys` because some of those
keys might be tombstoned or the manifests marked as `pending_delete`.

As the results from the map/reduce request finish, we determine if we have
enough results to return to the user, or we make another map/reduce
request starting at a new range of keys. Repeat.
### ETS keylist caching

For a 1MM key bucket, with batches of `max_keys=1000`, this means listing
all 1MM keys 1k times. In order to reduce this, this pull-request
also caches keylistings from Riak in an ETS table. This way, subsequent
paging requests will only have to list all the keys in the table
if the cache is not filled. The cache is currently only valid for
a bucket-name + HTTP connection. If the http connection is closed to the
end user, the cache will be cleared for that entry. This is for
several reasons. One is that the keylistings can become out-of-date, and
so we view the cache as 'valid for a single http sessions'. The cache
can also expire based on a TTL, which is controlled in the CACHE_TIMEOUT
value in `list_objects.hrl` (TODO: each of these options should probably
be app.config able). There are also options for the minimum number
of keys to cache, and the total size of the ETS table. If the table is full,
based on bytes, attempts to write to the table will fail.
## Testing

This PR should pass all unit, integration, client, and riak_test tests.
The end-user clients don't seem to mind our change from returning all keys
to now returning in batches of `max_keys`, after all, that's how S3 works.

h/t to @kellymclaughlin for also putting in a bunch of work on this.
## TODO
- [x] make ETS cache easier to bypass (app.config that is effective immediately
- [x] try to read the constants from `list_objects.hrl` from app.config first
- [x] easy way to get the current byte-size of the ETS cache (observability)


src/riak_cs_list_objects_fsm.erl
```diff
@@ -0,0 +1,623 @@
...
+filtered_keys_from_request(?LOREQ{marker=Marker}, KeyList, KeyListLength) ->
+    MarkerIndex = index_of_first_greater_element(KeyList, Marker),
+    lists:sublist(KeyList, MarkerIndex, KeyListLength).
+
+%% only works for positive numbers
+-spec round_up(float()) -> integer().
+round_up(X) ->
+    erlang:round(X + 0.5).
+
+%% @doc Return the index (1-based) where
```
kellymclaughlin, line 718: Maybe these comments should include a mention that the expectation is that `List` is sorted.

---

addressed in 76a0163a9d3d539aab22fbc05b04f69d6b867bd6
-- reid

reiddraper, line 718: Good idea, will push that up tomorrow.

reiddraper:
ps. the commit messages themselves should provide some more good information on understanding this PR as well.

src/riak_cs_list_objects_fsm.erl
```diff
@@ -0,0 +1,624 @@
...
+                         non_neg_integer(),
+                         float()) ->
+    {integer(), integer()}.
+next_mr_query_spec([], TotalNeeded, _NumHaveSoFar, KeyMultiplier) ->
+    StartIdx = 1,
+    EndIdx = round_up(TotalNeeded * KeyMultiplier),
+    {StartIdx, EndIdx};
+next_mr_query_spec(PrevRequests, TotalNeeded, NumHaveSoFar, KeyMultiplier) ->
+    MoreNeeded = TotalNeeded - NumHaveSoFar,
+    StartIdx = element(2, lists:last(PrevRequests)) + 1,
```
kellymclaughlin, line 539: The `+ 1` here is throwing off results for me and causing a key I expect to be in the results to be omitted. Is it really needed?

kellymclaughlin, line 539: The last comment may be too vague. I'll put together a more concrete example to illustrate later.

kellymclaughlin, line 539: Here's some data on this problem. I get failure running the modified list objects riak_test on the [`cs378-prefix-and-delimiter`](https://github.com/basho/riak_cs/blob/cs378-prefix-and-delimiter/riak_test/tests/list_objects_test.erl) branch (which is branched from this branch). I added lager output for the start and end indexes calculated in `next_mr_query_spec`. Here is what I saw:

```
2013-01-28 22:36:45.344 [info] <0.1981.0>@riak_cs_list_objects_fsm:next_mr_query_spec:478 Start Index: 1 End Index: 46
2013-01-28 22:36:45.464 [info] <0.1981.0>@riak_cs_list_objects_fsm:next_mr_query_spec:484 Start Index: 47 End Index: 79
2013-01-28 22:36:45.549 [info] <0.1981.0>@riak_cs_list_objects_fsm:next_mr_query_spec:484 Start Index: 80 End Index: 125
```

Since the start index is the previous end index + 1 then you can see from that how a result is potentially skipped each time a new MR is run.

Here's the failure output from the test to illustrate:

```
{{assertEqual_failed,[{module,list_objects_test},
                      {line,118},
                      {expression,"[ proplists : get_value ( key , O ) || O <- proplists : get_value ( contents , ObjList ) ]"},
                      {expected,["1","10","11","12","13","14","15","16","17",
                                 "18","19","2","20","21","22","23","24","25",
                                 "26","27","28","29","3","30","4","5","6",
                                 "7"]},
                      {value,["1","10","11","12","13","14","15","16","17",
                              "18","19","2","20","21","22","23","24","25",
                              "26","27","28","29","3","30","4","6","7","9"]}]},
 [{list_objects_test,'-verify_object_list/4-fun-1-',2,
                     [{file,"riak_test/tests/list_objects_test.erl"},
                      {line,118}]},
  {list_objects_test,confirm,0,
                     [{file,"riak_test/tests/list_objects_test.erl"},
                      {line,77}]}]}
```

The difference between the expected result and the actual result is that the actual returned result is missing the `"5"`. In this case "5" corresponds to index _46_ that was skipped. 

---

addressed in 53fecdd9285193d19f40ae63e383893439e0e57a --reid

jrwest:
i've tried a few clean checkouts to try and rule out my local environment (although still not 100%) but seeing this test failure regularly when running `make test-erlang` or `make`test-client`. The python/clojure/int/riak_test tests all pass for me. 

https://gist.github.com/336a225405577248dddf

---

fixed in f6e59670240f49dffee0aa91372f1ce20d4a230e -- reid

reiddraper:
@jrwest same issue for me. Lemme see what happens on master with that test too.

jrwest:
should have mentioned, all tests (except the two `riak_test`s that fail due to https://github.com/basho/riak_cs/issues/385) pass on master for me.

kuenishi:
I made a PR to fix #385 - sorry!

src/riak_cs_list_objects.erl
```diff
@@ -0,0 +1,116 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2007-2011 Basho Technologies, Inc.  All Rights Reserved.
```
kuenishi, line 3: It's already 2013, isn't it? :)

kuenishi:
I'm afraid prefix matching is not working . 

```
$ ./s3cmd ls s3://test3/setup
2013-01-30 08:45      2707   s3://test3/INSTALL
2013-01-30 08:44        28   s3://test3/setup.cfg
2013-01-30 08:44      2399   s3://test3/setup.py
```

in this case we'll expect 

```
$ ./s3cmd ls s3://test3/setup
2013-01-30 08:44        28   s3://test3/setup.cfg
2013-01-30 08:44      2399   s3://test3/setup.py
```

Oh, this will be done in #378 ? It would be fine!

reiddraper:
@kuenishi 

> Oh, this will be done in #378 ? It would be fine!

yep, h/t to @kellymclaughlin for pulling that together too

reiddraper:
`make test-erlang` now passing as of f6e59670240f49dffee0aa91372f1ce20d4a230e, thanks to @kellymclaughlin for the hand debugging that.

jrwest:
confirmed `test-erlang` is passing for me as well

src/riak_cs_wm_objects.erl
```diff
@@ -81,8 +81,8 @@ to_xml(RD, Ctx=#context{start_time=StartTime,
                         bucket=Bucket,
                         requested_perm='READ',
                         riakc_pid=RiakPid}) ->
-    riak_cs_dtrace:dt_bucket_entry(?MODULE, <<"bucket_list_keys">>, 
-                                      [], [riak_cs_wm_utils:extract_name(User), Bucket]),
+    riak_cs_dtrace:dt_wm_entry(?MODULE, <<"to_xml">>, [], [riak_cs_wm_utils:extract_name(User), Bucket]),
```
jrwest: not sure its an issue but `riak_cs_wm_common` also adds a `<<"to_xml">>` dtrace entry (https://github.com/basho/riak_cs/blob/master/src/riak_cs_wm_common.erl#L199) but without the user and bucket (although we should be able to at least log the user in `wm_common`).

src/riak_cs_wm_objects.erl
```diff
@@ -90,28 +90,51 @@ to_xml(RD, Ctx=#context{start_time=StartTime,
             CodeName = no_such_bucket,
             Res = riak_cs_s3_response:api_error(CodeName, RD, Ctx),
             Code = riak_cs_s3_response:status_code(CodeName),
-            riak_cs_dtrace:dt_bucket_return(?MODULE, <<"bucket_list_keys">>, 
-                                               [Code], [riak_cs_wm_utils:extract_name(User), Bucket]),
+            riak_cs_dtrace:dt_wm_return(?MODULE, <<"to_xml">>, [Code], [riak_cs_wm_utils:extract_name(User), Bucket]),
```
jrwest: same here re: dtrace entry also in `wm_common`

riak_test/tests/list_objects_test.erl
```diff
@@ -0,0 +1,154 @@
...
+    {RiakNodes, _CSNodes, _Stanchion} = rtcs:setup(4),
+
+    FirstNode = hd(RiakNodes),
+
+    {AccessKeyId, SecretAccessKey} = rtcs:create_user(FirstNode, 1),
+
+    %% User config
+    UserConfig = rtcs:config(AccessKeyId, SecretAccessKey, rtcs:cs_port(FirstNode)),
+
+    lager:info("User is valid on the cluster, and has no buckets"),
```
jrwest, line 20: maybe not in this PR but perhaps we should just move this assertion into `rtcs`? it seems to be in almost all of the cs r_ts.

src/riak_cs_list_objects_fsm.erl
```diff
@@ -0,0 +1,785 @@
...
+-spec maybe_write_to_cache(state(), list()) -> ok.
+maybe_write_to_cache(#state{use_cache=false}, _ListofListofKeys) ->
+    ok;
+maybe_write_to_cache(#state{cache_key=CacheKey,
+                            caller_pid=CallerPid}, ListofListofKeys) ->
+    case riak_cs_list_objects_ets_cache:can_write(CacheKey,
+                                                  CallerPid,
+                                                  lists:flatlength(ListofListofKeys)) of
+        true ->
+            lager:debug("writing to the cache"),
```
jrwest: minor nitpick: is this log message necessary since the cache itself logs a `debug` message when the write is submitted?

jrwest:
there are several dialyzer warnings like the one below that should probably be added to ignore warnings:

```
riak_cs_list_objects_ets_cache.erl:123: Expression produces a value of type 'ok' | {'error','lager_not_running'}, but this value is unmatched
```

src/riak_cs_wm_objects.erl
```diff
@@ -90,28 +90,51 @@ to_xml(RD, Ctx=#context{start_time=StartTime,
...
-                                                       [Code], [riak_cs_wm_utils:extract_name(User), Bucket]),
-                    X
+                    Response = riak_cs_s3_response:api_error(Reason, RD, Ctx),
+                    riak_cs_dtrace:dt_wm_return(?MODULE, <<"to_xml">>, [Code], [riak_cs_wm_utils:extract_name(User), Bucket]),
+                    riak_cs_dtrace:dt_bucket_return(?MODULE, <<"list_keys">>, [Code], [riak_cs_wm_utils:extract_name(User), Bucket]),
+                    Response
             end
     end.
+
+-spec get_options(#wm_reqdata{}) -> [{atom(), binary()}].
```
jrwest: the return value for `get_options` and `get_option` should use `{atom(), binary() | undefined}` no?

jrwest:
besides a few minor comments not directly related to the core changes +1. code looks great. all tests are passing for me. ran a few by-hand tests on large buckets locally as well. Only thing not specifically covered by review was performance since I don't have sufficient hardware.

src/riak_cs_list_objects.erl
reiddraper, line 3: true, will fix in a separate PR for the whole project

src/riak_cs_list_objects_fsm.erl
reiddraper: good call, will fix

riak_test/tests/list_objects_test.erl
kellymclaughlin, line 20: yep, that's a good thought. i'd prefer to make a separate pr for that, but we should do that.

include/list_objects.hrl
```diff
@@ -0,0 +1,73 @@
...
+
+        %% the key to start with
+        marker :: binary() | undefined,
+
+        %% The actual response -----------------------------------------------
+        is_truncated :: boolean(),
+
+        contents :: list(list_objects_key_content()),
+
+        common_prefixes :: list(list_objects_common_prefixes())}).
```
kellymclaughlin: @reiddraper I came across this while fixing some dialyzer issues. The type for this field is specified here, but when we call `riak_cs_wm_objects:new_response` we just set it as an ordset and never convert it to this record type. I think for now I'm going to change the type here to be an ordset, but if you would rather do the opposite i'm ok with that also.

kellymclaughlin:
Dialyzer is slightly more happy now

reiddraper:
Dialyzer issues look addressed to me, think we're about ready to merge after one more rebase with master.

kellymclaughlin:
yep, agreed

kellymclaughlin:
+1 to merge

reiddraper:
merged via a 'rebased from master' `keylist-then-mapred-squashed1-rebased2`

PR #384, opened 2013-01-28
Updates for new node package

Vagabond:


kuenishi:
+1. worked fine.

jaredmorrow:
+1 works fine with new changes.  The only weird thing that is unrelated to your PR is a 'bin' directory still gets created despite riak-cs using the sbin directory solely.  Maybe there is something we can do to fix that.

PR #378, opened 2013-01-29
Support for prefix and delimiter when listing objects in a bucket

kellymclaughlin:
Support prefix and delimiter options for GET Objects requests. Includes support for CommonPrefixes in response documents.

### DIFF
--- a/src/riak_cs_list_objects_fsm.erl
+++ b/src/riak_cs_list_objects_fsm.erl
@@ -130,6 +130,14 @@
                          {mapred, Phase :: non_neg_integer(), list()}} |
                         {ReqID :: non_neg_integer, {error, term()}}.
 
+-type key_and_manifest() :: {binary(), lfs_manifest()}.
+-type manifests_and_prefixes() :: {list(key_and_manifest()), ordsets:ordset()}.
+
+-type tagged_item() :: {prefix, binary()} |
+                       {manifest, {binary(), lfs_manifest()}}.
+
+-type tagged_item_list() :: list(tagged_item()).
+
 %%%===================================================================
 %%% API
 %%%===================================================================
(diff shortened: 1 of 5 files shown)


kellymclaughlin:
Running the updated list objects `riak_test` suite relies on [this](https://github.com/basho/erlcloud/pull/6) erlcloud pr.

reiddraper:
I think there is a case where we will return < `max-keys` but `isTruncated` `true`. Here's my rationale:

`riak_cs_list_objects_fsm:enough_results/3` determines whether we've satisfied `max-keys` yet, but we don't take into account the `delimiter` in this call. I suggest that the `num_considerable_keys` field in the record should be adjusted to reflect the keys that are grouped together by `delimiter`. Maybe there could be another call after `filtered_keys_from_request` that produces a length =< length after `filtered_keys_from_request` that takes the delimiter into account. Does this make any sense?

kellymclaughlin:
@reiddraper Pushed update for the riak_test module and rebased from `keylist-then-mapred-squashed1-rebased1`. Test is passing agan for me.

reiddraper:
Sweet! Will give that a shot this weekend.

reiddraper:
The `riak_test` is passing for me, woot!. I'm going to look at double-checking whether [this](https://github.com/basho/riak_cs/pull/378#issuecomment-12913869) is actually a problem or not, but don't see any other potential issues at the moment.

kellymclaughlin:
> I think there is a case where we will return < `max-keys` but `isTruncated` `true`. Here's my rationale:
> 
> `riak_cs_list_objects_fsm:enough_results/3` determines whether we've satisfied `max-keys` yet, but we don't >take into account the `delimiter` in this call. I suggest that the `num_considerable_keys` field in the record should >be adjusted to reflect the keys that are grouped together by `delimiter`. Maybe there could be another call after >`filtered_keys_from_request` that produces a length =< length after `filtered_keys_from_request` that takes the >delimiter into account. Does this make any sense?

From my reading it looks like we might actually have the opposite problem. I think since `enough_results` does not account for an entries in `CommonPrefixes` then we could return a result of size > than `max-keys`. Each `CommonPrefix` is supposed to count as 1 key in the result so maybe passing `CommonPrefixes` into `enough_results` and calculating the results length like this would work:

```
ResultsLength = length(Results) + ordsets:size(CommonPrefixes),
```

Thoughts?

src/riak_cs_list_objects_fsm.erl
```diff
@@ -352,15 +361,120 @@ handle_keys_received(Keys, State=#state{key_buffer=PrevKeyBuffer}) ->
...
+tagged_prefix_list(Prefixes) ->
+    [{prefix, P} || P <- ordsets:to_list(Prefixes)].
+
+-spec untagged_manifest_and_prefix(tagged_item_list()) ->
+    manifests_and_prefixes().
+untagged_manifest_and_prefix(TaggedInput) ->
+    Pred = fun({manifest, _}) -> true;
+              (_Else) -> false end,
+    NotPred = fun(A) -> not Pred(A) end,
+    {A, B} = {lists:filter(Pred, TaggedInput),
```
kellymclaughlin: I think [lists:partition/2](http://www.erlang.org/doc/man/lists.html#partition-2) might be a better choice here.

kellymclaughlin:
1 additional comment, but +1 to merge beyond that

---

addressed in 3dc63e539a157abfdd964a9e901754388554d126 --reid

PR #387, opened 2013-01-29
Fix the how the s3 rewrite module handles query parameters that are not subresources

kellymclaughlin:
This addresses an issue where non-subresource query parameters for requests that also included a subresource in the query parameters were not being transferred to the rewritten URL.


kellymclaughlin:
Closing in favor of [this](https://github.com/basho/riak_cs/pull/399) pr.

PR #388, opened 2013-01-29
use envvar for location of riak_cs riak_test builds

jrwest:
also changes default location to be more like riak_test/riak_ee


kellymclaughlin:
Hah, I love it. I had done this locally, but hadn't gotten around to a pr. +1 to merge, thanks!

PR #389, opened 2013-01-29
Update with latest webmachine changes

reiddraper:
- Use rebased branch of content-length streaming branch (when this gets
  merged to master we'll prefer webmachine master as the dep, this is just
  for testing)
- Don't let webmachine do content checksum validation for us
- Use new location for `webmachine_logger.hrl`


kellymclaughlin:
+1 to merge

Issue #385, opened 2013-01-29
500 Error when listing bucket that does not exist


jrwest:
https://gist.github.com/46238d452af6f86c856a

this causes the `cs296_regression_test` & `cs347_regression_test` to fail with the most recent master - fe6761cce40ee3a68c031b66f8cb76b45abb89b6.

The culprit seems to be a bad match on this line because the return from `get_bucket_acl_policy` is `{error, notfound}`: https://github.com/basho/riak_cs/blob/master/src/riak_cs_wm_utils.erl#L417

kuenishi:
now working on #393

kuenishi:
#393 has been merged.

PR #393, opened 2013-01-30
fix badmatch in case of non-existent bucket ( #385 )

kuenishi:


shino:
tiny typo, s/bad much/bad match/

src/riak_cs_wm_utils.erl
```diff
@@ -414,63 +414,74 @@ bucket_access_authorize_helper(AccessType, Deletable,
...
-            %% undefined here
-            AccessRD = riak_cs_access_logger:set_user(User, RD),
-            Access = PolicyMod:reqdata_to_access(RD, AccessType,
-                                                 User?RCS_USER.canonical_id),
-            case PolicyMod:eval(Access, Policy) of
-                true ->      {false, AccessRD, PermCtx};
-                undefined -> {false, AccessRD, PermCtx};
-                false ->     riak_cs_wm_utils:deny_access(AccessRD, Ctx)
-            end;
+        {error, Reason} when is_atom(Reason) ->
```
shino, line 421: I don't understand why this is_atom(Reason) is sperated from L.424.
This case clause responds "503 Service Unavailable".
On the other hand, {error, _} clause at L.424 responds "404 Object Not Found".

kuenishi:
Error atoms are defined in riak_cs_s3_response and get_bucket_acl_policy may or may not returns such atoms.

I thought returning 'false'  in authorize/2 would cause 403 not authorized, doesn't it?

src/riak_cs_wm_utils.erl
kuenishi, line 421: ok, this case should not be matched, thus I removed this match line to return 500, in case of unexpected errors.

shino:
Authorization is determined in `{ok, {BucketAcl, Policy}}` clause.
These clauses are executed, for example, PB/networking error, so 403 is misleading.

kuenishi:
yep, I removed that patter match to let it die and return 500.

jrwest:
ran the riak_tests against this branch. all previously failing tests passed

PR #396, opened 2013-01-31
Don't attempt to parse an empty XML document.

cmeiklejohn:


kellymclaughlin:
`BadRequest` is not a standard S3 error code. How about `MalformedXML` instead? It might also require some plumbing to add support for that error code to `riak_cs_s3_response`. Then maybe we change `xml_error_code` to have a function clause to check for an empty document and return `MalformedXML`. Thoughts?

Or maybe there is a more appropriate error response. [Here's](http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html) the list of codes.

kellymclaughlin:
I think you're right about `BadRequest`. I made a branch with some changes I thought were a nice improvement. If you think they're ok, then you can merge the [pr](https://github.com/basho/riak_cs/pull/414) into this branch and then I'm +1 to merge it to `master`.

kellymclaughlin:
BTW, I rebased and force pushed this branch to make sure it was current with the latest round of changes to `master`

cmeiklejohn:
Do we just want to close this one and reopen yours against master and merge it?

cmeiklejohn:
Merging.

PR #397, opened 2013-01-31
Cs378 prefix and delimiter eunit

reiddraper:
Add eunit tests for `riak_cs_list_objects_fsm:filter_prefix_keys/3`.


kellymclaughlin:
+1

PR #398, opened 2013-01-31
Reduce fragility of erlcloud_eqc test

kellymclaughlin:
Reduce fragility of empty object list checking in the erlcloud_eqc test. Instead of checking the result of a call to list_objects against a list of fixed response elements just verify the bucket name and an empty content list.


reiddraper:
+1

PR #413, opened 2013-02-12
updated riak_test script.

ksauzz:
- removed building R14B04 which is no longer used.
- checked if a package already exists before download.


kellymclaughlin:
+1

PR #417, opened 2013-02-13
Address some eunit test timeouts

kellymclaughlin:
- Use `eqc:testing_time` instead of a fixed number of iterations for `riak_cs_gc_d_eqc` and `riak_cs_get_fsm_eqc`
- Reduce the number of iterations used in `get_fsm_should_never_fail_intermittently_test_`. This test suite lives up to its name by not failing intermittently. Instead it fails constantly with a time out after 5 long minutes. The reduced iterations should give it a better chance of passing consistently on a wider variety of hardware.

### DIFF
--- a/test/riak_cs_gc_d_eqc.erl
+++ b/test/riak_cs_gc_d_eqc.erl
@@ -60,9 +60,9 @@
 eqc_test_() ->
     {spawn,
      [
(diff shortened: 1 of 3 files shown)


kuenishi:
It worked. +1 to merge.

PR #412, opened 2013-02-13
XML responses lack xmlns attribute

argv0:
This most visibly effects directory listings with s3fs, but might break compatibility elsewhere too.


kellymclaughlin:
+1 to merge

PR #265, opened 2013-02-18
Allow admin commands to be handled on a separate interface.

kellymclaughlin:


kellymclaughlin:
The behavior of `admin_auth_enabled` right now mimics that behavior of `auth_bypass` which means you must still specify an `Authorization` header, but with a valid `KeyId` as the only value. I think eventually it would be nice to have the ability to not even have to specify an `Authorization` header at all if the admin interface was adequately protected otherwise. The thing I am unsure of is if it is acceptable to leave the behavior as-is for now or should we go ahead and implement the desired behavior now.

kuenishi:
By this branch, can we remove special bucket "riak-cs" and "usage", when `admin_auth_enabled` is true?

kellymclaughlin:
@kota Yes, in that mode of operation I do not think we have to reserve any bucket names from the object storage API perspective. That brings up a good point and I will look into enforcing that as part of this set of changes.

reiddraper:
> The thing I am unsure of is if it is acceptable to leave the behavior as-is for now or should we go ahead and implement the desired behavior now.

Let's chat about this today, I think it'd be really handy to just be able to use curl for usage, instead of having to teach people about s3cmd/s3curl for those requests.

> Yes, in that mode of operation I do not think we have to reserve any bucket names from the object storage API perspective.

I think we should still reserve them though, personally. I think it makes things simpler to just _always_ say they are unavailable. It also allows you to turn admin_auth back on with no ill-consequences.

reiddraper:
Update from a mumble convo:
1. we decided to hold off on allowing what are effectively anonymous (no auth header) requests for the admin interface, as it would be too much of a change this far in the release cycle. Still something we want to do though.
2. The `riak-cs` bucket should just always be reserved, simpler that way
3. the `usage` bucket access for usage has been deprecated, 1.3 will drop it entirely, you'll need to use `riak-cs/usage`

kellymclaughlin:
- Verified that s3cmd requests still work for usage requests with the resource moved under `riak-cs`
- Made changes so that admin requests work with no `Authorization` header when `admin_auth_enabled` is `false`

reiddraper:
Maybe not a big deal, but Riak CS immediately exits if the alternative ip/port is the same as the regular one (exits with eaddrinuse)

kellymclaughlin:
Been digging more into webmachine and I believe I see a path forward that might be achievable by tomorrow. I'll try and get it working tonight and we can discuss tomorrow on how to proceed.

cmeiklejohn:
I've prepared a branch on fog which supports this change to the usage bucket, please let me know if this makes 1.3.

kuenishi:
Very nice patch! Or the just add validation code in riak_cs_sup (merge the ip/port into one webmachine and enable admin authorisation)? After the webmachine thing resolved, I'd be happy to +1.

src/riak_cs_web.erl
```diff
@@ -8,17 +8,22 @@
...
+%% @doc Setup the Webmachine dispatch table for the admin API
+-spec admin_api_dispatch_table() -> [dispatch_rule()].
+admin_api_dispatch_table() ->
+    admin_resources(stats_props()).
+
+%% @doc Setup the Webmachine dispatch table for the object storage API
+-spec object_api_dispatch_table() -> [dispatch_rule()].
+object_api_dispatch_table() ->
+    base_resources() ++
         one_three_resources(riak_cs_utils:cs_version()).
```
kuenishi, line 27: Due to this resources, before 1.3 API it looks possible to create a reserved bucket named `riak-cs` or `usage`, doesn't it?

reiddraper:
Working great with the WM branch, +1 from me!

PR #422, opened 2013-02-19
Update include/s3_api.hrl

jj1bdx:
`make rel` stopped with this bug.
- Added a comma into `-record(policy_v1)` definition


kellymclaughlin:
+1 to merge

PR #423, opened 2013-02-19
Fix call to riak_cs_get_fsm:continue after range changes

kellymclaughlin:


shino:
Is there one extra space character at chenged line?

Otherwise, +1 for merge

Issue #363, opened 2012-12-19
PUT copy


kuenishi:
s3fs is using x-amz-copy-source to update metadata.

slfritchie:
For example, to mimic what s3fs is doing
1. Delete the thing, e.g. bucket `test` object `foo`
   
   s3cmd del s3://test/foo
2. Create the thing with some metadata:
   
   s3curl.pl --debug --put ~/foo --id=$MY_AWS_KEY --key=$MY_AWS_SECRET --contentType "text/plain" -- -H "x-amz-meta-custom1: foo0" -H "x-amz-meta-custom2: bar0" -x localhost:8080 http://s3.amazonaws.com/test/foo -v
3. Update the thing with new metadata, like s3fs does:
   
   s3curl.pl --debug --put /tmp/zero --id=$MY_AWS_KEY --key=$MY_AWS_SECRET --contentType "text/plain" -- -H "x-amz-meta-custom1: foo2222" -H "x-amz-meta-custom2: bar2222" -H "x-amz-meta-custom3: bazbazbaz" -H "x-amz-copy-source: /test/foo" -x localhost:8080 http://s3.amazonaws.com/test/foo -v

Output looks like:

```
s3curl: Found the url: host=s3.amazonaws.com; port=; uri=/test/foo; query=;
s3curl: ordinary endpoint signing case
s3curl: StringToSign='PUT\n\ntext/plain\nMon, 18 Feb 2013 16:38:49 GMT\nx-amz-copy-source:/test/foo\nx-amz-meta-custom1:foo2222\nx-amz-meta-custom2:bar2222\nx-amz-meta-custom3:bazbazbaz\n/test/foo'
s3curl: signature='j7OXwFdDif/hAEz54Bj9uh9SVnk='
s3curl: exec curl -H Date: Mon, 18 Feb 2013 16:38:49 GMT -H Authorization: AWS J2IP6WGUQ_FNGIAN9AFI:j7OXwFdDif/hAEz54Bj9uh9SVnk= -L -H content-type: text/plain -T /tmp/zero -H x-amz-meta-custom1: foo2222 -H x-amz-meta-custom2: bar2222 -H x-amz-meta-custom3: bazbazbaz -H x-amz-copy-source: /test/foo -x localhost:8080 http://s3.amazonaws.com/test/foo -v
* About to connect() to proxy localhost port 8080 (#0)
*   Trying ::1...
* Connection refused
*   Trying 127.0.0.1...
* connected
[...]

PR #425, opened 2013-02-19
Add limited support for PUT copy + Content-Length: 0, to support S3FS

slfritchie:
Address GH #363 

The S3FS FUSE client uses the feature of PUT copy + "Content-Length: 0"
to update only the metadata of an S3 object.  This metadata update is
used to support inode-modifying operations such as "touch", "chmod",
and "chown".

Add a new member to the #key_context record:

```
update_metadata=false :: boolean()
```

In valid_entity_length(), check to see if "Content-Length: 0" and also
the "x-amz-copy-source" header is present and (if yes) if its value
is exactly equal to the object's path.

In accept_body(), check the new #key_context.update_metadata flag.
If true, update the ACL creation time (so that
riak_cs_manifest_resolution.erl logic can pick the newest version of the
manifest) and then write the updated manifest using
riak_cs_utils:set_object_acl().

### DIFF
--- a/src/riak_cs_wm_object.erl
+++ b/src/riak_cs_wm_object.erl
@@ -65,7 +65,7 @@ allowed_methods() ->
     ['HEAD', 'GET', 'DELETE', 'PUT'].
 
 -spec valid_entity_length(#wm_reqdata{}, #context{}) -> {boolean(), #wm_reqdata{}, #context{}}.
-valid_entity_length(RD, Ctx=#context{local_context=LocalCtx}) ->
+valid_entity_length(RD, Ctx) ->
     case wrq:method(RD) of
         'PUT' ->
             case catch(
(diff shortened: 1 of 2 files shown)


src/riak_cs_wm_object.erl
```diff
@@ -219,6 +218,27 @@ content_types_accepted(RD, Ctx=#context{local_context=LocalCtx0}) ->
 
 -spec accept_body(#wm_reqdata{}, #context{}) -> {{halt, integer()}, #wm_reqdata{}, #context{}}.
 accept_body(RD, Ctx=#context{local_context=LocalCtx,
+                             user=_User,
```
argv0: Do we need to match this out of #context if we're not using it?

argv0:
tested per instructions and with s3fs, works fine.  one minor nit in the source comments, otherwise +1 to merge.

PR #427, opened 2013-02-19
Wrd object access refactor

reiddraper:
See commit comment for details

### DIFF
--- a/src/riak_cs_wm_utils.erl
+++ b/src/riak_cs_wm_utils.erl
@@ -509,22 +509,34 @@ is_acl_request(_) ->
                                      RD::term(), Ctx::term()) ->
     authorized_response().
 object_access_authorize_helper(AccessType, Deletable,
-                               RD, #context{user=User,
-                                            policy_module=PolicyMod,
+                               RD, #context{policy_module=PolicyMod,
                                             local_context=LocalCtx,
                                             riakc_pid=RiakPid}=Ctx)
   when ( AccessType =:= object_acl orelse
          AccessType =:= object )
        andalso is_boolean(Deletable) ->
 
+    #key_context{bucket=Bucket} = LocalCtx,
+    case translate_bucket_policy(PolicyMod, Bucket, RiakPid) of
+        {error, multiple_bucket_owners=E} ->
+            %% We want to bail out early if there are siblings when
+            %% retrieving the bucket policy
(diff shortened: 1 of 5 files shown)


src/riak_cs_wm_utils.erl
```diff
@@ -544,56 +532,133 @@ object_access_authorize_helper(AccessType, Deletable,
...
+-spec translate_bucket_policy(atom(), binary(), pid()) ->
+    policy() | undefined.
+translate_bucket_policy(PolicyMod, Bucket, RiakPid) ->
+    case PolicyMod:bucket_policy(Bucket, RiakPid) of
+        {ok, P} ->
+            P;
+        {error, policy_undefined} ->
+            undefined;
+         {error, _Reason} ->
+             %% riak_cs_s3_response:api_error(Reason, RD, Ctx),
```
kellymclaughlin: This was one of the things motivating my suggestion to refactor this. The only other error that should be returned is `multiple_bucket_owners` and I think we should fail the request on that.

src/riak_cs_wm_utils.erl
```diff
@@ -515,26 +518,11 @@ object_access_authorize_helper(AccessType, Deletable,
...
-
-    Policy = PolicyMod:bucket_policy(Bucket, RiakPid),
+    #key_context{bucket=Bucket, manifest=Manifest} = LocalCtx,
+    CanonicalId = extract_canonical_id(User),
+    RequestedAccess = requested_access_helper(AccessType, Method),
+    ObjectAcl = extract_object_acl(Manifest),
+    Policy = translate_bucket_policy(PolicyMod, Bucket, RiakPid),
     Access = PolicyMod:reqdata_to_access(RD, AccessType, CanonicalId),
 
     case {riak_cs_acl:object_access(Bucket,
```
kellymclaughlin, line 540: Any reason not to take another step and refactor this case into a function? It was one of the more complex parts of the bucket helper refactor so I can understand if that's the reasoning, but just wanted to get your thoughts.

slfritchie:
The obnoxious `{error,function_clause,[{riak_cs_s3_policy,eval,[{access_v1,'PUT',.....` error has disappeared.  The refactoring look nice, no horrible dialyzer warnings added.  +1 to merge, thanks!

kellymclaughlin:
+1 from me also

PR #395, opened 2013-02-20
dialyzer and xref review before 1.3 release

argv0:
Avoid mistakes that can be caught with a little static-analysis.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.dialyzer.raw-output
 .eunit/*
 deps/*
 rel/riak-cs
--- a/Makefile
+++ b/Makefile
@@ -134,7 +134,7 @@ dialyzer: compile
 	@echo
 	@sleep 1
 	dialyzer -Wno_return -Wunmatched_returns --plt $(PLT) deps/*/ebin ebin | \
-	    egrep -v -f ./dialyzer.ignore-warnings
+	    tee .dialyzer.raw-output | egrep -v -f ./dialyzer.ignore-warnings
 
 cleanplt:
 	@echo
--- a/Makefile
+++ b/Makefile
@@ -144,6 +144,9 @@ cleanplt:
 	sleep 5
 	rm $(PLT)
 
+xref: compile
+	./rebar xref skip_deps=true | grep -v unused | egrep -v -f ./xref.ignore-warnings
+
 ##
 ## Packaging targets
 ##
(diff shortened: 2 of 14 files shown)


argv0:
This leaves us with 4 dialyzer warnings.  I've added a couple warnings to the ignore file  for things we suspect are dead code and have either instrumented to throw a stacktrace or have heavily commented.

I also added a simple 'xref' target and corresponding ignore file to the Makefile. 

Here's the remaining warnings:

```
riak_cs_acl.erl:126: The pattern {'error', 'notfound'} can never match the type {'error','acl_undefined' | 'multiple_bucket_owners'}
```

A weird one I hope to fix before 1.3.

```
riak_cs_gc.erl:84: The pattern [{UUID, _} | _] can never match the type []
```

Ditto.

```
riak_cs_mp_utils.erl:618: The variable _ can never match since previous clauses completely covered the type #lfs_manifest_v3{version::integer(),block_size::'undefined' | integer(),bkey::'undefined' | {binary(),binary()},metadata::'undefined' | [{_,_}],uuid::'undefined' | binary(),content_length::'undefined' | non_neg_integer(),content_type::'undefined' | binary(),state::'active' | 'deleted' | 'pending_delete' | 'scheduled_delete' | 'undefined' | 'writing',write_blocks_remaining::'undefined' | [integer()],delete_blocks_remaining::'undefined' | [integer()],acl::'no_acl_yet' | 'undefined' | #acl_v1{owner::{_,_} | {_,_,_},grants::[any()],creation_time::{_,_,_}} | #acl_v2{owner::{_,_} | {_,_,_},grants::[any()],creation_time::{_,_,_}},props::'undefined' | [atom() | tuple()],cluster_id::'undefined' | binary()}
```

Dialyzer says we don't need that clause:  @slfritchie - can you confirm/deny?

```
riak_cs_s3_policy.erl:284: Guard test is_list(KeyBin::binary()) can never succeed
```

Dialyzer says dead code, but its in a place I'm not familiar with and which we probably don't have extensive coverage for.

reiddraper:
> `riak_cs_gc.erl:84: The pattern [{UUID, _} | _] can never match the type []`

I spent a while tracking that one down before, the best I could come up with is that list comprehensions w/ dialyzer don't like it when the list could be empty, but I could never prove that for sure.

reiddraper:
> `riak_cs_acl.erl:126: The pattern {'error', 'notfound'} can never match the type {'error','acl_undefined' | 'multiple_bucket_owners'}`

This one is also confusing me

slfritchie:
+1 to merge(*), zillions of thanks to @argv0 for all of this work!

(*) Given we're so close to code freeze, I'll do the merge myself in a few minutes.

PR #428, opened 2013-02-20
update rebar to 2.1.0 pre

reiddraper:
See commit message, main deal is that it allows you to run dialyzer w/ deps now


kellymclaughlin:
+1 to merge

PR #429, opened 2013-02-21
Eliminate last Dialyzer warnings

slfritchie:
... or they're as harmless as I can figure out.  This brings us to zero, let's try to keep it that way.


argv0:
Awesome - merging now.

PR #434, opened 2013-02-25
Use named WM router table instead of starting them

reiddraper:
- Update to work with basho/webmachine#128
- Builds on #265

@seancribbs' Webmachine change uses a single Webmachine router
(the way it used to be) and uses multiple keys in an ETS table
to support multiple dispatch routes. This change removes the custom
routers from the supervisor and adds a `router_name` to the two
webmachine options. Only one of them is used if the admin and web
config share the same interface.

### DIFF
--- a/src/riak_cs_sup.erl
+++ b/src/riak_cs_sup.erl
@@ -139,7 +139,7 @@ web_spec(Name, Config) ->
 object_web_config(Options) ->
     [{dispatch, riak_cs_web:object_api_dispatch_table()},
      {name, object_web},
-     {router_name, object_web},
+     {dispatch_group, object_web},
      {ip, proplists:get_value(cs_ip, Options)},
      {port, proplists:get_value(cs_port, Options)},
      {nodelay, true},
@@ -152,7 +152,7 @@ object_web_config(Options) ->
 admin_web_config(Options) ->
     [{dispatch, riak_cs_web:admin_api_dispatch_table()},
      {name, admin_web},
-     {router_name, admin_web},
+     {dispatch_group, admin_web},
      {ip, proplists:get_value(admin_ip, Options)},
      {port, proplists:get_value(admin_port, Options, 8000)},
      {nodelay, true},
--- a/src/riak_cs_sup.erl
+++ b/src/riak_cs_sup.erl
@@ -110,7 +110,6 @@ web_specs(Options) ->
                 [{admin_web, admin_web_config(Options)},
                  {object_web, object_web_config(Options)}]
         end,
-    [router_spec(Name) || {Name, _} <- WebConfigs] ++
     [web_spec(Name, Config) || {Name, Config} <- WebConfigs].
 
 -spec pool_specs(proplist()) -> [supervisor:child_spec()].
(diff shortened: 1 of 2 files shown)


reiddraper:
Reminder, don't merge the `rebar.config` changes in when merging this, it's just for testing before basho/webmachine#128 gets merged.

slfritchie:
From Reid:

```
okie, so you'll notice in the app.config as of that branch,
there is a new commented section, for the 'admin interface'
that has a port and host
if you uncomment it, the admin resources should then only listen on that interface
(things like user creation)
so what i did is just change the client tests
to hardcode localhost:$PORT
to 8000 for user creation
and make sure the client tests still pass
```

slfritchie:
So, I've tried following that recipe, trying the following:
- Edit client_tests/clojure/clj-s3/test/java_s3_tests/test/client.clj client_tests/python/boto_test.py
  - Replace port 8080 with port 8000
- Uncomment the `admin_ip` and `admin_port` entries from app.config, add `{cs_version, 010300}`, restart CS.
- `make compile-client-test`
- env PATH=/opt/local/bin:$PATH make -k test-clojure test-python
  - Using `env` to get a version of Python that can access the Boto libs

But I see failures both for Clojure and for Python.

```
5 facts were not confirmed. (But 1 was.)
Subprocess failed
make: *** [test-clojure] Error 5
```

and

```
Ran 8 tests in 0.034s

FAILED (errors=7)
```

If I change the client proxy port number back to 8080, I see more errors than that, so there's some kind of difference at work.

If I use the old-school curl command to create a new user and aim it at port 8000, it works fine the first time and correctly complains about an existing user if I repeat it.

```
curl -v -H 'Content-Type: application/json' -X POST http://localhost:8000/riak-cs/user --data '{"email":"foobarYYY@foobar.com", "name":"foo barYYY"}'
```

reiddraper:
@slfritchie did you use the webmachine branch that the rebar.config calls for?

slfritchie:
I thought that I had, that it was included in the `rebar.config` diff, and I'd done `make clean`, `rebar update-deps` twice, then `make stage` followed by the app.config changes etc.  Hrrm.

reiddraper:
Just ran this again this morning without issue. Can you try checking out the branch and `rm -rf deps/webmachine; make deps compile` and then re-run the client tests with the port # change (and the app.config interface change too)?

src/riak_cs_sup.erl
```diff
@@ -145,8 +138,8 @@ web_spec(Name, Config) ->
 -spec object_web_config(proplist()) -> proplist().
 object_web_config(Options) ->
     [{dispatch, riak_cs_web:object_api_dispatch_table()},
-     {default_router, false},
      {name, object_web},
+     {router_name, object_web},
```
kellymclaughlin: s/router_name/dispatch_list based on latest wm change

src/riak_cs_sup.erl
```diff
@@ -159,6 +152,7 @@ object_web_config(Options) ->
 admin_web_config(Options) ->
     [{dispatch, riak_cs_web:admin_api_dispatch_table()},
      {name, admin_web},
+     {router_name, admin_web},
```
kellymclaughlin: s/router_name/dispatch_list based on latest wm change

seancribbs: dispatch_group, actually

reiddraper:
I can't change the base-branch target of this PR, so instead of merging to master once it's done, we'll just manually merge to `release-1.3`. So please don't press the merge button.

kellymclaughlin:
Tests are now passing for me with a little work. There is a bug in `riak_cs_s3_auth` right [here](https://github.com/basho/riak_cs/blob/develop/src/riak_cs_s3_auth.erl#L142) and [here](https://github.com/basho/riak_cs/blob/develop/src/riak_cs_s3_auth.erl#L152) where it calls `canonicalize_qs/1` instead of `canonicalize_qs/2`. That can probably be addressed in a separate pr though so +1 to merge this.

reiddraper:
> There is a bug in riak_cs_s3_auth right here and here where it calls canonicalize_qs/1 instead of canonicalize_qs/2.

Nice find. I'll get this merged in and then we can address the authorization issue today.

reiddraper:
Merged into `release-1.3`.

PR #435, opened 2013-02-26
Fix auth signature for multiple query strings

reiddraper:
The s3 auth signature was (potentially) calculated wrong when multiple
subresource query strings were in a request. The bug happened when
the query strings weren't already in sorted order. This change
sorts them first, before the signature is caclulated. The bug
was found using boto.


slfritchie:
+1, looks fine

PR #437, opened 2013-02-26
quoted etag

reiddraper:


seancribbs:
We should also fix webmachine to do this correctly. Making the change in webmachine-ruby was "interesting".

slfritchie:
It works with your Boto test script, s3cmd remains happy, and the dialyzer warning list is still clean.  +1 to merge, thanks!

PR #439, opened 2013-02-28
Add simple multipart upload tests to python test

reiddraper:
Make sure that `{enforce_multipart_part_size, false}` is set in your
`app.config` before running the `make test-python` test, because this
test uses small strings as the individual parts. A future test will
exercise larger objects.

A couple of notes on the Python. `StringIO` is basically a wrapper for
a string that acts like a file-like object. Keys in Boto can be iterated
(which streams the bytes from s3), which is how we calculate the Md5 in
`md5_from_key`.

This commit also does some general cleanup. I used `pylint` to guide me
in that. The old tests and the multipart tests are in separate classes,
so in the future they could be separate Make targets, if we wanted to
separate them for speed-of-running reasons. I also enabled a
commented-out test that now runs fine.


reiddraper:
Closing to retarget the `release/1.3` branch

PR #441, opened 2013-02-28
Fix issue with query string canonicalization

kellymclaughlin:


reiddraper:
I tested this PR w/ #443 , and now the `riak_test` passes. I also merged locally with my boto multipart test and that passes as well. +1

PR #436, opened 2013-02-27
Put to a non-existent bucket -> HTTP 500 error 

kuenishi:
s3cmd put ~/foo s3://testbotoXX/foo where the S3 bucket `testbotoXX` does not exist:

```
15:12:21.878 [error] webmachine error: path="/buckets/testbotoXX/objects/foo"
{error,{case_clause,{error,notfound}},[{riak_cs_wm_utils,translate_bucket_policy,3,[{file,"src/riak_cs_wm_utils.erl"},{line,596}]},{riak_cs_wm_utils,object_access_authorize_helper,4,[{file,"src/riak_cs_wm_utils.erl"},{line,520}]},{riak_cs_wm_common,authorize,2,[{file,"src/riak_cs_wm_common.erl"},{line,309}]},{riak_cs_wm_common,forbidden,2,[{file,"src/riak_cs_wm_common.erl"},{line,147}]},{webmachine_resource,resource_call,3,[{file,"src/webmachine_resource.erl"},{line,183}]},{webmachine_resource,do,3,[{file,"src/webmachine_resource.erl"},{line,141}]},{webmachine_decision_core,resource_call,1,[{file,"src/webmachine_decision_core.erl"},{line,48}]},{webmachine_decision_core,decision,1,[{file,"src/webmachine_decision_core.erl"},{line,212}]}]}
```

PR #442, opened 2013-02-28
Fix for 500 error when putting an object to a non-existent bucket

kellymclaughlin:
Fix issue where attempting to put an object to a non-existent bucket results in a `500` instead of the expected `404`.  This is a regression introduced by some recent refactoring. Also included is a `riak_test` regression test for this issue.

### DIFF
--- a/src/riak_cs_wm_utils.erl
+++ b/src/riak_cs_wm_utils.erl
@@ -522,6 +522,10 @@ object_access_authorize_helper(AccessType, Deletable,
             %% We want to bail out early if there are siblings when
             %% retrieving the bucket policy
             riak_cs_s3_response:api_error(E, RD, Ctx);
+        {error, notfound} ->
+            %% The call to `check_bucket_exists' returned `notfound'
+            %% so we can assume to bucket does not exist.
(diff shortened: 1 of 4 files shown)


reiddraper:
Test passes, manual test as well. Code looks good. +1

src/riak_cs_wm_utils.erl
```diff
@@ -598,8 +602,10 @@ translate_bucket_policy(PolicyMod, Bucket, RiakPid) ->
             P;
         {error, policy_undefined} ->
             undefined;
-        {error, multiple_bucket_owners}=Error ->
-             Error
+        {error, notfound}=Error1 ->
```
kuenishi, line 605: Before here, I think it's better to return `{error, no_such_bucket}` at `check_bucket_exists/2`, because return type of its error becomes `{error, so-called-s3-error-response-code-atom}` . `notfound` is an atom of Riak client error.

see https://github.com/basho/riak_cs/pull/436/files#L0R904 for code.

kuenishi:
Anyway, thank you for adding great regression test!

PR #443, opened 2013-03-01
Missing quotes around etag in list uploads part results

kellymclaughlin:
Add the expected quotes around etags for each part of a multipart upload when listing the parts of an upload. This resolves a test failure with the `mp_upload_test` `riak_test` module.


kellymclaughlin:
Just for clarification there is another open [pr](https://github.com/basho/riak_cs/issues/441) that addresses a different failure with the `mp_upload_test` module.

reiddraper:
I tested this PR w/ #441, and now the `riak_test` passes. I also merged locally with my boto multipart test and that passes as well. +1

PR #445, opened 2013-03-01
CS Ignores If-Modified-Since header

shino:
I don't know wherther it's a bug or just not supported, If-Modified-Since header is ignored in CS.


kellymclaughlin:
Can we get a `riak_test` module for this? If you don't have time let me know and I can write it.

shino:
I will do.

shino:
close this issue #445 in favor of #461

PR #449, opened 2013-03-01
Use tag of riakc with fixed deps

jaredmorrow:


slfritchie:
+1, merging now.

reiddraper:
This unfortunately breaks things, because the newly created riakc `1.3.2` relies on riak_pb `1.3.0` which does not have the equivalent change for bucket properties, 0709dedcba49b6508353c9e3695ecfa5e24cd5b1. So CS exits w/ 

```
18:58:30.544 [error] Supervisor riak_cs_access_archiver_sup had child access_archiver_riak_client started with {riakc_pb_socket,start_link,undefined} at <0.269.0> exit with reason call to undef
ined function riak_pb_codec:decode_bucket_props({rpbbucketprops,3,true}) in context child_terminated
```

reiddraper:
The two commits that need to either be included together or not are:
- [riak_pb](https://github.com/basho/riak_pb/commit/0709dedcba49b6508353c9e3695ecfa5e24cd5b1)
- [riak-erlang-client](https://github.com/basho/riak-erlang-client/commit/913fa1608a4561cc20a967dc428b2a606f7207ed)

jaredmorrow:
I can tag riak_pb with a new version and update the riakc tag to fix the issue.

PR #451, opened 2013-03-04
app.config cleanup

kellymclaughlin:
Add an entry for `cs_version` to the app.config template. I think this should work since most people who are upgrading will keep their existing config files. 

Also remove some obsolete or more obscure options. _e.g._ I could not find `put_fsm_buffer_size_max` referenced anywhere in the code. Some of the options I removed may be needed, but it seems like a better approach is to handle documenting them instead of cluttering up the config file with them. Open to feedback on my choices in that regard.

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -3,10 +3,27 @@
 [
  %% Riak CS configuration
  {riak_cs, [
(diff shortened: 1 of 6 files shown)


reiddraper:
+1 looks good

PR #454, opened 2013-03-04
Velvet dependency to `release/1.3` branch

reiddraper:
Should mean we use lager 1.2.2 now, as shown
from `./rebar list-deps`.


slfritchie:
+1 make it so!

PR #455, opened 2013-03-05
add riak_tests for bucket policy

kellymclaughlin:
before this we need add bucket policy interface to erlcloud.


kellymclaughlin:
Accidentally open pr against wrong branch, sorry.

PR #453, opened 2013-03-05
riak_test module to exercise stats

kellymclaughlin:
Add `riak_test` module for the stats resource. Also refactor the cluster setup in the `rtcs` module so that the admin user credentials are set and the `riak_cs` and `stanchion` nodes are restarted for those settings to take effect. The `rtcs:setup` function return value has been changed to also return the admin account info and allows for some duplicated code to be eliminated in most of the test modules. These changes resolve a race condition where user creation is attempted prior to the `riak_cs` and `stanchion` applications being fully started.

Two caveats for this pr:
1. The refactoring in `rtcs` could be improved even more. I have decided to leave it as-is for the time being because I think the time right now is better spent doing other testing. 
2. The stats test is not exhaustive and could be expanded to do more. The impetus for creating this test was more to test the authentication code path it uses rather than the correct functioning of stats collection. The test does some functional verification of stats collection, but again I have elected to defer more work on it in favor or other testing tasks.

### DIFF
--- a/riak_test/src/rtcs.erl
+++ b/riak_test/src/rtcs.erl
@@ -25,20 +25,26 @@
 -define(DEFAULT_PROTO, "http").
 
 setup(NumNodes) ->
-    Nodes = build_cluster(NumNodes),
+    %% Start the erlcloud app
+    erlcloud:start(),
+
+    {{AdminKeyId, AdminSecretKey},
+     {RiakNodes, _CSNodes, _Stanchion}=Nodes} = build_cluster(NumNodes),
+
+    AdminConfig = rtcs:config(AdminKeyId,
+                              AdminSecretKey,
+                              rtcs:cs_port(hd(RiakNodes))),
+
     %% STFU sasl
     application:load(sasl),
     application:set_env(sasl, sasl_error_logger, false),
-    %% Start the erlcloud app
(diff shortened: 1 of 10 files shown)


riak_test/src/rtcs.erl
```diff
@@ -324,22 +413,31 @@ deploy_stanchion(Config) ->
...
     Cmd="curl -s -H 'Content-Type: application/json' http://localhost:" ++
         integer_to_list(Port) ++
         "/riak-cs/user --data '{\"email\":\"" ++ EmailAddr ++  "\", \"name\":\"" ++ Name ++"\"}'",
     lager:info("Cmd: ~p", [Cmd]),
-    Output = os:cmd(Cmd),
+    Delay = rt:config(rt_retry_delay),
+    Retries = rt:config(rt_max_wait_time) div Delay,
+    OutputFun = fun() -> os:cmd(Cmd) end,
+    Condition = fun(Res) -> Res /= [] end,
+    Output = wait_until(OutputFun, Condition, Retries, Delay),
```
shino, line 426: Nice retry logic!

shino:
Looks good and the test passes.
Please rebase (as mentioned in #461) and refactor.
+1 to merge after that.

PR #460, opened 2013-03-06
Handle invalid values of max-keys

kellymclaughlin:
Return 400 Bad Request when a non-integer value is specified for the max-keys parameter to a list objects request.


shino:
+1 to merge.

PR #468, opened 2013-03-06
Riak_test driver of the CS client tests

slfritchie:
This is the minimal (?) amount necessary to get the client tests (Clojure, Python, Erlang) tests wrapped up and have `riak_test` drive them.

### DIFF
--- a/riak_test/src/rtcs.erl
+++ b/riak_test/src/rtcs.erl
@@ -47,6 +47,7 @@ setup(NumNodes, Configs) ->
 build_cluster(NumNodes, Configs) ->
     {_, {RiakNodes, _, _}} = Nodes =
         deploy_nodes(NumNodes, Configs),
+
     rt:wait_until_nodes_ready(RiakNodes),
     lager:info("Build cluster"),
     rtcs:make_cluster(RiakNodes),
@@ -129,9 +130,13 @@ ee_config() ->
     ].
 
 cs_config() ->
+    cs_config([]).
+
+cs_config(UserExtra) ->
     [
      lager_config(),
      {riak_cs,
+      UserExtra ++
       [
        {proxy_get, enabled},
        {anonymous_user_creation, true},
--- a/riak_test/tests/external_client_tests.erl
+++ b/riak_test/tests/external_client_tests.erl
@@ -6,20 +6,9 @@
 -define(TEST_BUCKET, "riak_test_bucket").
 
 confirm() ->
-    CS_extra = [{enforce_multipart_part_size, false}],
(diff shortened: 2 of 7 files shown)


slfritchie:
Reviewer: please add your documentation wish list for Santa Claus.

riak_test/src/rtcs.erl
```diff
@@ -25,7 +25,10 @@
 -define(DEFAULT_PROTO, "http").
 
 setup(NumNodes) ->
-    Nodes = build_cluster(NumNodes),
+    setup(NumNodes, ee_config(), stanchion_config(), cs_config()).
+
+setup(NumNodes, EE_config, Stanchion_config, CS_config) ->
```
kellymclaughlin: `EE_config`, `CS_config`, etc make it difficult to distinguish from the `ee_config()` and `cs_config()` function calls when reading. Can we change that to be camel-cased like `EEConfig`?

riak_test/tests/external_client_tests.erl
```diff
@@ -0,0 +1,52 @@
...
+    FirstNode = hd(RiakNodes),
+
+    {AccessKeyId, SecretAccessKey} = rtcs:create_user(FirstNode, 1),
+
+    %% User config
+    CS_http_port = rtcs:cs_port(FirstNode),
+    _UserConfig = rtcs:config(AccessKeyId, SecretAccessKey, CS_http_port),
+
+    lager:debug("cs_src_root = ~p", [rtdev:relpath(cs_src_root)]),
+    %% NOTE: This 'riak_cs_root' path must appear in
```
kellymclaughlin: I think `riak_cs_root` here should be `cs_src_root`.

kellymclaughlin:
The documentation I would suggest is to add a comment about the requirement for `cs_src_root` in the `riak_test` config file to `riak_test/README.md` and also add it to the example config file given there. That was all I had to add to get it working.

slfritchie:
Kelly, I think that your review comments have been addressed now.  Do you have time for another peek?

kellymclaughlin:
We have both solved the problem of wanting to set per-test config options, but slightly differently. If it's ok with you I will rebase this branch and work on sorting out the conflicts and make everything play nice together.

slfritchie:
Kelly, sorry if I overlooked some existing code!  Go ahead.

kellymclaughlin:
I've rebased and adjusted the `external_client_tests` module as needed. The tests are all passing for me now so +1 to merge if you are ok with my changes.

slfritchie:
I liked skipping the extra time required to knit and tear down a 4 node Riak cluster, but ... not a big deal.  Merging.

PR #467, opened 2013-03-06
Fix bucket ACL update incompatabilities

kellymclaughlin:
Fix two incompatibilities with bucket ACLs:
- Bucket owners should always be granted all ACL permissions except READ even when not explicitly granted. The first fix addresses the issue that bucket owners were being denied any access that was not explicitly granted. 
- Fix an issue with ACL updates when specified display_name for Owner element and a Grant element for the bucket owner do not match. This has the side-effect of avoiding a 2I query to riak in the case where the bucket owner is also the grantee.

These bugs were found running the `s3tests.functional.test_s3:test_bucket_acl_xml_write` test in the [ceph s3-test](https://github.com/ceph/s3-tests) suite.

### DIFF
--- a/src/riak_cs_acl.erl
+++ b/src/riak_cs_acl.erl
@@ -143,6 +143,9 @@ bucket_access(_, RequestedAccess, CanonicalId, RiakPid, Acl) ->
             true;
         true ->
             {true, owner_id(Acl, RiakPid)};
+        false when IsOwner == true
+                   andalso RequestedAccess /= 'READ' ->
+            true;
         _ ->
             false
     end.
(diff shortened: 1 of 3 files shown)


slfritchie:
There's a new Dialyzer warning, I think:

```
riak_cs_acl_utils.erl:322: The pattern <[], {{[], CanonicalId}, _Perms}, {DisplayName, _}, _> can never match the type <_,{'AllUsers' | 'AuthUsers' | {[any()],_},['FULL_CONTROL' | 'READ' | 'READ_ACP' | 'WRITE' | 'WRITE_ACP']},{string(),string(),string()},_>
```

slfritchie:
Other than that, I think this PR's OK.

kellymclaughlin:
Removed the offending function clause and dialyzer is happier.

PR #473, opened 2013-03-07
Add helper fun to handle non bools in dtrace returns

reiddraper:
- Add comment about rationale behind some HTTP status codes

Tested by either trying to upload a > 5GB file, or setting the default max content length to something like 1MB.

### DIFF
--- a/riak_test/src/rtcs.erl
+++ b/riak_test/src/rtcs.erl
@@ -30,13 +30,12 @@ setup(NumNodes) ->
 setup(NumNodes, Configs) ->
     %% Start the erlcloud app
     erlcloud:start(),
-    Cfgs = configs(Configs),
-    Nodes = build_cluster(NumNodes, Cfgs),
 
     %% STFU sasl
     application:load(sasl),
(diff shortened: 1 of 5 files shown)


kellymclaughlin:
This looks good, but I'm going to attempt to expand our riak_test module for mp uploads to test this.

kellymclaughlin:
+1 on this from me. Merge away once taking a look at my riak_test additions.

kellymclaughlin:
I've merged the riak_test refactoring now so probably should rebase this from `release/1.3` prior to merging to address any conflicts and update the expected return from `rtcs:setup` in the new test module. Happy to do that after you take a look.

reiddraper:
+1

> Happy to do that after you take a look.

Yeah, if you don't mind going through the rebase conflicts, I'd appreciate it.

kellymclaughlin:
Rebased and resolved all conflicts. All the `riak_test` modules are passing. So I think it's ready to merge if you're ok with it.

reiddraper:
+1 to merge

PR #459, opened 2013-03-06
Add eqc and basic tests for bucket policy

kuenishi:
still WIP, revival of #440

PR #477, opened 2013-03-08
add simple boto test of bucket policy

kuenishi:


kuenishi:
Will be rebased to ku-bucket-policy-eqc, and will be done via #459. closing.

PR #485, opened 2013-03-12
Compress terms before writing to Riak

reiddraper:
I was doing some multipart upload testing an was seeing max object sizes (from Riak stats) relatively quickly get over 2MB (if you bang on the same file name over and over again). This change uses the `compression` option with `erlang:term_to_binary/2` before writing thing to Riak. This is transparently decompressed by `binary_to_term/1`. After the change, max file size dropped by approx. a factor of 4, sometimes more. I also combined all our uses of t2b (for persistence) into one place, so we could change it there later if need be.

### DIFF
--- a/include/riak_cs.hrl
+++ b/include/riak_cs.hrl
@@ -9,6 +9,7 @@
 -define(PART_MANIFEST, #part_manifest_v1).
 -define(PART_MANIFEST_RECNAME, part_manifest_v1).
 -define(MULTIPART_DESCR, #multipart_descr_v1).
+-define(COMPRESS_TERMS, false).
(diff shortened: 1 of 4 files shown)


slfritchie:
Reid, any sense of the CPU overhead of this?  For lots of multiparts, there are going to be several/dozens/hundreds of UUID'ish bits which aren't going to compress well.  Intuition says that since CS has a higher CPU load than the underlying Riak nodes for most workloads that we've seen so far, this patch will push that CPU load higher.  Data is better than intuition, if we have it.

reiddraper:
Yes this is a good point. Agree more data is needed before merging.

kellymclaughlin:
+1 to merge

PR #488, opened 2013-03-13
Repin riakc and riak_repl_pb_api deps so they are compatible.

seancribbs:
This PR is to compensate for some miscommunication and bad release process from related repos.

New dependency tags and their transitive deps:
- `riakc = 1.3.1.1` (based off `1.3` release branch)
- `riak_repl_pb_api = 0.2.2`
- `riak_pb = 1.3.0`
- `protobuffs = 0.8.0`


reiddraper:
+1, thanks

PR #494, opened 2013-03-15
Update xref ignores

reiddraper:
`make xref` should just return

```
==> riak_cs (xref)
ERROR: xref failed while processing /Users/reid/Documents/repos/basho/riak_cs: rebar_abort
```


argv0:
+1

PR #493, opened 2013-03-15
License headers

argv0:
Review suggestion:

for i in `git ls-files`; do open -W -a /Applications/Emacs.app/ $i; done

(-W waits for the app to exit)

### DIFF
--- a/.plugins/rebar_test_plugin.erl
+++ b/.plugins/rebar_test_plugin.erl
@@ -1,8 +1,22 @@
-%% -------------------------------------------------------------------
+%% ---------------------------------------------------------------------
 %%
-%% Copyright (c) 2012 Basho Technologies, Inc.  All Rights Reserved.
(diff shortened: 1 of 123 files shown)


reiddraper:
Do we want to worry about doing this for the non-Erlang files too, like the python, clojure, etc. tests?

kellymclaughlin:
I would think we probably should add them to the tests too.

kuenishi:
OTP includes copyright notice in its tests.

https://github.com/erlang/otp/blob/maint/lib/kernel/test/code_a_test.erl

kuenishi:
I rebased to the latest and add some headers and seems covered the whole source code (except bash scripts). Anything else to add header?

kuenishi:
If nothing, +1 to merge

kellymclaughlin:
+1

PR #495, opened 2013-03-16
Chunked md5 function 

kellymclaughlin:


kellymclaughlin:
eqc test is in-progress

kellymclaughlin:
The one thing I'm not certain about is whether the default chunk size is the proper one or if we should be more conservative and go with 4096.

reiddraper:
Code looks good. The EQC test has a flaw though, it never generates binaries bigger than the smallest chunk size, so the following implementation of `chunked_md5` still passes the test:

``` erlang
chunked_md5(<<>>, Context, _ChunkSize) ->
    Context;
chunked_md5(Data, Context, ChunkSize) ->
    case byte_size(Data) < ChunkSize of
        true ->
            crypto:md5_update(Context, Data);
        false ->
            %% simply ignore the rest of the binary
            <<Chunk:ChunkSize/binary, _RestData/binary>> = Data,
            crypto:md5_update(Context, Chunk)
    end.
```

Changing the `md5_chunk_size` generator to something like

``` erlang
oneof([2, 4, 8, 16, 32, 64, 128, 256, 512, 1024]).
```

exhibits the bug (w/ the altered fun), and shows that the original implementation is still correct. I noticed this by running

``` erlang
eqc_gen:sample(eqc_gen:binary())
```

and noticing that the binaries produced by EQC were quite small.

reiddraper:
Pushed a small change to `bugfix/wrd-md5-chunked-eqc-enhancement` if you're interested.

kellymclaughlin:
Nice catch. I thought about that, but never followed up on it. I'll take a
look a your change and probably merge it in later.

kellymclaughlin:
I changed the test to call `eqc_gen:binary/1` and make it create 2048 byte binaries and also added 2048 to the option for chunk sizes in the generator.

reiddraper:
+1, nice find with `binary/1`

src/riak_cs_utils.erl
```diff
@@ -1484,3 +1487,30 @@ proxy_get_active() ->
 -spec pid_to_binary(pid()) -> binary().
 pid_to_binary(Pid) ->
     list_to_binary(pid_to_list(Pid)).
+
+%% @doc Rapid calls to `md5_update' with largish data blocks (e.g. 1MB)
+%% can lead to erlang scheduler collapse
+-spec chunked_md5(binary(), binary(), non_neg_integer()) -> {binary(), binary()}.
```
reiddraper, line 1493: The return value should just be `binary()`.

PR #496, opened 2013-03-16
Fix bug in riak_cs_acl_utils:process_owner

kellymclaughlin:
This either was overlooked during recent refactoring or was a mistake during conflict resolution on merge.


reiddraper:
+1

PR #498, opened 2013-03-18
Address some dialyzer warnings

kellymclaughlin:
Most of these popped up when we switched to the `1.3.1.1` tag of `riak-erlang-client`. More recent versions have better specs.

### DIFF
--- a/src/riak_cs_s3_policy.erl
+++ b/src/riak_cs_s3_policy.erl
@@ -185,7 +185,7 @@ policy_from_json(JSON) ->
                          E;
                       [] ->
                          {error, malformed_policy_missing};
-                      
+
                       Stmts ->
                          case {Version, ID} of
                              {undefined, <<"undefined">>} ->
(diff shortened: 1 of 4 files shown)


src/riak_cs_s3_policy.erl
```diff
@@ -561,7 +561,7 @@ int_to_prefix(0) -> 0.
 -spec statement_from_pairs(list(), #statement{})-> #statement{}.
 statement_from_pairs([], Stmt) ->
     case Stmt#statement.principal of
-        undefined ->
```
kellymclaughlin, line 564: @kuenishi I'd like to get your input on this. The change makes dialyzer happy, but I want to make sure it also makes sense to you. I do not see how the principal could ever be `undefined` since the default field value in `s3_api.hrl` is `[]`.

kuenishi, line 564: Nice! It was a bug. +1 to dialyze.

kuenishi:
Other diffs seems harmless to the code itself. +1 to merge.

PR #516, opened 2013-03-29
Proxy get tests for 1.4

bookshelfdave:
NOTE: core rtcs functions have changed, please validate all tests.

See also:
https://github.com/basho/riak_repl/pull/228

### DIFF
--- a/riak_test/src/repl_helpers.erl
+++ b/riak_test/src/repl_helpers.erl
@@ -188,6 +188,7 @@ start_and_wait_until_fullsync_complete(Node) ->
 
     lager:info("Fullsync on ~p complete", [Node]).
 
+
 wait_until_leader(Node) ->
     Res = rt:wait_until(Node,
         fun(_) ->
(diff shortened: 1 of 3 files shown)


reiddraper:
@metadave what version of Riak Enterprise should I be using to test this?

bookshelfdave:
@reiddraper riak_ee master

slfritchie:
Ping, says Ticket-about-a-month-old-bot

bookshelfdave:
Pong, we've been busy.

```
git clone riak_ee
rebar get-deps
cd deps/riak_repl
git checkout dip_pg14
cd ../.. (riak_ee root)
git checkout dip_pg_rebased
make && make devrel
cp ./riak_test/bin/rteedev-current.sh ./
./rteedev-current.sh
```

kellymclaughlin:
Pushed a commit to change the `rtcs` change to use the new `riak_test` config api.

riak_test/src/repl_helpers.erl
```diff
@@ -261,3 +262,155 @@ do_write(Node, Start, End, Bucket, W) ->
             lists:flatten([rt:systest_write(Node, S, S, Bucket, W) ||
                     {S, _Error} <- Errors])
     end.
+
+%% The functions below are for 1.3 repl (aka Advanced Mode MDC)
```
kellymclaughlin, line 171: I cannot find where or how these functions are used. Am I missing it or are they just here to enable tests to be written for advanced mode?

bookshelfdave, line 171: they are probably just carried over from https://github.com/basho/riak_test/blob/master/tests/repl_util.erl

kellymclaughlin:
The changes look ok to me and the tests mostly still pass and the failures that do occur do not seem related to these changes. Two outstanding questions before I +1:
1. Should the functions copied from repl_util.erl that are not being used be removed or would it be beneficial to leave them in to use in writing other tests?
2. Is it the case that the changes to the underlying replication system in use do not matter to the outcome of this test? That seems to be true from observation and the test results, but just want to make sure that is expected.

bookshelfdave:
1) extra functions removed from repl_util.erl.

2) The proxy_get API hasn't changed at all, so proxy_get will work with _default mode_ replication for 1.2 + 1.3, and _advanced mode_ replication in 1.4.

kellymclaughlin:
:+1: Merge away!

PR #512, opened 2013-03-28
Old object shows up after delete the latest

shino:
```
$ git show  --decorate | head -6
commit 41c81d4f10c59034ae1ba22ec5a4af1b303efb91 (HEAD, origin/develop, origin/HEAD, develop)
Merge: ff790c2 c404aa1
Author: Kelly McLaughlin <kelly@basho.com>
Date:   Thu Mar 21 09:46:21 2013 -0600

    Merge branch 'feature/configurable-pb-connect-timeout' into develop
```

First, put twice:

```
$ s3cmd la


$ echo "OLD" > data.local/1
$ s3cmd put data.local/1 s3://test
WARNING: Module python-magic is not available. Guessing MIME types based on file extensions.
data.local/1 -> s3://test/1  [1 of 1]
 4 of 4   100% in    0s   286.16 B/s  done
$ s3cmd get s3://test/1 -
s3://test/1 -> <stdout>  [1 of 1]
OLD
 4 of 4   100% in    0s   216.59 B/s  done
$ echo "NEW" > data.local/1
$ s3cmd put data.local/1 s3://test
WARNING: Module python-magic is not available. Guessing MIME types based on file extensions.
data.local/1 -> s3://test/1  [1 of 1]
 4 of 4   100% in    0s   230.60 B/s  done
$ s3cmd get s3://test/1 -
s3://test/1 -> <stdout>  [1 of 1]
NEW
 4 of 4   100% in    0s   236.07 B/s  done
```

After del it, old object shows up:

```
$ s3cmd del s3://test/1
File s3://test/1 deleted
$ s3cmd get s3://test/1 -
s3://test/1 -> <stdout>  [1 of 1]
OLD
 4 of 4   100% in    0s   413.61 B/s  done
```

One more del, object disappears:

```
$ s3cmd del s3://test/1
File s3://test/1 deleted
$ s3cmd la


```

PR #518, opened 2013-04-02
Fix delete resurrection #512

andrewjstone:
cs512 regression test

handle multiple active manifests during delete\n This fixes #512

slight refactoring and documentation

more documentation for the helpers of gc_active_manifests/3

fix typos and formatting

refactoring and code cleanup

more refactoring

### DIFF
--- a/src/riak_cs_gc.erl
+++ b/src/riak_cs_gc.erl
@@ -32,7 +32,7 @@
 -export([decode_and_merge_siblings/2,
          gc_interval/0,
          gc_retry_interval/0,
-         gc_active_manifests/5,
+         gc_active_manifests/3,
          gc_specific_manifests/5,
          epoch_start/0,
          leeway_seconds/0,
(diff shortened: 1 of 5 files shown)


kellymclaughlin:
[This](https://github.com/basho/riak_cs/blob/release/1.3/src/riak_cs_wm_object.erl#L240) is a pre-existing bug, but it is preventing me from running the client tests successfully. If you don't mind let's throw a fix for it in with this pr. The `lager` statement is just missing a formatting character so it is an easy fix.

kellymclaughlin:
Now I'm getting `500` errors on object deletes from the python client tests. Need to investigate more to see what's going on.

kellymclaughlin:
[This](https://github.com/basho/riak_cs/blob/release/1.3/src/riak_cs_wm_object.erl#L242) is also a bug. Even deletes for nonexistent objects should return a `204` so the tuple on that line should start with `true` instead of `false`. Hooray for exposing latent bugs!

kellymclaughlin:
Now I wonder if that lager message on [L240](https://github.com/basho/riak_cs/blob/release/1.3/src/riak_cs_wm_object.erl#L240) is going to be too spammy. Thoughts?

kellymclaughlin:
With the fix for the 2nd pre-existing bug previously mentioned patched: 
- [x] Client and integration tests pass
- [x] Dialyzer output is clean
- [x] All `riak_test` tests pass

src/riak_cs_gc.erl
```diff
@@ -43,30 +43,62 @@
...
-                                                           RiakcPid),
-                    return_active_uuids_from_gc_response(GCManiResponse,
-                                                         ActiveUUIDs);
-                updated ->
-                    updated
+%% @doc Keep requesting manifests until 
+%% 1) There are no more active manifests 
+%% 2) All previously active multipart manifests have had their unused parts cleaned
+%%    and become active+multipart_clean
+%% 3) All active manifests and active+multipart_clean manifests for multipart ar GC'd
```
kellymclaughlin: s/ar/are

andrewjstone:
I think the lager message for failures, especially delete failures, is handy. As long as logs are rotated, it shouldn't be a problem.

reiddraper:
Small style nit-pick, we tend to put newlines after any `->` occurrence, except lambdas. So even small functions like:

``` erlang
foo() -> ok.
```

would be written as:

``` erlang
foo() ->
    ok.
```

same for case clauses.

reiddraper:
Please add specs to any newly added functions too. Running dialyzer after adding them will keep you honest too.

kellymclaughlin:
Found an behavioral difference between CS and S3. To reproduce:
1. Start a multipart upload to CS to overwrite an existing object
2. While the multipart upload is progressing, delete the object at that key that the mp upload is writing
3. Allow the mp upload to complete.

With CS the mp uploaded object is present and valid, but with S3 accessing the key returns `404 Not Found`.

kellymclaughlin:
Also seeing cases where a `PUT` seems to succeed, but when I attempt to fetch it I get a `404 Not Found`. Here's the steps I have used to reproduce that behavior:
1. Start a multipart upload to a key that does not exist.
2. While the multipart upload is progressing, write a small object to the same key that the mp upload is writing
3. While the multipart upload continues, attempt to fetch the object at the key.

I see the following s3cmd results when doing this:

```
13:09:44:s3tmp $ s3cmd -c ~/.s3cfg-moss put a s3://simple-bucket/a
WARNING: Module python-magic is not available. Guessing MIME types based on file extensions.
a -> s3://simple-bucket/a  [1 of 1]
 110 of 110   100% in    0s   825.79 B/s  done
13:09:51:s3tmp $ s3cmd -c ~/.s3cfg-moss --force get s3://simple-bucket/a adown
s3://simple-bucket/a -> adown  [1 of 1]
ERROR: S3 error: 404 (Object Not Found):
```

Once the multipart upload completes the problem from my last comment takes effect and a fetch with s3cmd yields the following:

```
13:09:56:s3tmp $ s3cmd -c ~/.s3cfg-moss info s3://simple-bucket/a
s3://simple-bucket/a (object):
   File size: 1695547392
   Last mod:  Tue, 02 Apr 2013 19:09:30 GMT
   MIME type: application/octet-stream
   MD5 sum:   69d1cbdeb926414b8462c0a60c0a21ee
   policy: none
   ACL:       test1: FULL_CONTROL
```

andrewjstone:
@kellymclaughlin I think that first bug is due to not deleting writing manifests as well as active manifests during delete. Are you sure that behavior has changed since 1.3 release? Since it's an inconsistency, I agree it should be fixed, just wondering if it's directly related to my changes.

kellymclaughlin:
@andrewjstone Excellent point. Let me re-run those tests using the `release/1.3` branch and see what happens.

andrewjstone:
@kellymclaughlin I'm not able to reproduce the second error. The reason being, I think, is that the new smaller file becomes the active one since it has the later start time. When I run info it shows me the small file during and after mp upload of the large file.

kellymclaughlin:
I can reproduce the first error on `release/1.3` and I agree it is probably the `writing` state manifests issue. We can solve that separately. I am not able to reproduce the second error on `release/1.3`. There are times when I would not see the error so it could be some sort of race condition. Let me try to get more information about it.

kellymclaughlin:
Looks good, +1 to merge

src/riak_cs_gc.erl
```diff
@@ -43,30 +43,88 @@
...
+                  UUIDs :: [binary()]) ->
+      [binary()] | no_return().
+gc_manifest(M, RiakObject, Bucket, Key, RiakcPid, UUIDs) ->
+    UUID = M?MANIFEST.uuid,
+    check(gc_specific_manifests([UUID], RiakObject, Bucket, Key, RiakcPid), [UUID | UUIDs]).
+
+check({ok, _}, Val) -> 
+    Val;
+check({error, _}=Error, _Val) -> 
+    throw(Error).
```
reiddraper, line 127: why do we need to use exception throwing here? Feels like this and the caller `gc_manifests/5` could be pure code otherwise.

andrewjstone, line 127: It's a non-local return to escape the fold early in case of error in gc_manifests. The catch returns the error. Otherwise we'd have what I replaced, which is a fold with 2 clauses that must go through the list returning error in the accumulator each time. I wasn't sure which one I liked better, but this one has less code.

andrewjstone, line 127: I'm second guessing my change after talking with Reid. I think I'm going to change it to remove the exception.

andrewjstone, line 127: On the other hand, this works and has been tested. I still don't have client tests working. Merging as is. Will fix this in the near future.

reiddraper, line 127: Works for me.

PR #411, opened 2013-04-02
File overwrite upload can fail

reiddraper:
Andy reports that uploading a file and then overwriting the same file with an upload that takes > 60 seconds results in the file either disappearing or having 0 bytes.

Some relevant hipchat info:

```
i made some progress debugging it in japan, appears to be here:
https://github.com/basho/riak_cs/blob/master/src/riak_cs_put_fsm.erl#L202
that sometimes returns {ok, RiakObject}
i think it also happens here: https://github.com/basho/riak_cs/blob/master/src/riak_cs_put_fsm.erl#L256
i tried to trace it all the way through on the plane but i ran out of batteries
```


reiddraper:
derp, heya bug, that commit should not be attached to this issue

PR #520, opened 2013-04-02
Add Riak Test shell script error handling

reiddraper:
Removes this fun issue:

```
cd /dir/does/not/exist
git reset --hard HEAD
```

### DIFF
--- a/riak_test/bin/rtdev-build-releases.sh
+++ b/riak_test/bin/rtdev-build-releases.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# just bail out if things go south
+set -e
+
 # You need to use this script once to build a set of devrels for prior
 # releases of Riak (for mixed version / upgrade testing). You should
 # create a directory and then run this script from within that directory.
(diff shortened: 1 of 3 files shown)


shino:
I think it's good to merge as is.

An alternative way of error handling which may be also good is
to use `trap` and `ERR`. It can tell where an error occurs.

Sample script (error_trap.sh):

```
#!/bin/bash

trap 'echo "[ERROR] $0:$LINENO \"$BASH_COMMAND\"" ; exit 1' ERR

ls no_such_file
echo "This line does not appear"
```

Output:

```
$ ./error_trap.sh
ls: cannot access no_such_file: No such file or directory
[ERROR] /home/shino/tmp/error_trap.sh:5 "ls no_such_file"
```

shino:
Sorry. I pushed close button by mistake.

reiddraper:
> An alternative way of error handling which may be also good is
> to use trap and ERR. It can tell where an error occurs.

Definitely. I'm going to go ahead and merge this in, but we can revisit this with better error handling later if it continues to be an issue. Any objections?

shino:
No objection. I think your fix is sufficient :)
+1 to merge!

Issue #476, opened 2013-03-08
Use inet_parse:ipv4strict_address rather ipv4_address


shino:
Found in `riak_cs_s3_policy:parse_ip` .

```
4> inet_parse:ipv4_address("0").
{ok,{0,0,0,0}}
5> inet_parse:ipv4strict_address("0").
{error,einval}
```

andrewjstone:
Hi @shino. The change is easy to make, but I'm not sure we want to make it.
It causes an existing test to fail, leading me to believe that behavior may be expected elsewhere. 
https://github.com/basho/riak_cs/blob/develop/test/riak_cs_s3_policy_test.erl#L40

The behavior would change from returning {{0,0,0,0}, {0,0,0,0}} to crashing with a badmatch.

Are you sure this is a change we want to make?

shino:
@andrewjstone I can not determine whether the test case is appropriate or just a typo.

There are two choices, I think:
1. Try S3 policy behavior in brute-force way, and estimate its parsing logic and implement it.
2. Use strict.

@kota Do you know whether the IP range `192.3.1/16` in the test case is intentional or not?

kota:
@shino You are mentioning wrong person.

shino:
sorry for wong mention.

shino:
@andrewjstone I tried S3 bucket policy with "wrong" IP format and S3
responded with 500 Internal Error. So Riak CS's parse logic should be
strict and the test case should be corrected.

command:

```
s3cmd -d setpolicy scripts.local/s3/s3_policy_ip_range_put.txt s3://shino.shun.test-ip-range
```

policy json (`aws:SourceIp` has "wrong" format)

```
{
  "Version": "2008-10-17",
  "Id": "Policy1359983382391",
  "Statement": [
    {
      "Sid": "Stmt1359983375917",
      "Action": [
        "s3:PutObject"
      ],
      "Effect": "Allow",
      "Resource": "arn:aws:s3:::shino.shun.test-ip-range/*",
      "Condition": {
         "IpAddress" : {
                    "aws:SourceIp": "192.168/24" 
                }
      },
      "Principal": {
        "AWS": [
          "*"
        ]
      }
    }
  ]
}
```

andrewjstone:
@reiddraper @kellymclaughlin @slfritchie @shino 
[...]

PR #523, opened 2013-04-04
do not allow invalid addresses to succeed in parse_ip

andrewjstone:
I don't think this is ready to merge. There may be other corner cases here. Needs more testing as well. It's here just to allow a quick look at my approach.


andrewjstone:
This pull is towards a fix for #476

src/riak_cs_s3_policy.erl
```diff
@@ -744,8 +748,12 @@ condition_({<<"aws:SecureTransport">>, MaybeBool}) ->
     {ok, Bool} = parse_bool(MaybeBool),
     {'aws:SecureTransport', Bool};
 condition_({<<"aws:SourceIp">>, Bin}) when is_binary(Bin)->
-    IP = parse_ip(Bin),
-    {'aws:SourceIp', IP};
+    case parse_ip(Bin) of
+        {error, _} -> 
+            {error, malformed_policy_condition};
```
kuenishi: In case if you don't know yet, just to be sure: When you create a new error message, you must add response generating code to `riak_cs_s3_response.erl`.

andrewjstone: Thanks @kuenishi I was wondering if something like that existed. I'll add it in.

kuenishi: Also, to make this fail, errors should be thrown. I designed like that due to the deep call stack (something like maybe monad would be much better).

``` erlang
        {error, _} -> 
            throw({error, malformed_policy_condition});
```

After adding this code, in my local env it worked.

andrewjstone: I'm not so sure I agree with that. I had a conversation with @reiddraper last week about this type of thing and we decided that it was best to keep functions pure when possible unless there was a real measurable performance hit. The root caller (policy_from_json/1) actually does check for an error being returned from this, and the parse_bool function also returns an error from the same condition_/1 function.

reiddraper: Yeah assuming it's not prohibitive to return error tuples, I'd vote for that. I haven't looked at this code, but if it's a deep call stack that never previously accounted for error tuples being returned, exceptions may be the simplest hack for now. Whatever you guys think is best here.

andrewjstone:
Besides the minor disagreement on handling errors, I think this PR is good to go.

reiddraper:
@kuenishi you reviewing this?

kuenishi:
@reiddraper yeah, I was just commenting as an original writer (who wrote the bad code:) but now I'd love to.

kuenishi:
@andrewjstone Would be much better if a test hitting http interface added - but currently it's not easy to add policy interface to erlcloud, so I've been adding python boto tests for regression. See `BucketPolicyTest` in `client_tests/python/boto_test.py`.

andrewjstone:
@kuenishi I agree, it needs a test. I will add it in the morning. Thanks for the pointer.

-Andrew

kuenishi:
+1 to merge, thanks!
- riak_test passed (thanks, Kaz!)
- eunit including eqc passed
- python client test passed

andrewjstone:
Thanks for the review @kuenishi

PR #537, opened 2013-04-18
fold objects for list keys (proof-of-concept)

reiddraper:
## _DO NOT MERGE_

This PR is just to track work, and make it easy to share with others.
## TODO
- [x] Basic proof-of-concept
- [x] Feature parity w/ s3/current-list keys (prefix, delimiter, max-keys, etc.)
- [x] Delimiter key-skipping hack
- [x] Eliminate code-duplication between new and old listing
- [x] Switching of some sort based on capabilities for new and old code
- [x] Profiling and observability info
- [ ] 5-node cluster benchmarks
- [x] Check for memory leaks (at least a simple 'run this 1k times' and check for no memory growth)
- [x] Pure EQC tests
- [x] Share some of the old tests
- [x] Make sure riak_test passes
- [x] Add backward compatibility in riak-erlang-client 2i support branch


reiddraper:
> Switching of some sort based on capabilities for new and old code

Currently this is just an app.config, defaulting to false (old code). Until we can come up with something better.

reiddraper:
> Eliminate code-duplication between new and old listing

I managed to get rid of a lot more duplication than I originally thought, by changing `riak_cs_list_objects_fsm:clean_keys_and_manifests` to just return manifests (like in v2), instead of two-tuples of `{key(), manifest()}`.

reiddraper:
> Share some of the old tests

updated shared test in 22df88d462a0029af77157821800116de824bfc3

reiddraper:
Closing in favor of #600.

PR #538, opened 2013-04-19
create reproducible environments for python tests

reiddraper:
Currently the python tests require that you have `boto` globally installed. We should start using `virtualenv` with a `requirements.txt` file for `pip`. We should also consider having a separate `Makefile` in the dir.


reiddraper:
Closing in favor of #542 which has been rebased against the README changes from #539.

Issue #540, opened 2013-04-24
php tests failing


reiddraper:
I'm trying to review #539, and am having trouble running the PHP integration tests. Specifically, I'm getting this error:

``` bash
➜  php git:(a1766b7) ./phpunit

Parse error: parse error in /Users/reid/Documents/repos/basho/riak_cs/client_tests/php/tests/S3ClientTest.php on line 41

➜  php git:(a1766b7) php --version
PHP 5.3.15 with Suhosin-Patch (cli) (built: Aug 24 2012 17:45:44)
Copyright (c) 1997-2012 The PHP Group
Zend Engine v2.3.0, Copyright (c) 1998-2012 Zend Technologies
```

I _think_ I have everything set up properly, but could be missing something.

reiddraper:
cc @ksauzz

hectcastro:
I installed PHP54 to get this running:

``` bash
$ php --version
PHP 5.4.13 (cli) (built: Apr 19 2013 10:58:54)
Copyright (c) 1997-2013 The PHP Group
Zend Engine v2.4.0, Copyright (c) 1998-2013 Zend Technologies
$ composer --version
Composer version 1.0.0-alpha6
```

reiddraper:
Using 5.4.14 now:

```
PHPUnit 3.7.19 by Sebastian Bergmann.

Configuration read from /Users/reid/Documents/repos/basho/riak_cs/client_tests/php/phpunit.xml.dist

..[31;1mE[0m[31;1mE[0m.

Time: 0 seconds, Memory: 7.75Mb

There were 2 errors:

1) S3ClientTest::testPutGetDeleteObject
Invalid argument supplied for foreach()

/Users/reid/Documents/repos/basho/riak_cs/client_tests/php/tests/S3ClientTest.php:42

2) S3ClientTest::testBucketACL
Invalid argument supplied for foreach()

/Users/reid/Documents/repos/basho/riak_cs/client_tests/php/tests/S3ClientTest.php:42

[37;41m[2KFAILURES!
[0m[37;41m[2KTests: 5, Assertions: 10, Errors: 2.
[0m[2K
```

ksauzz:
I reproduced the issue. php-test can't run with aws-sdk-php-2.2 or later. I'll try to update php-tests.

reiddraper:
Another php README note, I had to use `php -d detect_unicode=Off` when installing composer. OSX 10.7. Though that was when I was using php 5.3.x.

[...]

PR #539, opened 2013-04-20
Added READMEs to each language client library test

hectcastro:
This pull request consists of the follow changes:
- Added the existing AWS PHP SDK test suite to the `Makefile`
- Added more consistent READMEs to each language's client library test suit

Despite the fact that most of these tests can be driven by the `Makefile`, I READMEs to each so that anyone who tries to learn how to use Riak CS from the tests has an easier time.

PR #541, opened 2013-04-25
php-tests: fixed to work with aws-sdk-php-2.3.x.

ksauzz:
php tests failed with aws-sdk-php-2.2.x or later. so I changed bad codes about list objects and fixed the sdk version in composer.json.

related issue #540.

In addition, I also added riak_test automation, so I'd like to add php-test to `make client-test` after merging #539.

### DIFF
--- a/client_tests/php/.gitignore
+++ b/client_tests/php/.gitignore
@@ -1,4 +1,3 @@
-test_services.json
 phpunit
 composer.lock
 vendor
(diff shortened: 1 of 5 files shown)


reiddraper:
Passes for me now. Do we need to keep `client_tests/php/test_services.json` in the repo still?

ksauzz:
yes, we do. `test_services.json` has default configurations except key_id, key_secret and proxy_uri. They are needed to run test.

reiddraper:
> yes, we do.

OK cool. +1 to merge!

Issue #533, opened 2013-04-11
Examine behavior dealing with writing state manifests and gc 


kellymclaughlin:
We need to verify and examine the behavior of `riak_cs` when dealing with writing state manifests when deletes or overwrites happen. There are several cases to examine and we need to verify that the behavior matches the behavior of S3. Each case is documented as a separate comment on this issue.

kellymclaughlin:
**Case 1**

Found an behavioral difference between CS and S3. To reproduce:
1. Start a multipart upload to CS to overwrite an existing object
2. While the multipart upload is progressing, delete the object at that key that the mp upload is writing
3. Allow the mp upload to complete.

With CS the mp uploaded object is present and valid, but with S3 accessing the key returns `404 Not Found`. 

Retrieving the new upload from CS results in this message from s3cmd
`WARNING: MD5 signatures do not match: computed=a20fece9f1ecc8222410be7919ddefb6, received="de0743ce29e84f7e9e07210078acbc90"` 

I think this is just for multipart, a bug in s3cmd

kellymclaughlin:
**Case 2**
1. Start a multipart upload to CS to overwrite an existing object
2. While the multipart upload is progressing, overwrite he object at that key that the mp upload is writing
3. Allow the mp upload to complete.

S3 (second overwrite finishes last due to larger file and later start): Shows the first overwrite while the second is uploading still. When second overwrite completes it is shown.
S3 (second overwrite finishes first due to smaller file, even though it started later): Shows the second overwrite when it finishes during upload of first and after first finishes.

Both during the final completing upload and after it finishes s3cmd returns a warning from CS
`WARNING: MD5 signatures do not match: computed=21608c9cef97d8baf5b493b72e895259, 
received="fa8552f380a3453f90f8c4c9429751ff"`

tldr: S3 uses write start time
CS: same behavior as S3

kellymclaughlin:
**Case 3**
1. Start a normal upload of a large object to CS to overwrite an existing object
2. While the upload is progressing, delete the object at that key that the upload is writing
3. Allow the upload to complete.

CS: Returns the object
S3: Returns `404 Not Found` during and after upload

kellymclaughlin:
**Case 4**
1. Start a normal upload of a large object to CS to overwrite an existing object
2. While the upload is progressing, overwrite the object at that key that the upload is writing
3. Allow the upload to complete.

[...]

PR #551, opened 2013-04-30
Feature/gc on read delete

andrewjstone:
If a delete occurs during an upload, when the upload finishes it should not be recognized as active. It will be garbage collected on the next read.

See #533


kellymclaughlin:
`test/.riak_cs_manifest_utils_eqc.erl.swp` snuck into your commit

kellymclaughlin:
- [x] Dialyzer output clean
- [x] `riak_test` tests pass

kellymclaughlin:
Got the following dialyzer output:

```
riak_cs_utils.erl:518: Expression produces a value of type {'error',_} | {'ok',{'riakc_obj','undefined' | binary(),'undefined' | binary(),'undefined' | binary(),'undefined' | [{_,_}],'undefined' | dict(),'undefined' | binary()}}, but this value is unmatched
```

src/riak_cs_gc.erl
```diff
@@ -128,13 +131,23 @@ gc_manifests(Manifests, RiakObject, Bucket, Key, RiakcPid) ->
...
     UUID = M?MANIFEST.uuid,
-    check(gc_specific_manifests([UUID], RiakObject, Bucket, Key, RiakcPid), [UUID | UUIDs]).
+    check(gc_specific_manifests_to_delete([UUID], RiakObject, Bucket, Key, RiakcPid), [UUID | UUIDs]).
 
 check({ok, _}, Val) -> 
     Val;
 check({error, _}=Error, _Val) -> 
     Error.
 
+-spec gc_specific_manifests_to_delete(UUIDsToMark :: [binary()],
```
kellymclaughlin, line 162: This seems to be the only spec that includes the parameter names. Do you think it is helpful to have that information included in the spec in this case?

andrewjstone, line 162: Nope, that doesn't need to be there. I'm going to remove it for consistency.

andrewjstone, line 162: Changed my mind. I misread the code, didn't see the rest. I discussed this with Reid when I started actually. When there are a lot of parameters, say more than 4, it is helpful so you can lineup the names with paramters. Alternatively I suppose we could define specific types for these things instead for clarity. I kinda prefer the latter actually, but the former can be helpful.

reiddraper, line 162: I'd be happy with either of those options.

kellymclaughlin, line 162: I was mostly curious if there was a reason for the difference and since there is it wfm.

src/riak_cs_manifest_utils.erl
```diff
@@ -307,18 +337,85 @@ orddict_values(OrdDict) ->
...
 
-%% NOTE: This is a foldl function, initial acc = no_active_manifest
-most_recent_active_manifest(Manifest=?MANIFEST{state=active}, no_active_manifest) ->
-    Manifest;
-most_recent_active_manifest(_Manfest, no_active_manifest) ->
-    no_active_manifest;
-most_recent_active_manifest(Man1=?MANIFEST{state=active}, Man2=?MANIFEST{state=active}) ->
+-spec delete_time(lfs_manifest()) -> erlang:timestamp() | undefined.
+delete_time(Manifest) ->
+    case proplists:is_defined(deleted, Manifest?MANIFEST.props) of
```
kellymclaughlin, line 385: `lists:keyfind` is probably a much better choice generally. This is probably not a huge deal right now because our list of props is small, but that may not always be true.

andrewjstone, line 385: I did not know that, thanks. I'll keep it in mind.

src/riak_cs_manifest_utils.erl
```diff
@@ -307,18 +337,85 @@ orddict_values(OrdDict) ->
...
+    Dt=delete_time(Manifest), 
+    {MostRecent, later(Dt, DeleteTime)};
+most_recent_active_manifest(Manifest=?MANIFEST{state=pending_delete}, {MostRecent, undefined}) -> 
+    {MostRecent, delete_time(Manifest)};
+most_recent_active_manifest(Manifest=?MANIFEST{state=pending_delete}, {MostRecent, DeleteTime}) -> 
+    Dt=delete_time(Manifest), 
+    {MostRecent, later(Dt, DeleteTime)};
+most_recent_active_manifest(Manifest=?MANIFEST{state=active}, {no_active_manifest, undefined}) ->
+    {Manifest, undefined};
+most_recent_active_manifest(Man1=?MANIFEST{state=active}, {Man2=?MANIFEST{state=active}, DeleteTime}) ->
```
kellymclaughlin: More or less noisy to make the case statement check a guard and use two function clauses instead?

andrewjstone: I like that better. Changing it now.

andrewjstone:
superceded by 560

PR #560, opened 2013-05-08
Feature/mark deleted with read check

andrewjstone:
When a manifest is deleted directly by a user and not overwritten to enter the pending delete_state, it is marked as `{deleted, true}` inside #MANIFEST.props. When an uploading manifest completes, the writing manifest will become the most recent active manifest.

When a user goes to read the key, the following occurs:
- If there is an existing manifest in the pending_delete or scheduled_delete state with {deleted, true} in props, and the delete_marked_time of that manifest is greater than the write_start_time of the most recent active manifest, then not_found is returned.
- If there is no existing manifest with {deleted, true} in props, then the most recent active manifest is returned.
- If there is an existing manifest with {deleted, true} in props, but the delete_marked_time is less than the write_start_time of the most recent active manifest, then the most recent active manifest is returned.

Supercedes #551


reiddraper:
After this is reviewed, I might suggest re-opening the PR as a single, squashed commit.

andrewjstone:
I can do that now if you want.

reiddraper:
I might also suggesting using the PR description as the commit message (or something suitably descriptive)

andrewjstone:
will do.

andrewjstone:
I Opened a new one with the squashed commit: #561

PR #561, opened 2013-05-08
Object deleted while updating behavior changes

andrewjstone:
When a manifest is deleted directly by a user and not overwritten to
enter the pending delete_state, it is marked as `{deleted, true}`
inside #MANIFEST.props. When an uploading manifest completes, the
writing manifest will become the most recent active manifest.

When a user goes to read the key, the following occurs:
- If there is an existing manifest in the pending_delete or
  scheduled_delete state with {deleted, true} in props, and the
  delete_marked_time of that manifest is greater than the
  write_start_time of the most recent active manifest, then not_found
  is returned.
- If there is no existing manifest with {deleted, true} in props,
  then the most recent active manifest is returned.
- If there is an existing manifest with {deleted, true} in props,
  but the delete_marked_time is less than the write_start_time of
  the most recent active manifest, then the most recent active
  manifest is returned.

Do not show active manifests on read time if a delete occurred after writing started #533

fix riak_cs_manifest_utils:most_recent_active_manifest/2

gc active manifests that were in writing state when a deletion occured
and have since completed. #533

return a list of UUIDs not Manifests from riak_cs_manifest_utils:find_deleted_active_manifests/2

quickcheck property for the active manifest

add type specs and fix a bug in the active manifest handling found by quickcheck

fix spec with hint from dialyzer

fix dialyzer error

remove accidental .swp checkin

use a guard instead of case statement and shorten some clauses

### DIFF
--- a/src/riak_cs_gc.erl
+++ b/src/riak_cs_gc.erl
@@ -39,6 +39,9 @@
          move_manifests_to_gc_bucket/3,
          timestamp/0]).
 
+%% export for repl debugging and testing
+-export([get_active_manifests/3]).
+
 %%%===================================================================
 %%% Public API
 %%%===================================================================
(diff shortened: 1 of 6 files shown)


andrewjstone:
Supercedes #533

kellymclaughlin:
- [x] Dialyzer output clean
- [x] `riak_test` tests pass

kellymclaughlin:
I see the expected behavior when the upload is a multipart upload, but not when it is a standard upload. Gonna dig into it a little more and see if I can figure out why.

andrewjstone:
Damn. I can't believe I didn't test that properly. Thx Kelly

kellymclaughlin:
Oops, disregard my last comment. Complete reviewer derpage. I forgot to reload the modules used for listing bucket contents on the riak side. Once I did that thing work just as expected. Sorry for the confusion.

+1 to merge

andrewjstone:
Awesome! I felt bad for giving you broken stuff.

PR #562, opened 2013-05-08
Bump lager version and add diagnostics

andrewjstone:
start of a diagnostics module for running systems

start of a table module and table of manifests

working tables and one for manifests

more useful manifest tables

bump lager to 2.0 and use a velvet tag that supports lager 2.0

pretty print manifests now

remove "dead" column from manifest table

### DIFF
--- a/include/riak_cs.hrl
+++ b/include/riak_cs.hrl
@@ -260,6 +260,12 @@
         %%       until sometime later.
         %% 'undefined' is for backward compatibility with v3 manifests
         %% written with Riak CS 1.2.2 or earlier.
+        %% Currently, the 'dead' property can be in props indicating
+        %% that this manifest was deleted while in the writing state.
+        %% The 'deleted' property may also exist on manifests in the
+        %% pending_delete and scheduled_delete states. This indicates
+        %% that the manifest was actually deleted by a user and not
+        %% due to an upload.
         props = [] :: 'undefined' | proplists:proplist(),
 
         %% cluster_id: A couple of uses, both short- and longer-term
--- a/include/riak_cs.hrl
+++ b/include/riak_cs.hrl
@@ -260,12 +260,6 @@
         %%       until sometime later.
         %% 'undefined' is for backward compatibility with v3 manifests
         %% written with Riak CS 1.2.2 or earlier.
-        %% Currently, the 'dead' property can be in props indicating
-        %% that this manifest was deleted while in the writing state.
-        %% The 'deleted' property may also exist on manifests in the
-        %% pending_delete and scheduled_delete states. This indicates
-        %% that the manifest was actually deleted by a user and not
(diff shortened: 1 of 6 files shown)


include/riak_cs.hrl
```diff
@@ -260,6 +260,12 @@
         %%       until sometime later.
         %% 'undefined' is for backward compatibility with v3 manifests
         %% written with Riak CS 1.2.2 or earlier.
+        %% Currently, the 'dead' property can be in props indicating
+        %% that this manifest was deleted while in the writing state.
+        %% The 'deleted' property may also exist on manifests in the
+        %% pending_delete and scheduled_delete states. This indicates
+        %% that the manifest was actually deleted by a user and not
+        %% due to an upload.
```
reiddraper: Is this related to this PR? No big deal if it's not, just making sure I understand.

reiddraper:
- [x] `riak_cs_diags` successfully gets restarted after `exit(whereis(riak_cs_diags), kill).`.
- [x] eunit/eqc
- [x] client test
- [x] xref
- [x] dialyzer

reiddraper:
There's a couple of small xref warnings that'd be nice to fix. Probably just moved line numbers. I'd say just remove the line numbers altogether too.

reiddraper:
Some dialyzer issues too, but maybe that's better addressed as a separate PR to get us back on track.

reiddraper:
Sorry for leaving a bunch of small comments... A few of the dialyzer issues are `riak_cs_diags` related, they should be fixed in this PR.

andrewjstone:
No worries. I appreciate the review. Will answer soon :)

include/riak_cs.hrl
andrewjstone: That comment is outdated. I'll remove it.

andrewjstone:
I fixed a few simple warnings. I know the following are fixed elsewhere: 

```
riak_cs_s3_policy.erl:499: Function will never be called
riak_cs_s3_policy.erl:503: Function eval_all_ip_addr/2 will never be called
riak_cs_s3_policy.erl:510: Function ipv4_band/2 will never be called
riak_cs_s3_policy.erl:513: Function ipv4_eq/2 will never be called
riak_cs_s3_policy.erl:816: The contract riak_cs_s3_policy:parse_tokenized_ip([string()]) -> {inet:ip_address(),inet:ip_address()} cannot be right because the inferred return for parse_tokenized_ip([nonempty_string()]) on line 808 is {nonempty_string(),{non_neg_integer(),non_neg_integer(),char(),byte()}}
```

It looks like lager 2.0 introduced some that should probably be fixed in lager. I'm not sure about the others, but they aren't from this code as far as I can tell.

reiddraper:
+1 to merge

PR #563, opened 2013-05-10
Updated object metadata can be lost during manifest resolution

kellymclaughlin:
A copy operation of an object to itself to update the object's
metadata can result in the updated metadata being lost during manifest
conflict resolution. In the case that there are two manifests in the
active state the ACL creation time is used as conflict resolution
criteria. Currently, the most recent ACL and the resolved manifest
properties are set in the manifest record that is given as the first
parameter to the resolution function and other fields of the manifest
record are ignored. Therefore, if the most recent active manifest is
given as the second parameter and includes an updated value for its
manifest metadata field, that updated metadata is lost. This change
makes sure all fields of the active manifest with the most recent ACL
are preserved.


andrewjstone:
- [x] Dialyzer output clean
- [x] `riak_test` tests pass

+1 to merge

PR #569, opened 2013-05-23
Better resolution handling for active manifests with identical ACLs

kellymclaughlin:
Better confliect resolution handling for case where two active state
manifests that have identical ACLs. Prior to this change this situation
could result in an object becoming unavailable.


andrewjstone:
- [x] eunit and eqc tests
- [x] manual test verifcation of fix

+1 to merge

andrewjstone:
I gave this a +1 since the tests you told me to run worked @kellymclaughlin. However, I just reverted to 1.2 and running the same test with the dupe manifest already in the object in riak still works fine. I'm a bit confused. We can discuss later.

PR #573, opened 2013-05-30
Minor dialyzer fixes

reiddraper:
- Remove some hardcoded line numbers
- Pattern match a lager return val


slfritchie:
The dialyzer.warnings file can take regexps, so in theory you deleted a bit too much ... but I'm OK with that. +1, thanks!

reiddraper:
This isn't going to merge cleanly, and on second thought, some of the repeated lines of just `Callback info about the riak_cs_auth behaviour is not available` is kind of ugly. Going to make another one off of current develop that is a little cleaner.

PR #575, opened 2013-05-31
Fixed default StorageClass from 'regular' to 'standard' per API spec

glickbot:
The API spec defines the possible options of StorageClass:

> StorageClass  
> The class of storage (STANDARD or REDUCED_REDUDANCY) that will be used to store the object when the multipart upload is complete.

This pull changes the default from REGULAR to STANDARD.

### DIFF
--- a/include/riak_cs.hrl
+++ b/include/riak_cs.hrl
@@ -354,7 +354,7 @@
     owner_key_id :: string(),
 
     %% storage class: no real options here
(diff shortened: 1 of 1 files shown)


reiddraper:
Should we also make a change to rewrite/display old values written with `regular`? cc/ @kellymclaughlin

reiddraper:
Closing in favor of #588.

PR #577, opened 2013-05-31
Integrate the ceph s3-tests suite into automated client tests

kellymclaughlin:
Currently this only attempts to run the functional test suite and there are failures. Many of these failures are due to error response bodies not matching the API specs. Fixing those failures is dependent on [this](https://github.com/basho/webmachine/pull/144) webmachine pr and will be addressed in a separate set of changes.

Running the tests depends on [this](https://github.com/basho/s3-tests/pull/1) pr to the basho fork of the `s3-tests` repo that adds support for specifying a proxy host and port.


reiddraper:
Works as advertised. I haven't looked at the WM PR yet though. My only concern is that since this integrates the test with our `test-client` target, through the `test-boto` target. This means the client-tests fail, which is a little bit of a false-alarm.

reiddraper:
Since the python tests are now run from a Makefile in client_tests/python, the main Makefile should just call this 'python', not 'test-boto'.

kellymclaughlin:
Updated `test-python` Makefile target in [73c2e7e](https://github.com/basho/riak_cs/commit/73c2e7e7e16072a7d71f0e1969129b4e610aa7a7)

reiddraper:
Update looks good.

reiddraper:
+1 with #578 merged in first.

reiddraper:
+1

PR #586, opened 2013-06-07
Stop all application nodes after riak_test run

kellymclaughlin:
Updates so that the `riak_test` modules use the `rt_cs_dev` test harness instead
of the `rtdev` Riak harness. This has changes to properly stop all of the CS and
Stanchion nodes. It also requires some configuration changes. See the updates
to the `README` in the `riak_test` subdirectory for more details.

This PR relies on [this](https://github.com/basho/riak_test/pull/297) PR introducing the RiakCS test harness for `riak_test`.

### DIFF
--- a/riak_test/README.md
+++ b/riak_test/README.md
@@ -12,45 +12,52 @@
 1. Setup a `~/.riak_test.config` file like this:
 
     <pre>
+    {default, [
+           {rt_max_wait_time, 180000},
+           {rt_retry_delay, 1000}
+          ]}.
+
     {rtdev, [
-        {rt_deps, [
-                   "/Users/dparfitt/test-releases/riak_ee/deps",
-                   "/Users/dparfitt/test-releases/riak_cs/deps",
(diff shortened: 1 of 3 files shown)


riak_test/README.md
```diff
@@ -80,5 +87,5 @@ commands `make clean-client-test` and then `make compile-client-test`.
 1. To execute a test, run the following from the `riak_test` repo:
 
     ```
-    ./riak_test -c rtdev -d <PATH-TO-RIAK-CS-REPO>/riak_test/ebin/ -v
+    ./riak_test -c rt_cs_dev -d <PATH-TO-RIAK-CS-REPO>/riak_test/ebin/ -v
```
reiddraper, line 90: Are you able to run a single test with `-t`? If you provide both `-d` and `-t`, it seems like it just runs the whole directory. Maybe there's a way to add the Riak CS `riak_test/ebin` to the path in the `~/.riak_test.config`?

reiddraper:
+1

Issue #581, opened 2013-06-04
Support CORS [JIRA: RCS-177]


reiddraper:
Add support for [CORS](http://en.wikipedia.org/wiki/Cross-origin_resource_sharing)

[S3 CORS docs](http://docs.aws.amazon.com/AmazonS3/latest/dev/cors.html).

[W3C CORS doc](http://www.w3.org/TR/cors/)

xiaoliuliu2050:
i just encountered the same problem ,how to solve it?

shino:
CORS is on the roadmap of Riak CS but not implemented yet and not determined when.

xiaoliuliu2050:
whether could i use the aws s3 interface  named 'putBucketCors ' to add a cors configuration to support  
it ?

shino:
Riak CS does NOT implement the S3 API. I'm afraid you could not.

xiaoliuliu2050:
http://docs.basho.com/riakcs/latest/references/apis/storage/s3/
this chapter describes that riak cs support some of s3 api,
including PUT Object interface ,
but when i use it ,it just encountered the cors problme

xiaoliuliu2050:
hello shino ,do you know some other ways to solve this problem?
i have used the putobject interface  riakcs supported,
and the the putbucket interface riakcs not supported.
which do not work any more .

kuenishi:
I still don't understand exact problem. Lacking in CORS support may not interfere correct PUT Object call, but we need more information about WHAT and HOW you are failing in detail, to know whether we can solve your problem or not.

whw987:
Hi, Mr kuenishi

I and xiaoliuliu2050 work in a same development group, currently we encounter a problem about how to put and get files in browser environment to the riak-cs system, our riak cs is version 1.5.4. 
The following is our questions:
1. If lacking in CORS support in current riak-cs, can we make operation put and get files in browser environment to the riak-cs system (we adopt Amazon S3 javascript REST API in aws-sdk-2.1.4.min.js)
2. Based on question 1, if we can, then how to or what API we can use?

Thank you in advance.

kuenishi:
[...]

PR #587, opened 2013-06-10
Can't do multi-part uploads with query string authentication

kellymclaughlin:
When multi-part uploads and query string authentication are used together, Riak CS raises several different exceptions.

I want to upload some files with an AJAX client. To this end, I need to apply the multi-part upload and the query string authentication. However, I met some problems.

The first issue is the lack of CORS support. I fixed that by setting up nginx to reply to `OPTIONS` requests with the correct headers.

After that, when uploading with the `PUT` method, Riak CS replies with a 405 (Method not allowed) HTTP code with no additional message.
If I use the `POST` method, I get a 403 (Forbidden) HTTP code and "Access Denied" in the XML response.

I tried with both URL formats:
- /bucket/key?Signature=X&Expires=X&AWSAccessKeyId=X&uploadId=X&partNumber=X
- /bucket/key/uploads/uploadId?Signature=X&Expires=X&AWSAccessKeyId=X&partNumber=X

My script works well on Amazon S3 (with the first URL format) but fails on Riak CS 1.3.1 on Debian 64-bits.

### DIFF
--- a/src/riak_cs_s3_auth.erl
+++ b/src/riak_cs_s3_auth.erl
@@ -126,6 +126,7 @@ calculate_signature(KeyData, RD) ->
            Date,
            AmazonHeaders,
            Resource],
+    lager:debug("STS: ~p", [STS]),
     base64:encode_to_string(
       crypto:sha_mac(KeyData, STS)).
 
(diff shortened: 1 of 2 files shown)


reiddraper:
> The first issue is the lack of CORS support. I fixed that by setting up nginx to reply to OPTIONS requests with the correct headers.

I've created #581 to track CORS support work.

kellymclaughlin:
I have reproduced the authentication issue. I am working on a fix now and should have a pull request ready very shortly. Thanks for the report!

kellymclaughlin:
Need to write a `riak_test` or two for pre-signed request processing, but that will also require some `erlcloud` hacking so going to punt on it for the time being.

andrewjstone:
- [x] eunit & eqc tests
- [x] riak_test tests
- [x] dialyzer clean

+1 to merge

PR #585, opened 2013-06-12
Downloads of multi-part uploaded files fail with s3cmd and the Java client

kellymclaughlin:
s3cmd and the Amazon AWS Java client (and potentially others) require that the ETag returned  for multi-part uploaded files contain a "-" character.  This requirement is not captured in the API spec, and appears to be a quirk implemented in these clients.  The issue was discovered as part of CloudStack development and testing (see http://mail-archives.apache.org/mod_mbox/cloudstack-dev/201306.mbox/%3C77B337AF224FD84CBF8401947098DD87038FCD%40SJCPEX01CL01.citrite.net%3E).

Thanks to Edison Su (@edisonsu) and Min Chen for running down the issue.

### DIFF
--- a/client_tests/clojure/clj-s3/test/java_s3_tests/test/client.clj
+++ b/client_tests/clojure/clj-s3/test/java_s3_tests/test/client.clj
@@ -76,6 +76,26 @@
 (defn etag-suffix [etag]
   (subs etag (- (count etag) 2)))
 
+(defn create-manager [c]
+  (TransferManager. c))
+
+(defn configure-manager [tm]
+  (let [tm-config (.getConfiguration tm)]
+    (.setMultipartUploadThreshold tm-config 19)
+    (.setMinimumUploadPartSize tm-config 10)
+    (.setConfiguration tm tm-config)))
+
+(defn create-and-configure-manager [c]
+  (let [tm (create-manager c)]
+    (configure-manager tm)
+    tm))
+
+(defn upload-file [tm bucket-name object-name file-name]
+  (let [f (clojure.java.io/file file-name)
+        u (.upload tm bucket-name object-name f)]
+    (.waitForCompletion u)
+    (.delete f)))
+
 (fact "bogus creds raises an exception"
       (let [bogus-client
             (s3/client "foo"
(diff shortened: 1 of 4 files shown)


client_tests/clojure/clj-s3/test/java_s3_tests/test/client.clj
```diff
@@ -125,3 +135,28 @@
...
                  c bucket-name object-name))))
         => md5-sum))
+
+(let [bucket-name (random-string)
+      object-name (random-string)
+      value (str "aaaaaaaaaa" "bbbbbbbbbb")
+      upload-file "./clj-mp-test.txt"]
+  (fact "mulitpart upload works"
+        (with-random-client c
+          (do
```
reiddraper, line 165: there's an implicit `do` with this macro (since it expands to a `let` block)

client_tests/clojure/clj-s3/test/java_s3_tests/test/client.clj
```diff
@@ -125,3 +135,28 @@
...
+            (s3/create-bucket c bucket-name)
+            (def tm (TransferManager. c))
```
reiddraper: I'd use nested `let` blocks instead of `def` here

client_tests/clojure/clj-s3/test/java_s3_tests/test/client.clj
```diff
@@ -125,3 +135,28 @@
...
+            (def tm-config (.getConfiguration tm))
+            (.setMinimumUploadPartSize tm-config 10)
+            (.setMultipartUploadThreshold tm-config 19)
+            (.setConfiguration tm tm-config)
```
reiddraper: check out [`doto`](http://clojuredocs.org/clojure_core/clojure.core/doto) for doing multiple stateful/side-effecting things to a single java object.

client_tests/clojure/clj-s3/test/java_s3_tests/test/client.clj
reiddraper, line 165: actually, I'm wrong about this. hmm. I think there's a bug in my macro so that it doesn't support multiple forms

reiddraper:
It appears this change doesn't affect existing uploads. Is that intended?

reiddraper:
Also seems like s3cmd fails either way. Hmmm.

jburwell:
I thought the information provided in the cloudstack-dev thread was sufficient to reproduce the defect.  I am following up with Edison and Min to get more details and stack trace information.

reiddraper:
Seems to just be a bug in s3cmd, it also fails with real s3.

reiddraper:
> It appears this change doesn't affect existing uploads. Is that intended?

We've decided to punt on this for now.

reiddraper:
+1

PR #591, opened 2013-06-12
Gh555 fs2 stability

slfritchie:
Initial review request.  I'm going to be traveling & partially out-of-the-office for the next week, so having another set of eyes look at this would be A Good Thing.

I haven't Dialyzed anything yet, sorry, so any bugs there are definitely fair game.  The riak_cs_fs2_backend.erl analysis works better in the riak_kv repo.


src/riak_cs_utils.erl
```diff
@@ -109,20 +111,13 @@ anonymous_user_creation() ->
...
-              Hex = erlang:integer_to_list(X, 16),
-              if
-                  X < 16 ->
-                      lists:flatten(["0" | Hex]);
-                  true ->
-                      Hex
-              end
-          end || X <- binary_to_list(Bin)],
-    string:to_lower(lists:flatten(XBin)).
+binary_to_hexlist(B) ->
```
kellymclaughlin, line 114: I don't understand this. I can see not wanting to make the fs2 module dependent on `cs_utils`, but I don't think reversing that dependency is a good idea either. The fs2 module might someday live in `riak_kv` so I think in this case it's better to just have some code duplication.

kellymclaughlin:
I noticed when reading through there were some significant differences from latest `develop`. Pushed [feature/fs2-backend-stability](https://github.com/basho/riak_cs/tree/feature/fs2-backend-stability) that has been rebased. Any issues if we close this PR and continue discussion there? cc/ @reiddraper

reiddraper:
>  Pushed feature/fs2-backend-stability that has been rebased. Any issues if we close this PR and continue discussion there?

wfm

kellymclaughlin:
I'm gonna leave it for @slfritchie to open the new pr. It's just strange if I open the PR because then it seems like I'm making comments to myself. :)

src/riak_cs_utils.erl
```diff
@@ -1203,8 +1214,8 @@ generate_access_creds(UserId) ->
 -spec generate_canonical_id(string(), string()) -> string().
 generate_canonical_id(KeyID, Secret) ->
     Bytes = 16,
-    Id1 = crypto:md5(KeyID),
-    Id2 = crypto:md5(Secret),
+    Id1 = md5(KeyID),
```
kellymclaughlin, line 1217: The input parameters to this function are both type `string` and `crypto:md5` accepts `iolist` or `binary`, but the local `md5` only accepts `binary`. This causes the following test failure:

```
riak_cs_utils_test: bucket_resolution_test...*failed*
in function riak_cs_utils:md5_update/2 (src/riak_cs_utils.erl, line 476)
  called as md5_update(<<1,35,69,103,137,171,205,239,254,220,186,152,118,84,...>>,"festersquest")
in call from riak_cs_utils:md5/1 (src/riak_cs_utils.erl, line 467)
in call from riak_cs_utils:generate_canonical_id/2 (src/riak_cs_utils.erl, line 1181)
in call from riak_cs_utils:user_record/5 (src/riak_cs_utils.erl, line 1391)
in call from riak_cs_utils_test:bucket_resolution_test/0 (test/riak_cs_utils_test.erl, line 246)
**error:function_clause
```

kellymclaughlin:
I pushed a few commits to `feature/fs-backend-stability` to allow `riak_cs_utils:md5` to accept a string in addition to binary and to account for some functions that have been re-homed to `riak_cs_config`. I can now get the client tests running. Many of them pass, but the tests dealing with multipart upload are failing. Still have to investigate that further.

kellymclaughlin:
The problem with multipart uploads was that the `block_size` i set for fs2 in the riak app.config was exactly 1MB and it actually needs to be slightly larger. I used `1050000` and that seemed sufficiently large and got the client test suites all passing.

kellymclaughlin:
Starting Riak after changing the fs2 `block_size` when there is still data written with another size makes Riak unhappy due to the backend refusing to start. It would probably be good to try and log a meaningful error message in this case because currently there is no information to indicate what the problem might be.

kellymclaughlin:
I'm concerned about the fixed block size issue. There seems to be a number of problems with the current scheme. One I mentioned in my earlier comments about having `block_size` too small on the Riak side of things and it failing with a non-obvious error. This scheme would also preclude us from having a variable block size on the CS side. Then if you do need or want to change the block side the portion of the data set that has been already written has to be thrown away.

It seems advantageous to be more flexible and it could lead to less head scratching by users (or me). I know Reid proposed a way to use varying block sizes that also avoids having to restrict the siblings within the backend. 

A middle of the road option I thought about was forgoing the `block_size` specification on the Riak side and having the block size be set per file. The first entry in an `fs2` file would be the block size followed by the data blocks. The blocks in each file are the same size so offset calculation is still trivial. This should be safe since a CS object uses the same block size for all of its data and each `fs2` file only contains data from a single CS object. It would still require siblings to be restricted, but would avoid the headaches and hassles associated with a single, fixed block size. Thoughts?

src/riak_cs_get_fsm.erl
```diff
@@ -267,6 +267,17 @@ waiting_chunks({chunk, Pid, {NextBlock, BlockReturnValue}},
...
                      }=State) ->
     _ = lager:debug("Retrieved block ~p", [NextBlock]),
+    case BlockReturnValue of
+        {error, _} = ErrorRes ->
+            #state{bucket=Bucket, key=Key} = State,
+            lager:error("~p: Cannot get S3 ~p ~p block# ~p: ~p\n",
+                        [?MODULE, Bucket, Key, NextBlock, ErrorRes]),
+            %% Our terminate() will explicitly stop dependent processes,
+            %% we don't need an abnormal exit to kill them for us.
+            exit(normal);
```
reiddraper, line 277: why `exit/1` here instead of returning `{stop, ...}` from this function?

src/riak_kv_fs2_backend.erl
```diff
@@ -18,21 +16,146 @@
 %% specific language governing permissions and limitations
 %% under the License.
 %%
-%% -------------------------------------------------------------------
+%% ---------------------------------------------------------------------
 
-% @doc riak_kv_fs2_backend is filesystem storage system, Mark III
```
reiddraper, line 23: I kind of dug the `Mark III` comment ;)

slfritchie:
Re: block size.  The requirement for block size + extra for headers is a documentation detail, or a code detail to automatically add the header fudge factor beyond what the config file says.  I'm flexible.

For the non-obvious error ... the backend is returning an error, so the vnode returns an error, so the put FSM returns an error, and the CS client side is assuming success always, IIRC.  Which is a bug.

As for the variable block sizes, the CS client side yet doesn't add the non-default block size metadata header.  And the write path in the fs2 code doesn't support it yet.  But the read path of fs2 does, e.g., https://github.com/basho/riak_cs/blob/gh555-fs2-stability/src/riak_kv_fs2_backend.erl#L868

```
BlockSize = read_blocksize_from_file_header(FH, State),
```

The block size found there is the fs2 instance's default, but it can indeed be changed on a per-object basis.

reiddraper:
> Re: block size. The requirement for block size + extra for headers is a documentation detail, or a code detail to automatically add the header fudge factor beyond what the config file says. I'm flexible.

Between these two options, I think I like the code detail to automatically add the header fudge factor.

> For the non-obvious error ... the backend is returning an error, so the vnode returns an error, so the put FSM returns an error, and the CS client side is assuming success always, IIRC. Which is a bug.

Yeah, it's probably time to make the Riak CS GET and PUT FSMs more resilient to this, or at the very least, not just 'let it crash'.

reiddraper:
Another 'global' that kind of scares me is the `FS2_CONTIGUOUS_BLOCKS`. In the sense that if it's misconfigured on one node, Bad Things will happen. Unfortunately I don't have a great idea on how to fix this. Maybe (ducks), store it in the ring...?

reiddraper:
I've unfortunately not been able to get dialyzer to work at all in the kv repo. [Here's](https://gist.github.com/reiddraper/1ea404246d1d1b3b5837) the error I get. Note, this is before even putting the fs2 file in the repo.

justinsheehy:
This is just a note to capture an area of curiosity about fs2 generally. Please correct any misunderstandings that may have led to it.

If I understand correctly, one of the primary motivations (though not the sole motivation) for switching from bitcask to fs2 for Riak CS block storage is the scenario where bitcask would consume an unacceptable amount of RAM. This scenario occurs when, roughly, users wish to store more than tens of millions of objects per GB of available RAM. Generally, this will mean that users are storing very small objects.

If fs2 will store a separate file, even a sparse one, for the content of each separate riak object, then this means that even for a very small object (less than 1KB) we will consume at least one filesystem block, typically 4KB.

(This would be terribly worse -- by multiple orders of magnitude -- on a filesystem that did not efficiently handle sparse files, but I will assume for now that we are not concerned with such deployments)

One recurring concern among customers, prospects, and partners is that we multiply the amount of disk needed to store their data by 3x per cluster. Unless I am mistaken, the use of fs2 means that in exactly the scenario where people are scared of bitcask (huge number of tiny files) we will be trading away one failure (running out of RAM) but gaining another: an additional multiplier on the amount of disk space used.

It may be that this does not matter. For instance, the concrete magnitude of disk usage in a "tiny files" scenario might be low enough that multiplying disk usage is far preferable to a RAM constraint. Even if so, we should be aware of this consequence to that we can educate our users accordingly.

kellymclaughlin:
> If fs2 will store a separate file, even a sparse one, for the content of each separate riak object, then this means that even for a very small object (less than 1KB) we will consume at least one filesystem block, typically 4KB.

I am still a relative zfs noob so please correct me if I am wrong here, but from my reading it seems like zfs uses the `recordsize` as a maximum block size, but will dynamically size blocks for data smaller than the `recordsize` and might avoid wasting space in the manner Justin mentioned. Can anyone confirm that? I'm planning to do some testing to try and verify this for myself, but if that _is_ the case it's another compelling reason to perhaps push people in the zfs direction when using this backend.

src/riak_cs_utils.erl
slfritchie, line 114: If there's consensus on "just duplicate the damn code", I'm fine, I guess.

slfritchie:
Sorry, I'm behind the times.  I'm supposed to open a new PR?  Yet there are comments here following the rebasing of the branch.  Hm.

slfritchie:
Closing in favor of #599.

PR #592, opened 2013-06-13
Fix regression in generate_etag function in riak_cs_wm_object

kellymclaughlin:
This regression snuck in as part of e827ff1ce415b51d4aa374c16c3c00fa381833a1. 
The etag returned from generate_etag should not be quoted. This caused the 
object_get_conditional_test riak_test module to fail.

### DIFF
--- a/src/riak_cs_utils.erl
+++ b/src/riak_cs_utils.erl
@@ -132,7 +132,9 @@ etag_from_binary(Binary, Suffix) ->
 
 %% @doc Return a hexadecimal string of `Binary', without double quotes
 %% around it.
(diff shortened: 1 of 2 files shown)


kellymclaughlin:
[Here](https://github.com/basho/riak_cs/commit/b942114ce8460c1b93d734931b2c1289f9ba27fb#L3L133) is where I erroneously changed the code in e827ff1ce415b51d4aa374c16c3c00fa381833a1.

reiddraper:
+1. r_t fails before this patch, passes after.

PR #593, opened 2013-06-14
Fix GET FSM race conditions

reiddraper:
- Fix two bugs in the GET FSM
  - introduced in b14d86d9a5ca8ed7bddb40ed0e82d13b63d925af
- Fix one bug in the accompanying test
  - introduced in fec985e156fc308f40c682e9fc88ab3aec423575

The previous code would mistakenly stop the GET FSM when the
`blocks_intransit` queue was empty. Small values of the get buffer would
cause this to happen even more often. The GET FSM should only stop once
all blocks have been sent. This change introduces a new record field
called `num_sent` to keep track of the number of blocks returned to the
caller. We now also keep track of the `From' param if we aren't able to
send any bytes back to the user in case of a`get_next_chunk` call.

The eunit test had a bug where it would expect Content-Length (10000
hardcoded) chunks from the GET FSM, instead of:

Content-Length / BlockSize

This was masked by the fact that one the `foldl` got a done message, it
would happily loop a few extra times. And since Content-Length was
always larger than the number of blocks, we never say anything.


reiddraper:
Pulse tests incoming...

reiddraper:
Now also see the commit message in bdd38af54b420b1495db85ec77c9505f6d3f6d3c.

kellymclaughlin:
- [x] Pulse tests
- [x] Dialyzer output clean
- [x] `riak_test` tests pass

test/riak_cs_get_fsm_eqc.erl
```diff
@@ -114,7 +114,13 @@ prop_get_fsm() ->
...
 
 start_fsm(ContentLength, BlockSize) ->
-    {ok, FSMPid} = riak_cs_get_fsm:test_link(<<"bucket">>, <<"key">>, ContentLength, BlockSize),
+    %% these should probably turn into inputs for the EQC test
+    FetchConcurrency = 2,
+    BufferFactor = 32,
+    {ok, FSMPid} = riak_cs_get_fsm:test_link(<<"bucket">>, <<"key">>,
+                                             ContentLength, BlockSize,
+                                             FetchConcurrency,
+                                            BufferFactor),
```
kellymclaughlin: Indent looks off by one space

kellymclaughlin:
Dialyzer output is not _exactly_ clean. There are 6 warnings, but the same 6 are also present on `develop`.

kellymclaughlin:
:+1:

PR #594, opened 2013-06-20
Support build_type and backend configuration options

kellymclaughlin:
Allow riak_test suites to be run using different backend configurations and
build types.  The backend options are using the multi_backend in the standard
RiakCS setup or the memory backend.  The build type options are oss or ee and
indicates whether an open-source or enterprise version of Riak should be used
for the test run.


reiddraper:
+1

PR #597, opened 2013-06-21
Update `clj-aws-s3` to 0.4.2

reiddraper:
This uses the Java AWS SJD 1.4.7 which eliminates some of the connection
reset errors we were seeing.


kellymclaughlin:
:+1:

Issue #555, opened 2013-05-02
fs2 backend stability [JIRA: RCS-166]


slfritchie:
General goal: get the fs2 backend into a stable & usable state.  Bonus points for highest possible performance, but that will probably require talking directly to `prim_file` and so is probably best left as a separate work item.

shino:
Close. Next step :)

Basho-JIRA:
Ummm, this GH issue was closed on March 30, 2015. Closing in JIRA. 

**_[posted via JIRA by Scott Fritchie]**_

PR #599, opened 2013-06-25
Feature/fs2 backend stability (GH 555)

slfritchie:
See #555 and the soon-to-be-closed #591 for prior discussion.


slfritchie:
Note: PR #591 was closed after Kelly's rebasing of the feature branch.  All discussion & review should be happening here.

src/riak_cs_block_server.erl
```diff
@@ -234,6 +208,104 @@ handle_cast({delete_block, ReplyPid, Bucket, Key, UUID, BlockNumber}, State=#sta
...
+    {noreply, State};
+do_get_block(ReplyPid, Bucket, Key, ClusterID, UUID, BlockNumber,
+             State=#state{riakc_pid=RiakcPid}, RetryPause) ->
+    if RetryPause < 1000 -> ok;
+       true              -> timer:sleep(RetryPause)
+    end,
+    dt_entry(<<"get_block">>, [BlockNumber], [Bucket, Key]),
+    {FullBucket, FullKey} = full_bkey(Bucket, Key, UUID, BlockNumber),
+    StartTime = os:timestamp(),
+    GetOptions1 = [{r, 1}, {n_val, 1}, {sloppy_quorum, false}],
```
kellymclaughlin, line 240: Since we're not planning to include the fs2 backend as part of the next release, but the custom `n_val` and `sloppy_quorum` stuff is part of the next riak release, how about extracting this into a separate PR that we can merge in for the next release? Either way it seems like we need to have a way to disable the `GetOptions1` options for users using versions of Riak that do not support these options.

slfritchie:
Re: Reid's ?FS2_CONTIGUOUS_BLOCKS comment in #591 

An instance of the fs2's customizable variables, including max_blocks (which defaults to ?FS2_CONTIGUOUS_BLOCKS), is written to the instance's ".version.data" file and thus will be constant for the life of that instance.  Changing that parameter won't affect old/preexisting fs2 instances.

The sharing of ?FS2_CONTIGUOUS_BLOCKS with the custom hash algorithm, however, is a problem.  If that changes, then we need to start using a new bucket prefix so that we can use the old constant for old data and the new constant for new data (that always uses the new bucket prefix).  ... Having said that, I'm not expecting this constant to change frequently, if ever.

slfritchie:
Re: @justinsheehy 's general comments in #591 

Yes, space waste via minimum file system page size can be a problem.  If all you store in Riak CS is 1 kilobyte files, you're going to have lots of problems:
1. Unusable disk space, since most file systems nowadays have at least a 4 kilobyte page size
2. Low performance (due to zillions of tiny files in the fs2 backend instances)
3. Facepalming that you didn't use plain old Riak instead.

My thinking is that there will be more use of the multi backend:
1. eleveldb, holding all of the CS metadata that it does today.  Also: the default backend.
2. bitcask, for holding all legacy file data, until the time that we have a method to migrate this legacy CS file data into another medium.
3. fs2, for storing large file data.  Needs a new bucket prefix, e.g. <<"1b:">>.
4. ???, probably eleveldb, for storing small file data, e.g. files between 0 and X bytes, for some value of X smaller than 1 megabyte.  Needs a new bucket prefix, e.g. <<"2b:">>.

We've contemplated doing item number 4 by stuffing the data directly into the manifest record.  We haven't measured the cost of doing that against bucket listing speed: if manifests grow from a couple hundred bytes to many kilobytes each, how much slower is a bucket listing when a bucket has 23 million objects in it?

kuenishi:

shino:
Even if we won't make fs2 as production ready, I think it is better to merge this PR to record the attempt and code. Thoughts?

kuenishi:
This can't be cleanly merged to current develop, maybe rebasing might help. Or re-opening against release/1.5 or older and leave it as it is?

shino:
Repo cleanup sweeping.
It seems this PR is stale? I will close it but
please feel free to re-open if I'm too fast :smile:

PR #600, opened 2013-06-25
Improve object listing performance and efficiency

reiddraper:
When using Riak CS with Riak 1.4 or greater, we can take advantage of paginated
object listing. This allows us to no longer have to do list-keys on the entire
bucket and do a multi-get of all of the objects. This ends up being faster and
cause less load on the cluster. For backward compatibility, this is implemented
as a new FSM, `riak_cs_list_objects_fsm_v2`, which is only used if explicitly
set in the application. The default is false. This commit adds the option to
the `app.config`, so that it's easy for users to see the new flag. It can also
be set in the console like this:

`application:set_env(riak_cs, fold_objects_for_list_keys, true).`

Furthermore, the built in statistics for the new FSM can be seen by turning on
debug logging. This can be done just for the module like this:

`lager:trace_console([{module, riak_cs_list_objects_fsm_v2}], debug).`

There is an EQC test that is added for some pure, helper functions in the new
FSM. There is also a new `riak_test` test that runs the current list objects
test, but uses the new FSM.
- A bunch of duplicated effort in the two FSMs has been put into a shared
  module
- Whether or not to use the new or old FSM is abstracted by the helper module
- The full set of S3 list objects options are supported


kellymclaughlin:
- [x] Pulse tests
- [x] Dialyzer output clean
- [x] `riak_test` tests pass

kellymclaughlin:
Seeing quite a few dialyzer warnings not present on `develop` including some from `riak_cs_list_objects_fsm_v2`.

kellymclaughlin:
Just a reminder that we will need to update `rebar.config` once the riak erlang client has a new tag, preferably as part of this PR.

reiddraper:
> Seeing quite a few dialyzer warnings not present on develop including some from riak_cs_list_objects_fsm_v2.

Whoops. Will get those fixed up tonight and tomorrow I hope.

> Just a reminder that we will need to update rebar.config once the riak erlang client has a new tag, preferably as part of this PR.

Yeah, no particular rush to get this merged, and we should hopefully have tags of riak-erlang-client and friends soon.

kellymclaughlin:
[PR](https://github.com/basho/riak-erlang-client/pull/105) to address dialyzer warnings from `riak_cs_mp_utils`.

kellymclaughlin:
Resolved dialyzer warnings with [2218620](https://github.com/basho/riak_cs/commit/2218620e2054da2106f2e9f0944cca1c0a55db91)

src/riak_cs_list_objects_fsm_v2.erl
```diff
@@ -0,0 +1,495 @@
...
+    ok.
+
+code_change(_OldVsn, StateName, State, _Extra) ->
+    {ok, StateName, State}.
+
+%%%====================================================================
+%%% Internal helpers
+%%%====================================================================
+
+handle_done(State=#state{object_buffer=ObjectBuffer,
```
kellymclaughlin, line 198: This function is pretty long. What do you think about breaking it up? The thing that really jumps out at me is having a separate function to handle the results of calling `enough_results` to get rid of the nested `case` there.

src/riak_cs_list_objects_fsm_v2.erl
```diff
@@ -0,0 +1,495 @@
...
+
+-spec last_result_is_common_prefix(state()) -> boolean().
+last_result_is_common_prefix(#state{object_list_ranges=Ranges,
+                                    req=Request}) ->
+    Key = element(2, lists:last(Ranges)),
+    key_is_common_prefix(Key, Request).
+
+-spec key_is_common_prefix(binary(), list_object_request()) ->
+    boolean().
+%% TODO: please refactor this
```
kellymclaughlin: :+1:

kellymclaughlin:
Seeing some errors like this when listing a bucket with 2000 objects:

```
riak-cs@127.0.0.1)5> 10:33:42.071 [error] gen_fsm <0.16252.0> in state waiting_object_list terminated with reason: no function clause matching riak_cs_list_objects_fsm_v2:waiting_object_list({1359032,[{objects,[]},{continuation,<<"g20AAAADYTE5">>}]}, {state,<0.16041.0>,{list_objects_request_v1,<<"bucket">>,1000,undefined,<<"/">>,undefined},{<0.15425.0>,...},...}) line 143
```

```
10:33:42.079 [error] CRASH REPORT Process <0.16252.0> with 1 neighbours exited with reason: no function clause matching riak_cs_list_objects_fsm_v2:waiting_object_list({1359032,[{objects,[]},{continuation,<<"g20AAAADYTE5">>}]}, {state,<0.16041.0>,{list_objects_request_v1,<<"bucket">>,1000,undefined,<<"/">>,undefined},{<0.15425.0>,...},...}) line 143 in gen_fsm:terminate/7 line 611
```

kellymclaughlin:
Can you think of any way we could make use of the continuations provided with the 2i results? It is not clear to me what could be improved by using them,  but just curious about your thoughts.

reiddraper:
> Can you think of any way we could make use of the continuations provided with the 2i results? It is not clear to me what could be improved by using them, but just curious about your thoughts.

Not specifically for the list objects code, since we basically can create the continuation ourselves.

reiddraper:
> I made a comment on the fs2 PR about maybe extracting the use of the custom n_val and sloppy_quorum options that are also part of Riak 1.4.0. If that happens, maybe it makes sense to use a single config option for all Riak 1.4 features. Not saying it should be changed right now, but something to consider if we do end up with reliance on more than one Riak 1.4 feature.

Yeah, I'm +1 on having an all-encompassing `{'riak-1.4', true}` or something config value.

kellymclaughlin:
How about adding a test case to exercise the continuation case? Perhaps just automating the 2000 object scenario I manually ran.

kellymclaughlin:
Tests are still passing with the latest commits.

riak_test/tests/list_objects_v2_test.erl
```diff
@@ -0,0 +1,31 @@
...
+%% ---------------------------------------------------------------------
+
+-module(list_objects_v2_test).
+
+%% @doc Integration test for list the contents of a bucket
+
+-export([confirm/0]).
+
+confirm() ->
+    Config = [{riak, rtcs:ee_config()}, {stanchion, rtcs:stanchion_config()},
```
kellymclaughlin: `rtcs:ee_config/0` no longer exists on develop. It is replaced by `rtcs:riak_config/0`.

kellymclaughlin:
:+1: Great work!

PR #598, opened 2013-06-28
Manifest pretty-printing utilities

shino:
@shino has some pretty-printing utilities for manifests that we think are generally useful and should be included in the CS package ... with warnings of "you must be at least {X} experienced to use them".  Escript-style packaging is preferred, with packaging path someplace hard-to-use-accidentally.  Useful common functions should be put into BEAM files for use outside of the escripts?


shino:
Escript added. This script is for debugging purpose, so it fetches information directly from Riak, not applying CS utility logic such as manifest resolution.

Sample usage and outputs are here: https://gist.github.com/shino/5882750

In this PR, some executable scripts are moved to `rel/files/` where is suitable place.
And `priv` in top level is linked from `apps/riak_cs/` so that the script goes into release packages.

reiddraper:
This seems to be related to the [`diags`](https://github.com/basho/riak_cs/blob/develop/src/riak_cs_diags.erl) stuff that was recently merged in too.

rel/reltool.config
```diff
@@ -65,8 +65,8 @@
            {template, "files/cert.pem", "etc/cert.pem"},
            {template, "files/key.pem", "etc/key.pem"},
```
kuenishi, line 67: don't forget symlinking to `priv/tools/riak_cs_inspector.erl` !!!

shino, line 67: Ooops. I added a symlink. Please check it again.

kuenishi:
worked for me. +1 to merge.
`make rel` and .deb package creation successfully done

shino:
@reiddraper I think diag tool is used from erlang shell (is it
right?). This tool is escript so it can be used from normal (not erlang)
shell or built in shell scripts. Both usage pattern are useful.

I started to write this tool in order to debug riak cs, so I don't use
CS utility logic such as  manifest record
upgrading or manifest resolution. (Yes I use simple utils funcs to save my time :P )
Also I try to list/show not only manifests but
CS users, CS buckets, manifests in GC bucket.

If this point is aimed by both tools, the best way may be sharing
same logic from both tools finally and used from erlang shell and normal shell.

reiddraper:
@shino awesome. Having a tool from the regular shell is useful too.

PR #605, opened 2013-07-05
document libevent as a riak_test dep

andrewjstone:
#604


reiddraper:
Which test needs this? Just out of curiosity.

andrewjstone:
It looked to be ceph/python related

reiddraper:
+1

PR #606, opened 2013-07-09
Fix riak_test failures as a result of httpc

andrewjstone:
This commit fixes riak_tests that fail due to erlcloud_s3 using httpc.
erlcloud was patched to use ibrowse instead of httpc and the following
changes support api changes required by erlcloud.

fix cs512_regression riak_test

fix object_get_test when using erlcloud_s3 with ibrowse

fix 3 more riak_test tests


kellymclaughlin:
+1 to merge

PR #608, opened 2013-07-09
Print empty twop sets in the GC bucket

shino:


andrewjstone:
+1 to merge

andrewjstone:
Merging this in since I just merged in the code to fix the empty twop_set creation it's checking.

PR #609, opened 2013-07-10
Feature/poolboy stats

glickbot:
Add poolboy statistics for request_pool and bucket_list_pool.


kellymclaughlin:
- [x] Pulse tests
- [x] Dialyzer output clean
- [x] Xref output clean
- [x] `riak_test` tests pass

reiddraper:
I'm not a big fan of the -1 standing in for 'error'. If someone isn't aware of this and starts to treat all poolboy stats numerically, these will just throw off any statistics they try and do, maybe even without them realizing it. Right now I don't think the stats reporting system has a mechanism to say 'this one stat isn't available', but I think we could at least change `raw_report_pool` to just wrap things in `{ok, ...}` or `{error, Reason}`. Thoughts?

kellymclaughlin:
The `-1` seems to be used elsewhere in our stats module so I would suggest keeping it as-is for now and perhaps reexamining the usage later. Just my 2¢.

reiddraper:
> The -1 seems to be used elsewhere in our stats module so I would suggest keeping it as-is for now and perhaps reexamining the usage later. Just my 2¢.

Sounds good to me. +1 as well then.

kellymclaughlin:
:+1: from me as well

kellymclaughlin:
Merging [this](https://github.com/basho/riak_cs/tree/feature/poolboy-stats) rebased and squashed branch instead.

Issue #613, opened 2013-07-15
riakc_pb_socket usage: avoid reuse after timeout [JIRA: RCS-338]


slfritchie:
After a `timeout` using the `riakc_pb_socket` API, the client-side PID is invalid, so trying to use that PID again is invalid.

Suggestion: have the pooling API have a "replace" option where a PID checked out of a pool can be replaced with a (hopefully) better one, without worrying about the deadlock/high latency of trying to check out another PID from the same pool?

shino:
Now, riak_cs connection pool of riakc_pb_socket uses `auto_reconnect` option. It's not an ideal fix for this issue... But I will close this issue because I think new issue/PR is better place to discuss.

PR #614, opened 2013-07-15
Feature/riak client n val

slfritchie:
Use the Riak client's custom `n_val` to lower I/O impact on the Riak cluster.

NOTE: the timeout issues mentioned in https://github.com/basho/riak_cs/issues/613 are not addressed here

### DIFF
--- a/src/riak_cs_block_server.erl
+++ b/src/riak_cs_block_server.erl
@@ -406,7 +406,7 @@ find_md_usermeta(MD) ->
 
 resolve_block_object(RObj, RiakcPid) ->
     {{MD, Value}, NeedRepair} =
-        riak_kv_fs2_backend:resolve_robj_siblings(riakc_obj:get_contents(RObj)),
+        riak_cs_utils:resolve_robj_siblings(riakc_obj:get_contents(RObj)),
     if NeedRepair andalso is_binary(Value) ->
             RBucket = riakc_obj:bucket(RObj),
             RKey = riakc_obj:key(RObj),
--- a/src/riak_cs_block_server.erl
+++ b/src/riak_cs_block_server.erl
@@ -49,8 +49,6 @@
          terminate/2,
          code_change/3]).
 
--define(SERVER, ?MODULE).
-
 -record(state, {riakc_pid :: pid(),
                 close_riak_connection=true :: boolean()}).
 
@@ -114,7 +112,8 @@ get_block(Pid, Bucket, Key, ClusterID, UUID, BlockNumber) ->
 
 -spec put_block(pid(), binary(), binary(), binary(), pos_integer(), binary()) -> ok.
 put_block(Pid, Bucket, Key, UUID, BlockNumber, Value) ->
-    gen_server:cast(Pid, {put_block, self(), Bucket, Key, UUID, BlockNumber, Value}).
+    gen_server:cast(Pid, {put_block, self(), Bucket, Key, UUID, BlockNumber,
+                          Value, riak_cs_utils:md5(Value)}).
 
 -spec delete_block(pid(), binary(), binary(), binary(), pos_integer()) -> ok.
 delete_block(Pid, Bucket, Key, UUID, BlockNumber) ->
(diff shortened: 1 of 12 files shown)


src/riak_cs_block_server.erl
```diff
@@ -329,6 +401,58 @@ full_bkey(Bucket, Key, UUID, BlockId) ->
     FullKey = riak_cs_lfs_utils:block_name(Key, UUID, BlockId),
     {PrefixedBucket, FullKey}.
 
+find_md_usermeta(MD) ->
+    dict:find(?MD_USERMETA, MD).
+
+resolve_block_object(RObj, RiakcPid) ->
+    {{MD, Value}, NeedRepair} =
+        riak_kv_fs2_backend:resolve_robj_siblings(riakc_obj:get_contents(RObj)),
```
kellymclaughlin: Lingering `fs2` reference.

kellymclaughlin:
- [x] Pulse tests
- [x] Dialyzer output clean
- [x] Xref output clean
- [x] `riak_test` tests pass

kellymclaughlin:
Seeing the following dialyzer warnings:

```
riak_cs_block_server.erl:191: The call riak_cs_block_server:do_put_block(FullBucket::binary(),FullKey::<<_:32,_:_*8>>,'undefined',Value::any(),MD::dict(),RiakcPid::'undefined' | pid(),FailFun::fun((_) -> 'ok' | {'error','lager_not_running'})) will never return since it differs in the 3rd argument from the success typing arguments: (binary(),'undefined' | binary(),binary(),binary(),dict(),pid(),fun((_) -> 'ok' | {'error','lager_not_running'}))
riak_cs_block_server.erl:286: Expression produces a value of type 'ok' | {'error','lager_not_running'}, but this value is unmatched
riak_cs_block_server.erl:290: Expression produces a value of type 'ok' | {'error','lager_not_running'}, but this value is unmatched
riak_cs_block_server.erl:409: Call to missing or unexported function riak_kv_fs2_backend:resolve_robj_siblings/1
riak_cs_block_server.erl:410: Expression produces a value of type 'ok' | {'error',_} | {'ok',binary() | {'riakc_obj','undefined' | binary(),'undefined' | binary(),'undefined' | binary(),'undefined' | [{_,_}],'undefined' | dict(),'undefined' | binary()}}, but this value is unmatched
riak_cs_block_server.erl:419: Expression produces a value of type 'ok' | {'error','lager_not_running'}, but this value is unmatched
riak_cs_block_server.erl:452: Expression produces a value of type 'ok' | {'error','lager_not_running'}, but this value is unmatched
riak_cs_get_fsm.erl:311: Expression produces a value of type 'ok' | {'error','lager_not_running'}, but this value is unmatched
```

bookshelfdave:
an eyeball of riak_repl_pb_api:get usage looks good

kellymclaughlin:
Cannot compile after the last commit:

```
src/riak_cs_block_server.erl:251: function do_get_block/9 undefined
src/riak_cs_block_server.erl:252: variable 'BClass' is unbound
ERROR: compile failed while processing /Users/kelly/basho/repos/riak_cs: rebar_abort
make: *** [compile] Error 1
```

kellymclaughlin:
> Cannot compile after the last commit:

Fixed

src/riak_cs_block_server.erl
```diff
@@ -240,6 +214,104 @@ handle_cast({delete_block, ReplyPid, Bucket, Key, UUID, BlockNumber}, State=#sta
...
+    {noreply, State};
+do_get_block(ReplyPid, Bucket, Key, ClusterID, UUID, BlockNumber,
+             State=#state{riakc_pid=RiakcPid}, RetryPause) ->
+    if RetryPause < 1000 -> ok;
+       true              -> timer:sleep(RetryPause)
+    end,
+    dt_entry(<<"get_block">>, [BlockNumber], [Bucket, Key]),
+    {FullBucket, FullKey} = full_bkey(Bucket, Key, UUID, BlockNumber),
+    StartTime = os:timestamp(),
+    GetOptions1 = [{r, 1}, {n_val, 1}, {sloppy_quorum, false}],
```
kellymclaughlin, line 235: @reiddraper raised a good point about having a config option to only use the pre-Riak 1.4 options for people who are using Riak 1.3 or earlier to avoid a request using unsupported options.

slfritchie, line 235: I am heartily in favor of demanding (at a minimum) Riak 1.4.0.

kellymclaughlin, line 235: I don't think that is a viable option for us though. We have users who like to upgrade one application at a time, wait and verify stability, and later upgrade the other or maybe punt and wait for another release in some cases.

reiddraper, line 235: Does `{notfound_ok, false}` work with `[{n_val, 1}, {sloppy_quorum, false}]`? I'm guessing the answer is no, but just want to double-check my understanding.

src/riak_cs_block_server.erl
```diff
@@ -240,6 +214,104 @@ handle_cast({delete_block, ReplyPid, Bucket, Key, UUID, BlockNumber}, State=#sta
...
+    end.
+
+get_block_local(RiakcPid, FullBucket, FullKey, GetOptions, Timeout) ->
+    case riakc_pb_socket:get(RiakcPid, FullBucket, FullKey, GetOptions, Timeout) of
+        {ok, RiakObject} ->
+            resolve_block_object(RiakObject, RiakcPid);
+            %% %% Corrupted siblings hack: just add another....
+            %% [{MD,V}] = riakc_obj:get_contents(RiakObject),
+            %% RiakObject2 = setelement(5, RiakObject, [{MD, <<"foobar">>}, {MD, V}]),
+            %% resolve_block_object(RiakObject2, RiakcPid);
```
reiddraper, line 301: Is this commented out code, or a comment?

src/riak_cs_block_server.erl
slfritchie, line 235: I'd removed it IIRC because it sounded nonsensical when combined with {n_val,1}.

src/riak_cs_block_server.erl
slfritchie, line 301: Comment -- it was handy to leave in for occasional testing.  Adding a knob to control it for automated testing could be good.

src/riak_cs_block_server.erl
reiddraper, line 235: Yeah, it would require the Riak GET FSM to make an n_val=1 request, see that the response was notfound, then make another n_val=1 request to a different vnode. Which might be what we want eventually, but is fine for now I think.

src/riak_cs_block_server.erl
reiddraper, line 301: sounds good

reiddraper:
I think this looks really good overall. I'd like to eventually find a way to test the siblings-from-corruption handling, but I think some manual testing may be best for now. And then eventually to have some fake `riakc_pb_socket` server that can generate corrupt siblings as 'faults' in some EQC test.

reiddraper:
I'm +1 on this. We'll add the 1.4 feature flag in a separate PR. I have a rebased branch [here](https://github.com/basho/riak_cs/compare/develop...feature;riak-client-n_val-rebased-and-such), or we can just deal with the (relatively minor) merge conflicts. No real preference from me.

kellymclaughlin:
:+1: from me as well

kellymclaughlin:
Merging [this](https://github.com/basho/riak_cs/tree/feature/riak-client-n_val-rebased) branch that also has commits squashed.

Issue #607, opened 2013-07-09
GET object generates an empty twop_set in the gc bucket


shino:
Count keys in GC bucket:

```
$ ./scripts.local/riak_cs_inspector.sh 127.0.0.1 dev1 riak-cs-gc count
Connecting to 127.0.0.1:10017...
count(riak-cs-gc): 881
```

GET an object:

```
$ s3cmd -c s3cfg.dev1.alice get s3://test/1 - > /dev/null
```

Count again:

```
$ ./scripts.local/riak_cs_inspector.sh 127.0.0.1 dev1 riak-cs-gc count
Connecting to 127.0.0.1:10017...
count(riak-cs-gc): 882
```

shino:
With #608 empty twop sets can be seen:

```
$ ./scripts.local/riak_cs_inspector.sh 127.0.0.1 dev1 riak-cs-gc | tail -5
1373437379                      : 1        empty_twop_set   empty_twop_set                                  0 empty_twop_set
1373437385                      : 1        empty_twop_set   empty_twop_set                                  0 empty_twop_set
1373437391                      : 1        empty_twop_set   empty_twop_set                                  0 empty_twop_set
1373437414                      : 1        empty_twop_set   empty_twop_set                                  0 empty_twop_set
1373437696                      : 1        empty_twop_set   empty_twop_set                                  0 empty_twop_set
```

shino:
Temporary fix (not examined yet)

```
diff --git a/src/riak_cs_gc.erl b/src/riak_cs_gc.erl
index 32c952a..d417096 100644
--- a/src/riak_cs_gc.erl
+++ b/src/riak_cs_gc.erl
@@ -306,6 +306,8 @@ move_manifests_to_gc_bucket(Manifests, RiakcPid) ->
 %% `riak-cs-gc' bucket to schedule them for deletion.
 -spec move_manifests_to_gc_bucket([lfs_manifest()], pid(), boolean()) ->
     ok | {error, term()}.
+move_manifests_to_gc_bucket([], _RiakcPid, _AddLeewayP) ->
+    ok;
 move_manifests_to_gc_bucket(Manifests, RiakcPid, AddLeewayP) ->
     Key = generate_key(AddLeewayP),
     ManifestSet = build_manifest_set(Manifests),
```

shino:
Another (unrelated) useless PUT elimination:

```
diff --git a/src/riak_cs_gc.erl b/src/riak_cs_gc.erl
[...]

PR #615, opened 2013-07-15
fix for #607 from @shino

andrewjstone:
Prevent the creation of empty twop_sets when there are no manifests to gc.
Prevent unnecessary puts to riak when there are no manifests to gc.

H/T @shino


andrewjstone:
- [x] doesn't create new empty twop_sets on get
- [x] eunit + eqc
- [x] riak_test tests
- [x] dialyzer clean - a few things showed up, that I believe are unrelated.

shino:
Thank you for every confirmation and merging, Andrew!!

PR #596, opened 2013-07-17
Return HTTP 400 if content-md5 does not match

kellymclaughlin:
On PUT, Riak currently will not return HTTP 400 if the Content-MD5 header and the MD5 of the body do not match. S3 [does](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html). It does return the correct MD5 as the ETAG, and some clients do recognize this and throw and exception. Returning HTTP 400 is easy. Making sure we properly garbage collect the 'bad' version is a bit trickier.

### DIFF
--- a/src/riak_cs_oos_response.erl
+++ b/src/riak_cs_oos_response.erl
@@ -22,6 +22,7 @@
 -export([api_error/3,
          error_code_to_atom/1,
          error_response/5,
+         invalid_digest_response/3,
          respond/3,
          status_code/1]).
 
@@ -69,6 +70,22 @@ api_error({error, Reason}, RD, Ctx) ->
 
 error_response(StatusCode, _Code, _Message, RD, Ctx) ->
     {{halt, StatusCode}, RD, Ctx}.
+
+%% @TODO Figure out how this should actually be formatted
+invalid_digest_response(ContentMd5, RD, Ctx) ->
+    XmlDoc = {'Error',
+              [
+               {'Code', [error_code(invalid_digest)]},
+               {'Message', [error_message(invalid_digest)]},
+               {'Content-MD5', [ContentMd5]},
+               {'HostId', ["host-id"]}
+              ]},
+    Body = riak_cs_xml:export_xml([XmlDoc]),
+    UpdRD = wrq:set_resp_body(Body,
+                              wrq:set_resp_header("Content-Type",
+                                                  ?XML_TYPE,
+                                                  RD)),
+    {{halt, status_code(invalid_digest)}, UpdRD, Ctx}.
 
 %% @doc Convert an error code string into its corresponding atom
 -spec error_code_to_atom(string()) -> atom().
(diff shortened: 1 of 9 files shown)


reiddraper:
thoughts on making a separate TODO for writing a client-test for this?

kellymclaughlin:
> thoughts on making a separate TODO for writing a client-test for this?

yeah good idea

reiddraper:
I'm reviewing locally merged or rebased with #622.

reiddraper:
- [x] Pulse tests
- [x] Dialyzer output clean
- [x] Xref output clean
- [x] `riak_test` tests pass

reiddraper:
client test passed, but int-test is failing. Will see if this is happening on develop too.

kellymclaughlin:
> client test passed, but int-test is failing. Will see if this is happening on develop too.

Addressed by [a352742](https://github.com/basho/riak_cs/commit/a352742c0cfc06746f07e455dc509b4054f49cdb) and [bc7bccf](https://github.com/basho/riak_cs/commit/bc7bccf5834e8cc8aa6aa979d282c3d88e8b2543).

kellymclaughlin:
Rebased from latest `develop`

reiddraper:
> Addressed by a352742 and bc7bccf.

Nice, that fixes the test-int problems.

src/riak_cs_oos_response.erl
```diff
@@ -70,6 +71,22 @@ api_error({error, Reason}, RD, Ctx) ->
 error_response(StatusCode, _Code, _Message, RD, Ctx) ->
     {{halt, StatusCode}, RD, Ctx}.
 
+%% @TODO Figure out how this should actually be formatted
```
reiddraper, line 74: In what way are we not sure how this should be formatted? Does this just mean we haven't tested that this makes clients throw the correct exception? If so, that can be part of #624.

reiddraper, line 74: Oh right, this is also just for OpenStack, right?

kellymclaughlin, line 74: Right, they don't have any documentation I have found about expected error responses or error formatting and I haven't had a chance to manually determine it yet so punting for now.

reiddraper:
+1, great work

reiddraper:
See #624 to track testing of this PR.

src/riak_cs_put_fsm.erl
```diff
@@ -282,25 +275,53 @@ not_full({augment_data, NewData}, From,
...
+                                         mani_pid=ManiPid,
+                                         riakc_pid=RiakPid,
+                                         timer_ref=TimerRef}) ->
+    _ = erlang:cancel_timer(TimerRef),
+    %% The digest is invalid. Write the manifest and immediately
+    %% schedule it for gc.
+    _ = maybe_update_manifest_with_confirmation(ManiPid, Manifest),
+    _ = riak_cs_gc:gc_active_manifests(Bucket, Key, RiakPid),
+    gen_fsm:reply(From, {error, invalid_digest}),
+    {stop, invalid_digest, State};
```
reiddraper, line 304: This causes a crash report to be printed. I might argue that this is still a 'normal' exit.

PR #619, opened 2013-07-18
Feature/user mgmt cli

reiddraper:
## Do Not Merge

This is just to start a conversation about how a Riak CS user-management CLI might look/work. I wrote it in Ruby because _fog_ already has good support for our user-management API.


reiddraper:
Currently usage is something like this:

``` shell
bundle install
export KEY=R2MWYA07NVSCX4-NYSSP
export SECRET=5qKDKQhgRBSVDKatl48L4aCYJfLlUQtLVJyPpA==

# list the users in the system, one json object per line:
bundle exec bin/mgmt -k $KEY -s $SECRET list

# create a new user
bundle exec bin/mgmt -k $KEY -s $SECRET create 'First Last' 'first@example.com'
```

hectcastro:
Haha, I was going to mess with this over the weekend since you mentioned it last week.

reiddraper:
> Haha, I was going to mess with this over the weekend since you mentioned it last week.

Please don't hesitate to add on to what I've done, or take another approach if this doesn't look right. I have very poor ruby skills...

reiddraper:
ps. the use of global variables and magic in this code goes against everything I stand for. Oops.

hectcastro:
@reiddraper Not sunshine and roses, but here is my stab at reproducing the same functionality you had in a different way: https://github.com/hectcastro/riak-cs-user-mgmt

reiddraper:
@hectcastro very cool. I'll play with that too this week.

reiddraper:
Closing this as a PR, I think I was liking @hectcastro's approach better anyway.

PR #633, opened 2013-08-01
Add kill switch for n_val=1 block get requests

reiddraper:


andrewjstone:
Tested against Riak 1.3.1
- [x] riak_tests pass

Tested against Riak 1.4
- [x] riak_tests pass
- [x] eunit + eqc tests pass

Keep it clean yo
- [x] dialyzer clean
- [x] xref clean

reiddraper:
Please test this with both Riak 1.3.x and 1.4.x. I want someone else to verify the extra PB messages are ignored in 1.3.x.

andrewjstone:
@reiddraper You just read my mind. Will do. 

Is there an easy way to test the specific n_val functionality, besides just ensuring it doesn't break current tests?

reiddraper:
> Is there an easy way to test the specific n_val functionality, besides just ensuring it doesn't break current tests?

That's a good question. Programmatically, I'm not sure. My first thought would be to run Riak on a single node and use redbug to see how many vnodes were being contacted.

reiddraper:
@andrewjstone +1'd from hipchat.

PR #636, opened 2013-08-02
Align ERL_MAX_PORTS with Riak default: 64000

macintux:
At some point Riak was bumped from 4096 by default to 64000. R16 is actually higher by default, at 65536.

We'll eventually have to address the fact that R17 moves this to +Q


reiddraper:
I think we probably want to target this at the `release/1.4` branch.

reiddraper:
+1

PR #638, opened 2013-08-02
Update to node package 1.3.2

reiddraper:


jaredmorrow:
+1, will fix two issues in RPMs, stanchion will need similar treatment

PR #372, opened 2013-08-05
bucket policy: use X-Forwarded-For instead of peer

kuenishi:
> If riak-cs is behind a load balancer or reverse proxy, we should be looking for the IP address from a header, like [X-Forwarded-For](http://en.wikipedia.org/wiki/X-Forwarded-For). How about two app.config settings

from Reid's comment: https://github.com/basho/riak_cs/pull/364#discussion_r2543118

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -145,6 +145,12 @@
               %% 21600 is 6-hours.
               {gc_retry_interval, 21600},
 
+              %% If your load balancer adds 'X-Forwarded-For' header
+              %% and it is reliable (able to gaurantee it is not added
+              %% by malicious user), turn this true. Otherwise, by
+              %% default, Riak CS takes source IP address as an input.
+              {trust_x_forwarded_for, false},
+
               %% == DTrace ==
 
               %% If your Erlang virtual machine supports DTrace (or
--- a/src/riak_cs_config.erl
+++ b/src/riak_cs_config.erl
@@ -40,7 +40,8 @@
          response_module/1,
          riak_cs_stat/0,
          set_md5_chunk_size/1,
-         use_t2b_compression/0
+         use_t2b_compression/0,
+         trust_x_forwarded_for/0
         ]).
 
 %% OpenStack config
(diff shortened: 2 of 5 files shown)


kuenishi:
This is nearly a security issue so I'd like to include into 1.4 release.

TODO:
- [ ] add client_tests
- [ ] pass riak_test

andrewjstone:
While defaulting to `{trust_x_forwarded_for, false}`, is certainly safer, I'm concerned that changing this to default to false may break existing customers using HA-Proxy. @reid @kelly @kuenishi are we sure this isn't an issue? If it is, should we just add the note to the release?

Also, @kuenishi are you adding tests or do you want me to take a stab at them?

kellymclaughlin:
> While defaulting to {trust_x_forwarded_for, false}, is certainly safer, I'm concerned that changing this to default to false may break existing customers using HA-Proxy. @reid @kelly @kuenishi are we sure this isn't an issue? If it is, should we just add the note to the release?

I would say that since it was not an explicitly supported behavior before, we are probably okay to change the default to `false` as long as we call it out in the release notes and docs.

reiddraper:
> I would say that since it was not an explicitly supported behavior before, we are probably okay to change the default to false as long as we call it out in the release notes and docs.

+1

andrewjstone:
- [x] eunit + eqc tests
- [x] manual testing
- [x] client tests
- [x] riak_test

:+1:

kuenishi:
> Also, @kuenishi are you adding tests or do you want me to take a stab at them?

No, not yet. I don't have enough time to even set up and write r_t for now...  I'd add my tests at end of  `boto_test.py` with existing tests around bucket policy. I'd be very happy if someone had taken over me.

andrewjstone:
Cool @kuenishi. I can take over. Thanks for the pointer of where to start.

kuenishi:
@andrewjstone Sweet and thank you so much!

andrewjstone:
I manually confirmed that webmachine will try to use the x-forwarded-for header for an ip-address and that the new option prevents this. The only spot in our tests that hits the security check here: https://github.com/basho/riak_cs/commit/8f8ad8b3e70c46c1ce5b46ce8580b3750731fc00#L2R496
is 
https://github.com/basho/riak_cs/blob/develop/client_tests/python/boto_tests/boto_test.py#L428

shino:
Does this work with https? For example `webmachine_request:get_peer` treats both plain tcp and ssl socket.
cc: @kuenishi @andrewjstone

kuenishi:
Ooops! No it doesn't work because in case of ssl #wm_request.peer={ssl, socket()}. But webmachine_request:get_peer refers #wm_reqstate.peer, which may refer "x-forwarded-for" header via x_peername/2. So use direct pattern matching will work, rather than using that function. Will open a new PR.

https://github.com/basho/webmachine/blob/1.10.3/src/webmachine_request.erl#L116

PR #645, opened 2013-08-06
Add a new test case the the user riak_test module

kellymclaughlin:
Add a test case to test that an ordinary user may not change the status of
another ordinary user.


reiddraper:
+1

PR #647, opened 2013-08-08
can't get proper IP address in case of trust_x_forwarded_for=true and scheme is ssl

kuenishi:
in favor of #372, pointed out by @shino

### DIFF
--- a/src/riak_cs_s3_policy.erl
+++ b/src/riak_cs_s3_policy.erl
@@ -493,13 +493,15 @@ eval_bool(#wm_reqdata{scheme=Scheme} = _Req, Conds) ->
 
 -spec eval_ip_address(#wm_reqdata{}, [{'aws:SourceIp', binary()}]) -> boolean().
 eval_ip_address(Req, Conds) ->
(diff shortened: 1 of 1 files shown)


andrewjstone:
Wow, quick work @kuenishi ! I'll review this today. 

Also, a great find by @shino

andrewjstone:
- [x] eunit + eqc
- [x] riak_test
- [x] manual verification of code
  - enable ssl
  - set bucket policy to deny localhost
  - test get with localhost - ensure failure
  - set bucket policy to deny some other ip
  - test get with localhost - ensure success

:+1: to merge

kuenishi:
Verifying code is much harder and more valuable than writing it. I'll appreciate a lot, thank you!

andrewjstone:
In order to test this, I came across a serious issue. None of the S3 clients we use support ssl in proxy mode. I tested s3cmd, s3curl and erlcloud. I therefore created a patched version of s3cmd that will work in proxy mode with ssl. With a few modifications it should probably be submitted back upstream, but for now it's a quick hack to ensure I can continue to test this issue.

https://github.com/andrewjstone/s3cmd

PR #649, opened 2013-08-12
Add config option to avoid local bucket check at creation time

kellymclaughlin:
Add a disable_local_bucket check option to disable local checking for a
bucket's existence when determining if a bucket creation request should be
sent to stanchion. The default is false and represents an optimization for
work flows that unnecessarily create the same bucket many times. Setting it to
true can help in situations where the user's view of bucket ownership differs
from the canonical ownership record.


reiddraper:
+1

PR #657, opened 2013-08-28
add riak-cs-stanchion command to switch stanchion manually

kuenishi:
Add a command to switch stanchion location setting. This command
assumes a use case where:
1. stanchion fails (or kill the old one, ensure killed)
2. decide a new stanchion with a brand new location (host,port)
3. put a new stanchion location via `riak-cs-stanchion switch`

Also, users can confirm stanchion setting by `show` subcommand.
I believe this will make operation in case of stanchion failure much easier
than before.

I'd like to discuss about the risk that this command may lead operational
mistake to diverse stanchion settings.

### DIFF
--- a/rel/files/riak-cs-stanchion
+++ b/rel/files/riak-cs-stanchion
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+# Pull environment for this install
+. "{{runner_base_dir}}/lib/env.sh"
+
+# Make sure the user running this script is the owner and/or su to that user
+check_user $@
+
+# Make sure CWD is set to runner run dir
+cd $RUNNER_BASE_DIR
+
+# Check the first argument for instructions
+case "$1" in
+    switch|show)
+        # Make sure the local node IS running
+        node_up_check
+
+        $NODETOOL rpc riak_cs_stanchion_console $@
+        ;;
+    *)
+        echo "Usage: $SCRIPT { switch HOST PORT | show }"
+        exit 1
+        ;;
+esac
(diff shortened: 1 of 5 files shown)


src/riak_cs_stanchion_console.erl
```diff
@@ -0,0 +1,75 @@
...
+        end).
+
+-define(SCRIPT_NAME, "riak-cs-stanchion").
+-include("riak_cs.hrl").
+
+%%%===================================================================
+%%% Public API
+%%%===================================================================
+
+%% @doc Kick off a gc round, unless one is already
```
reiddraper: copy-paste comment maybe?

kuenishi: eew, copied.

rel/files/riak-cs-stanchion
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+
```
reiddraper, line 2: thoughts on a `set -e` for this script?

reiddraper:
I like this. My biggest fear is that people won't understand that this is only a runtime change. It does not _persist_ the change, so if Riak CS is restarted, it will revert back to the old value. Maybe a warning will help?

kuenishi:
fmm, that might be a problem, because people including me usually don't read warning. My idea, plus a warning, is to enable `/ping` at stanchion and let Riak CS ping stanchion when starting up the process. Anyway this should be documented strongly.

I can't wait built-in Rafter.

rel/files/riak-cs-stanchion
kuenishi, line 2: I thought I had set this script executable. Or I messed up sth with git?

kuenishi:
Ready for review again! English composition might be a bit better.

kuenishi:
I'm merging this PR, but still open to additional or critical discussions on persistence and diversing stanchion settings.

reiddraper:
Cool. Thanks @kuenishi.

PR #664, opened 2013-09-06
Bump erlcloud dependency to 0.4.3

kellymclaughlin:
The 0.4.3 tag of erlcloud increases some ibrowse timeouts and this
helps with spurious riak_test failures when running on slow hosts.


shino:
+1

PR #665, opened 2013-09-06
Set pb_backlog in Riak config for riak_test runs

kellymclaughlin:
Specify a pb_backlog of 256 in the Riak app.config for riak_test runs to avoid
spurious riak_test failures due to disconnected Riak pids.

This problem was observed when setting up testing on the fedora 17
buildbot builder VM.


shino:
+1

PR #667, opened 2013-09-09
Bugfix/reduce wastefully large object size

shino:
I embedded typo in object size and riak_test uses large memory and takes time. Oops, sorry...
This fix reduce 400MB object to 400 bytes.

P.S. This fix changes only a test case of riak_test. So does not affect 1.4.1 release.


slfritchie:
Ha!  Yeah, this should be a bit faster.  +1

PR #670, opened 2013-09-11
Quick fix for manifest props=undefined, needed for upgrades from very old versions of Riak CS

slfritchie:
Witnessed after upgrading from an ancient version of Riak CS to 1.4.1.

```
2013-09-11 09:05:05.871 UTC [error] <0.25995.4> CRASH REPORT Process <0.25995.4> with 0 neighbours exited with reason: no function clause matching proplists:is_defined(deleted, undefined) line 184 in gen_fsm:terminate/7 line 611
2013-09-11 09:05:05.872 UTC [error] <0.25994.4> gen_fsm <0.25994.4> in state prepare terminated with reason: {{function_clause,[{proplists,is_defined,[deleted,undefined],[{file,"proplists.erl"},{line,184}]},{riak_cs_manifest_utils,delete_time,1,[{file,"src/riak_cs_manifest_utils.erl"},{line,342}]},{riak_cs_manifest_utils,'-latest_delete_time/1-fun-0-',2,[{file,"src/riak_cs_manifest_utils.erl"},{line,368}]},{lists,foldl,3,[{file,"lists.erl"},{line,1197}]},{riak_cs_manifest_utils,deleted_while_writing,1,[{file,"src/riak_cs_manifest_utils.erl"},{line,355}]},{riak_cs_utils,gc_deleted_while_writing_manifests,...},...]},...}
2013-09-11 09:05:05.873 UTC [error] <0.25994.4> CRASH REPORT Process <0.25994.4> with 0 neighbours exited with reason: {{function_clause,[{proplists,is_defined,[deleted,undefined],[{file,"proplists.erl"},{line,184}]},{riak_cs_manifest_utils,delete_time,1,[{file,"src/riak_cs_manifest_utils.erl"},{line,342}]},{riak_cs_manifest_utils,'-latest_delete_time/1-fun-0-',2,[{file,"src/riak_cs_manifest_utils.erl"},{line,368}]},{lists,foldl,3,[{file,"lists.erl"},{line,1197}]},{riak_cs_manifest_utils,deleted_while_writing,1,[{file,"src/riak_cs_manifest_utils.erl"},{line,355}]},{riak_cs_utils,gc_deleted_while_writing_manifests,...},...]},...} in gen_fsm:terminate/7 line 611
```

This patch is a quick, end-of-work-day-want-to-get-out-of-the-office fix.  It is extremely lightly tested by the compiler only.

PR #671, opened 2013-09-11
Handle undefined `props` in manifests

reiddraper:
Prior to Riak CS 1.3.0 (667319391c4ef3d5c0d757e11952a5202cc64050 in
particularly), manifests were written with `undefined` in their `props`
field, despite the record definition setting it to `[]`. This change
fixes the manifests as they are read, and 'upgraded'. That way, the rest
of the code-base can assume the field is a proper proplist.

This is another approach to fixing #670.


andrewjstone:
- [x] eunit & eqc tests
- [x] riak_test
- [x] dialyzer
- [x] xref

:+1: to merge

PR #658, opened 2013-09-12
Reset xref and dialyzer ignore files

kuenishi:
Particularly if we move to a different version of Erlang, I think it's good practice to start fresh with our xref and dialyzer 'grep -v' files.


kuenishi:
Here's the raw result of dyalizer: https://gist.github.com/kuenishi/6537524

reiddraper:
Careful, this wants to merge into master.

kuenishi:
Sorry! I'll open a new issue for convenience.

Issue #672, opened 2013-09-12
Can't build source tarball since 1.4.1


kuenishi:
We can't compile source package distributed from our download page .  rebar said this was due to `git describe --always --tags` failed. This was led since `src/riak_cs.app.src` had `{vsn, git}` entry, which was changed at f26c5fb21810ae1fa25706d998164e4c3ee1bc4f .  There might have been many workarounds to avoid this but I don't know which is the best.

```
$ pwd
/home/kuenishi/staging2/riak-cs-1.4.1
$ ./rebar get-deps && ./rebar -vvv compile
(snip)
INFO:  Skipped src/riak_cs_delete_fsm_sup.erl
DEBUG: Worker exited cleanly
INFO:  Skipped src/riak_cs_wm_bucket_location.erl
DEBUG: Worker exited cleanly
INFO:  Skipped src/riak_cs_kv_multi_backend.erl
DEBUG: Worker exited cleanly
INFO:  No app_vars_file defined.
DEBUG: vcs_vsn: Fallback to vcs for /home/kuenishi/staging2/riak-cs-1.4.1
INFO:  sh info:
        cwd: "/home/kuenishi/staging2/riak-cs-1.4.1"
        cmd: git describe --always --tags
DEBUG:  opts: [{cd,"/home/kuenishi/staging2/riak-cs-1.4.1"},
               {use_stdout,false}]
ERROR: git describe --always --tags failed with error: 128 and output:
fatal: Not a git repository (or any parent up to mount point /home)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).

ERROR: compile failed while processing /home/kuenishi/staging2/riak-cs-1.4.1: rebar_abort
```

PR #678, opened 2013-09-24
Create vsn.git at priv/ under top level directory

shino:
This PR addresses #672 .

Riak does not have top level `src/*.app.src` but Riak CS and Stanchion have.
By creating `priv/vsn.git`, rebar finds it and compiles `*.app.src` .


kuenishi:
.deb and source package successfully built. +1 to merge.

PR #680, opened 2013-09-26
Fix rebar.config.script check for OTP version

kellymclaughlin:
Some new functions were added to the crypto API after R15B01, but the hmac/3
and hmac/4 functions did not arrive until R16. Only set new_hash if the OTP
version is R16 or above.


reiddraper:
+1. Let's update velvet to c643ab9042fc22bd623df86980663a3e6b55d0dd when basho/velvet#13 gets merged too.

Issue #684, opened 2013-10-02
riak-cs-stanchion command added to OSS but not EE riak-cs


jaredmorrow:
This commit https://github.com/basho/riak_cs/commit/3ef109b7b77a195030f292dd066442d0d2265b90 added the riak-cs-stanchion command to the open source build of riak-cs, but not riak-cs-ee.

kuenishi:
nice work!

PR #685, opened 2013-10-02
Add `riak-cs-stanchion` script to riak-cs-ee packages

jaredmorrow:
Addresses: #684

The `riak-cs-stanchion` script was added to the riak-cs packages, but not riak-cs-ee.


kuenishi:
Oops. Thanks. +1 to merge.

PR #686, opened 2013-10-03
Add default proxy get cluster id

reiddraper:
Can be set with:
`application:set_env(riak_cs, default_proxy_cluster_id, RemoteClusterID).`

---

Not sure if this is the approach we want to go with for long-term code, but starting it as a conversation.


kellymclaughlin:
+1 to merge

PR #688, opened 2013-10-04
Add option to allow a delay for the initial GC collection

kellymclaughlin:
Add an option that allows the initial run of the GC daemon to be delay by some
number of seconds. This provides a convenient way for the GC daemon to be run
on multiple nodes without overlapping the collection and cleanup effort among
nodes. The option is called initial_gc_delay and is specified in seconds. The
value only applies to the initial collection run when the riak_cs node is
started.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,7 +168,16 @@ xref: compile
 ## Packaging targets
 ##
 .PHONY: package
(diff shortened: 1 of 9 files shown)


andrewjstone:
- [x] eunit + eqc
- [x] dialyzer clean  - found some unrelated errors

:+1: to merge

kellymclaughlin:
Closing and will reopen a new pr to merge into `release/1.4` instead.

Issue #681, opened 2013-09-30
Performance of DELETE does not scale


kuenishi:
[IIJ presented a benchmark (mosly in Japanese, sorry)](http://www.slideshare.net/takashisogabe/riak-19684891) on Riak CS - the result was mostly good except scalability of DELETE. Probably this is due to PUT contention onto a single key in GC bucket when deleting an object.

The fix for this problem might be adding a suffix to a key in GC bucket like `<epoch>++<somerandomstuff>` . If we choose this fix list_keys in gc_d must be changed to stream_list_keys before.

note: this also addresses #6134

reiddraper:
> This fix for this problem might be adding a suffix to a key in GC bucket like <epoch>++<somerandomstuff> . If we choose this fix list_keys in gc_d must be changed to stream_list_keys before.

This is definitely worth trying. Either way, I'd love to see the streaming list keys stuff added to the gc daemon.

reiddraper:
Anyone know if we've seen the same performance bottlenecks with PUTs that overwrite a key? That should hit the same, synchrounous GC bucket writing code path as DELETE.

shino:
Got some numbers for random prefix

code is here: https://github.com/basho/riak_cs/compare/release%2F1.4...feature%2Fgc-key-scattering
- 4 boxes, one riak + one riak-cs in each box
- basho_bench in another (5th) box, cs driver
- delete operation with preload 10KB objects
- compare release/1.4 branch and the branch above

```
| branch            | concurrent | DELs/sec |
|-------------------+------------+----------+
| release/1.4       |          1 |       50 |
| release/1.4       |          2 |       80 |
| release/1.4       |          4 |      100 |
| release/1.4       |          8 |      100 |
|-------------------+------------+----------+
| gc-key-scattering |          1 |       50 |
| gc-key-scattering |          8 |      220 |
| gc-key-scattering |         16 |      270 |
| gc-key-scattering |         32 |      300 |
```

The release/1.4 branch does not go beyond 100 DELs/sec, and
there happens 9 siblings at concurrent=4 and 36 siblings at concurrent=8.

By prepending random numbers, siblings is at most 2 with concurrent=8 or 16.
I don't understand why there is ~300 DELS/sec limit. User/Bucket get hot keys?

reiddraper:
Nice work, thanks @shino!

shino:
I have almost same throughput result for millisec suffix as random suffix.
[...]

PR #673, opened 2013-09-12
Reset xref and dialyzer ignore files

kuenishi:
`Particularly if we move to a different version of Erlang, I think it's good practice to start fresh with our xref and dialyzer 'grep -v' files.` 

in favor of https://github.com/basho/riak_cs/pull/658 which I broke the thread by pull-requesting to master :fearful: 
- [x] fix velvet
- [x] fix riak-erlang-client and another
- [x] fix xref (nothing to do)
- [x] discuss on `behaviour_info/1` (and remove the commented out code)
- [x] review

PR #689, opened 2013-10-07
Append random number to epoch binary

shino:
This PR addresses #681 .

Every CS DELETE execute GET/PUT to a single BKey = {riak-cs-gc, Epoch}.
Appending random number to Epoch (hopefully) solve this hot key problem.

GC daemon can collect garbage with suffixed key because it accesses
keys by binary range 2i.

### DIFF
--- a/src/riak_cs_gc.erl
+++ b/src/riak_cs_gc.erl
@@ -250,6 +250,7 @@ leeway_seconds() ->
 timestamp() ->
     timestamp(os:timestamp()).
 
+-spec timestamp(erlang:timestamp()) -> non_neg_integer().
 timestamp(ErlangTime) ->
     riak_cs_utils:second_resolution_timestamp(ErlangTime).
 
@@ -344,17 +345,17 @@ build_manifest_set(Manifests) ->
 -spec generate_key(boolean()) -> binary().
 generate_key(AddLeewayP) ->
     Now = os:timestamp(),
-    list_to_binary([integer_to_list(key_timestamp(Now, AddLeewayP)),
-                    $_,
-                    key_suffix(Now)]).
+    list_to_binary([key_timestamp(Now, AddLeewayP), $_, key_suffix(Now)]).
 
+-spec key_timestamp(erlang:timestamp(), boolean()) -> string().
 key_timestamp(Time, true) ->
-    timestamp(Time) + leeway_seconds();
+    integer_to_list(timestamp(Time) + leeway_seconds());
 key_timestamp(Time, false) ->
-    timestamp(Time).
+    integer_to_list(timestamp(Time)).
 
+-spec key_suffix(erlang:timestamp()) -> string().
 key_suffix(Time) ->
-    random:seed(Time),
+    _ = random:seed(Time),
     integer_to_list(random:uniform(riak_cs_config:gc_key_suffix_max())).
 
 %% @doc Given a list of riakc_obj-flavored object (with potentially
--- a/src/riak_cs_gc.erl
+++ b/src/riak_cs_gc.erl
@@ -346,14 +346,18 @@ build_manifest_set(Manifests) ->
 -spec generate_key(boolean()) -> binary().
 generate_key(AddLeewayP) ->
     Now = os:timestamp(),
-    random:seed(Now),
-    list_to_binary(
-      [integer_to_list(
-         timestamp(Now) + if not AddLeewayP -> 0;
-                             true           -> leeway_seconds()
-                          end),
-       $_,
-       integer_to_list(random:uniform(?GC_KEY_SUFFIX_MAX))]).
+    list_to_binary([integer_to_list(key_timestamp(Now, AddLeewayP)),
+                    $_,
+                    key_suffix(Now)]).
+
+key_timestamp(Time, true) ->
+    timestamp(Time) + leeway_seconds();
+key_timestamp(Time, false) ->
+    timestamp(Time).
+
+key_suffix(Time) ->
+    random:seed(Time),
+    integer_to_list(random:uniform(?GC_KEY_SUFFIX_MAX)).
 
 %% @doc Given a list of riakc_obj-flavored object (with potentially
 %%      many siblings and perhaps a tombstone), decode and merge them.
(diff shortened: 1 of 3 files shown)


kellymclaughlin:
I did some micro-benchmarking of `crypto:rand_uniform/2` versus `random:uniform/1`. I wrote a small test module to facilitate testing:

```
-module(random_test).

-compile(export_all).

test1(X) ->
    timer:tc(?MODULE, crypto_test, [X]).

test2(X) ->
    timer:tc(?MODULE, random_test, [X]).

crypto_test(X) ->
    [integer_to_list(crypto:rand_uniform(0, 1000)) || _ <- lists:seq(1, X)].

random_test(X) ->
    [integer_to_list(random:uniform(1000)) || _ <- lists:seq(1, X)].
```

My results show that using `random:uniform/1` would be more efficient. Some sample results follow.

Timings for 10K iterations:

```
(riak-cs@127.0.0.1)19> random_test:test1(10000).
{93860,
 ["231","77","754","198","580","718","215","922","356","71",
  "486","896","10","692","523","548","698","162","414","74",
  "964","805","464","676","922","388",
  [...]|...]}
(riak-cs@127.0.0.1)20> random_test:test2(10000).
{14011,
 ["609","635","133","368","417","316","594","57","903","28",
  "508","323","49","438","592","566","638","965","242","451",
  "869","874","201","822","883","246",
  [...]|...]}
```

Timings for 1M iterations:

```
(riak-cs@127.0.0.1)17> random_test:test1(1000000).
{11322768,
 ["862","772","761","124","496","523","836","6","28","529",
  "454","692","4","775","133","378","260","579","729","811",
  "777","892","403","446","377","78",
  [...]|...]}
(riak-cs@127.0.0.1)18> random_test:test2(1000000).
{1560378,
 ["991","499","883","879","342","567","96","866","45","474",
  "503","723","491","857","473","999","596","40","105","412",
  "857","104","584","75","80","569",
  [...]|...]}
```

slfritchie:
We only need to generate a single random number per delete, right?  So using random:uniform() should be fine, we don't need super-crypto-randomness.  We just need to make certain that each FSM uses a unique seed (e.g. using os:timestamp()), otherwise random:uniform() won't do what we want.

kellymclaughlin:
>  So using random:uniform() should be fine, we don't need super-crypto-randomness.

The really interesting thing I learned poking through the source of the `crypto` module today is that even `crypto:rand_uniform/2` does not give you super-crypto-randomness. Instead of calling the openssl function `BN_rand_range` which (according to the [docs](http://www.openssl.org/docs/crypto/BN_rand.html)) "...generates a cryptographically strong pseudo-random number..." the Erlang crypto nif calls `BN_pseudo_rand_range` which "...does the same, but is based on BN_pseudo_rand(), and hence numbers generated by it are not necessarily unpredictable." Just not what I expected from a function in the `crypto` module.

shino:
Thank you for precise comments, Kelly and Scott.

I had first used `random:uniform` and incorrectly remembered seed was
initialized automatically by timestamp :/
And,... I saw many siblings generated (yeah, no surprise).

> We just need to make certain that each FSM uses a unique seed (e.g. using os:timestamp())

This time I did :)
I force-pushed code using `random:uniform`.

kellymclaughlin:
The change to `random:uniform` looks good.

src/riak_cs_gc.erl
```diff
@@ -340,11 +345,15 @@ build_manifest_set(Manifests) ->
...
 generate_key(AddLeewayP) ->
+    Now = os:timestamp(),
+    random:seed(Now),
     list_to_binary(
-      integer_to_list(
-        timestamp() + if not AddLeewayP -> 0;
-                         true           -> leeway_seconds()
-                      end)).
+      [integer_to_list(
+         timestamp(Now) + if not AddLeewayP -> 0;
```
kellymclaughlin: I find that the `if` statement here really makes it harder to understand the logic at first glance. I think a small refactoring to remove the use of the `if` improves readability and clarity. The way I would rewrite it would be like this:

```
generate_key(AddLeewayP) ->
    Now = os:timestamp(),
    list_to_binary([key_timestamp(Now, AddLeewayP), $_, key_suffix(Now)]).

key_timestamp(Time, true) ->
    timestamp(Time) + leeway_seconds();
key_timestamp(Time, false) ->
    timestamp(Time).

key_suffix(Time) ->
    random:seed(Time),
    integer_to_list(random:uniform(?GC_KEY_SUFFIX_MAX)).
```

shino:
Refactored. It looks far better than before ;) Thanks, Kelly.

I discussed with Kota about `GC_KEY_SUFFIX_MAX`. It was 1000 yesterday.

Lager value is better for scalability (keys are well distributed). But
more keys may stress GC list keys.  In order to be safe, I changed the
default value from 1000 to 256. Also make it configurable by app env.
If delete throughput is needed for ring size > 512, we can advice to
set it larger value and be careful about GC timeout.
Comments and advice are welcome (of course, always :) ).

kellymclaughlin:
Two last comments on this.
- Can you add specs for `key_timestamp`, `key_suffix`, and `timestamp/1` in `riak_cs_gc`?
- `key_timestamp` returns an integer, but `key_suffix` calls `integer_to_list` and returns a string. I would vote to move the `integer_to_list` call for the `key_timestamp` result inside `key_timestamp` to make it congruent with `key_suffix`.

shino:
Push one commit fc139a8

There are some dialyzer output : 

```
meck_code_gen.erl:193: The pattern <Mod, Func, Args, [{'meck_code_gen', 'exec', _AriOrArgs} | Stack]> can never match the type <atom(),atom(),[any()],[{atom(),atom(),[any()] | byte(),[any()]},...]>
meck_proc.erl:77: Invalid type specification for function meck_proc:start/2. The success typing is (atom(),[any()]) -> 'ok'
riak_cs_config.erl:191: Function will never be called
riak_cs_stats.erl:196: The variable _ can never match since previous clauses completely covered the type {atom(),integer(),integer(),integer()}
```

I want to defer them to #673 (if no one has objection...)

kuenishi:
Sure, I can work on it after 1.4 merged to develop branch.

kellymclaughlin:
+1 to merge. Great work!

PR #692, opened 2013-10-07
Add operation timing logging

reiddraper:
Starting a conversation here. Originally these were cooked up for a quick deploy, so there might be better, cleaner ways of organizing things for a PR. But I wanted to get this started to document that conversation.

### DIFF
--- a/src/riak_cs_utils.erl
+++ b/src/riak_cs_utils.erl
@@ -573,7 +573,13 @@ get_object(BucketName, Key, RiakPid) ->
     {ok, riakc_obj:riakc_obj()} | {error, term()}.
 get_manifests_raw(RiakcPid, Bucket, Key) ->
     ManifestBucket = to_bucket_name(objects, Bucket),
-    riakc_pb_socket:get(RiakcPid, ManifestBucket, Key).
+    StartTime = os:timestamp(),
+    Manifests = riakc_pb_socket:get(RiakcPid, ManifestBucket, Key),
+    EndTime = os:timestamp(),
+    TimeDiff = timer:now_diff(EndTime, StartTime),
+    _ = lager:debug("Took ~B microseconds to read manifest: ~p",
+                    [TimeDiff, [Bucket, Key]]),
+    Manifests.
 
 %% @doc
 -spec get_manifests(pid(), binary(), binary()) ->
--- a/src/riak_cs_utils.erl
+++ b/src/riak_cs_utils.erl
@@ -606,6 +606,7 @@ manifests_from_riak_object(RiakObject) ->
                        %% Original value had many contents
                        riakc_obj:get_contents(RiakObject)
                end,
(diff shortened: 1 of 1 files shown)


src/riak_cs_utils.erl
```diff
@@ -394,12 +394,21 @@ delete_object(Bucket, Key, RiakcPid) ->
...
         true ->
             term_to_binary(Term, [compressed]);
         false ->
             term_to_binary(Term)
-    end.
+    end,
+    EndTime = os:timestamp(),
+    TimeDiff = timer:now_diff(EndTime, StartTime),
+    Size = byte_size(Binary),
+    lager:debug("term_to_binary took ~B microseconds for ~B bytes, with"
```
kellymclaughlin:
```
riak_cs_utils.erl:408: Expression produces a value of type 'ok' | {'error','lager_not_running'}, but this value is unmatched
```

reiddraper: Amended this into the first commit, now cc92bdb8e4810e92d0d069962a55e971ae3897ef.

slfritchie:
Hrm.  What do we know about lager:debug() overhead vs. DTrace overhead vs. Erlang built-in tracing overhead?  If lager's level checking is far more expensive than the others, then I'd recommend the others?

reiddraper:
> Hrm. What do we know about lager:debug() overhead vs. DTrace overhead vs. Erlang built-in tracing overhead? If lager's level checking is far more expensive than the others, then I'd recommend the others?

This is probably worth asking @Vagabond about. I'm under the impression that messages below the current log-level threshold are 'cheap', but for what value of 'cheap', I don't know.

Vagabond:
They are 'cheap' but not free. An ETS (formerly mochiglobal) lookup is done per log message to see if the current log level is elegible for logging.

reiddraper:
@kellymclaughlin any thoughts?

reiddraper:
The current consensus is that this is too risky to ship with until we've done performance testing. We'll have the basho-patches option until then, and the option of looking into how to use Erlang's tracing for a similar effect.

Vagabond:
I'd be interested in hearing more on why this is 'risky'.

reiddraper:
@Vagabond suppose it's still open for debate. @kellymclaughlin and I were thinking that since we didn't fully understand the performance impact, it was a bit rushed to get it in a point release (due to go out in a day or so) without doing our normal performance testing. That being said, I'm open minded about this.

kellymclaughlin:
> @Vagabond suppose it's still open for debate. @kellymclaughlin and I were thinking that since we didn't fully understand the performance impact, it was a bit rushed to get it in a point release (due to go out in a day or so) without doing our normal performance testing. That being said, I'm open minded about this.

Agreed, I doubt there would be any significant performance impact, but without running performance tests it's just conjecture. I'd rather keep it as something to slide into basho-patches until we have time to gather real data and we just don't have that time right now.

PR #698, opened 2013-10-11
Set pb_backlog in Riak config for riak_test runs

shino:
Backport of #665 to `release/1.4` branch.

Specify a pb_backlog of 256 in the Riak app.config for riak_test runs
to avoid spurious riak_test failures due to disconnected Riak pids.


kuenishi:
passed r_t, +1.

PR #711, opened 2013-10-17
Add dependency graphing

reiddraper:


reiddraper:
Ala [Riak](https://github.com/basho/riak/commit/941bd6508213ed38b16b58954b6d92f1251370ca).

reiddraper:
h/t to @jared and @beerriot who did the real work getting this on Riak.

shino:
In my environment, I cloned riak_cs to `riak_cs-1.5`. Then the output is:

Quick fix:

```
diff --git a/misc/mapdeps.erl b/misc/mapdeps.erl
index 6775c7a..28381c9 100755
--- a/misc/mapdeps.erl
+++ b/misc/mapdeps.erl
@@ -102,8 +102,7 @@ map_rebar(BaseDir, Path, Acc) ->

 app_name(Path) ->
     %% assumes Path ends in rebar.config
-    re:replace(filename:basename(filename:dirname(Path)),
-               "-", "_", [global, {return, list}]).
+    filename:basename(filename:dirname(Path)).

 file_start() ->
     io:format(standard_io, "digraph {~n", []).
@@ -112,4 +111,4 @@ file_end() ->
     io:format(standard_io, "}~n", []).

 file_edge(From, To) ->
-    io:format(standard_io, " ~s -> ~s;~n", [From, To]).
+    io:format(standard_io, " \"~s\" -> ~s;~n", [From, To]).
```

Quoting of node ID is specified at http://www.graphviz.org/content/dot-language .
(I'm not familiar with graphviz :P )

>  An ID is one of the following:
>  ...
>  any double-quoted string ("...") possibly containing escaped quotes ('");

By quoting, i guess replacement from `-` to `_` becomes needless so remove it.
riak-cs

Everything else looks good! If you confirm and accept the above fix, :+1: to merge.

reiddraper:
Nice catch. We'll probably want to get this updated in Riak and Riak EE as well.

PR #715, opened 2013-10-22
Change GC daemon to optionally use paginated 2I queries

kellymclaughlin:
Change the GC daemon to optionally use paginated 2I queries to avoid system
stress and timeouts when the amount of garbage to be collected is large. The
`paginated_indexes` boolean option is used to enable the use of paginated
queries and may be used when the version of Riak being used is >= 1.4.0. The
default for the option is false. There is also an option, `gc_batch_size`, to
configure the amount of results that should be requested by the 2I query at
one time. The default for this option is 1000.


th-m-vogel:
Hello,

and thank you. Pull and rebuild worked fine. Can you please describe more detailed how and where to configure this Options?

best regards
Thomas

Date: Mon, 21 Oct 2013 17:00:27 -0700

th-m-vogel:
Hello,

think I figured it out myself, just added the following two lines to my riak-cs app.config:

```
          {gc_batch_size, 100},
```

{paginated_indexes, true},

GC is still running in timeouts on some files, but cleaning up slowly. In such situations (based on the configured batch size of 100), if I run riak-cs-gc status the batch is running for 60 seconds without decreasing the number of files left in current run. If I look in detail to the crash log for each run the data block behind 

Data  == {state,<<"test">>,<<"Folder/filename.ext">>,

is shrinking at each GC run, looks like until the file is cleaned up. I reduced the time between GC runs also.

There is no noticeable load on the system during such a GC run.

regards
Thomas

From: vogel@startplus.de
To: reply@reply.github.com
Subject: RE: [riak_cs] Change GC daemon to optionally use paginated 2I queries (#715)
Date: Tue, 22 Oct 2013 13:44:06 +0200

Hello,

and thank you. Pull and rebuild worked fine. Can you please describe more detailed how and where to configure this Options?

best regards
Thomas

Date: Mon, 21 Oct 2013 17:00:27 -0700

src/riak_cs_config.erl
```diff
@@ -147,6 +153,10 @@ set_key_list_multiplier(Multiplier) ->
 policy_module() ->
     get_env(riak_cs, policy_module, ?DEFAULT_POLICY_MODULE).
 
+-spec paginated_indexes() -> atom().
+paginated_indexes() ->
+    get_env(riak_cs, paginated_indexes, false).
```
reiddraper: should this also be prefixed with `gc_`?

reiddraper:
An emacs file `include/#riak_cs_gc_d.hrl#` snuck in.

src/riak_cs_config.erl
kellymclaughlin: yep, probably should. will make that change.

reiddraper:
+1. Great work.

PR #720, opened 2013-10-28
Improve handling of Riak disconnection errors on object downloads

kellymclaughlin:
Change the behavior when a Riak connection is closed during an object 
downloading. There is existing retry logic to handle the case when n_val=1
block requests fail with a notfound or timeout error. It is desirable to also
retry requests in the case of a transient connection failure, especially in
the case of the download of a sizable object. Allowing the disconnect error to
trigger the retry logic provides time for the automatic reconnection process
to work and provides a mechanism for in-progress downloads to continue and 
complete successfully.


reiddraper:
+1

PR #724, opened 2013-10-31
Don't export variables from case statements

reiddraper:
I temporarily used the `warn_export_vars` compile option to spot these. I've also spotted a few areas where we repeat ourselves, but I think it'd be best to tackle that in a separate issue. Or a separate commit on this same branch.

### DIFF
--- a/src/riak_cs_access_archiver.erl
+++ b/src/riak_cs_access_archiver.erl
@@ -185,18 +185,8 @@ riak_connection() ->
 %% that has been merged.
 -spec riak_host_port() -> {string(), pos_integer()}.
 riak_host_port() ->
-    case application:get_env(riak_cs, riak_ip) of
-        {ok, Host} ->
-            ok;
-        undefined ->
-            Host = "127.0.0.1"
-    end,
-    case application:get_env(riak_cs, riak_pb_port) of
-        {ok, Port} ->
-            ok;
-        undefined ->
-            Port = 8087
-    end,
+    Host = application:get_env(riak_cs, riak_ip, "127.0.0.1"),
+    Port = application:get_env(riak_cs, riak_pb_port, 8087),
     {Host, Port}.
 
 cleanup(Table, Pid, Mon) ->
--- a/src/riak_cs_app.erl
+++ b/src/riak_cs_app.erl
@@ -128,16 +128,6 @@ riak_connection() ->
 %% that has been merged.
 -spec riak_host_port() -> {string(), pos_integer()}.
 riak_host_port() ->
-    case application:get_env(riak_cs, riak_ip) of
-        {ok, Host} ->
-            ok;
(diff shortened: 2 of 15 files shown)


src/riak_cs_access_archiver.erl
```diff
@@ -185,18 +185,8 @@ riak_connection() ->
...
-        undefined ->
-            Host = "127.0.0.1"
-    end,
-    case application:get_env(riak_cs, riak_pb_port) of
-        {ok, Port} ->
-            ok;
-        undefined ->
-            Port = 8087
-    end,
+    Host = application:get_env(riak_cs, riak_ip, "127.0.0.1"),
```
kellymclaughlin, line 188: How about making some functions in `riak_cs_config` for these?

kellymclaughlin, line 188: Or even a single function that returns `{Host, Port}`

src/riak_cs_app.erl
```diff
@@ -128,16 +128,6 @@ riak_connection() ->
...
-        undefined ->
-            Host = "127.0.0.1"
-    end,
-    case application:get_env(riak_cs, riak_pb_port) of
-        {ok, Port} ->
-            ok;
-        undefined ->
-            Port = 8087
-    end,
+    Host = application:get_env(riak_cs, riak_ip, "127.0.0.1"),
```
kellymclaughlin, line 131: `application:get_env/3` only appeared in R16 so builds with R15 fail at start up

kellymclaughlin:
- [x] Eunit tests
- [x] EQC tests
- [x] Integration tests
- [x] Client tests
- [ ] Dialyzer output clean
- [ ] xref output
- [x] riak_test run

kellymclaughlin:
Should we use `warn_export_vars` by default?

reiddraper:
> Should we use warn_export_vars by default?

A bunch of our dependencies fail with this on. Maybe there's some way we can turn it on for CS, but not our dependencies?

src/riak_cs_app.erl
reiddraper, line 131: Yeah, I was thinking we were going all-in for R16 for the next version, but maybe instead we should DRY-up all of the times we use this and use the case statement there, so we can be backward compatible.

src/riak_cs_access_archiver.erl
reiddraper, line 188: +1, and that way we can avoid using get_env/3 in multiple places. Just use get_env/2 once.

kuenishi:

kuenishi:
Moving to 1.5.1 as cleaning up the code is good, while forcing `warn_export_vars` would be another issue (it's bigger fish to fry).

reiddraper:
Doesn't merge cleanly. Closing.

PR #730, opened 2013-11-13
Upgrade to lager 2.0.1

reiddraper:
This allows us to use lager's tracing again without fear. The fix is in
basho/lager#175.


kellymclaughlin:
:+1:

PR #732, opened 2013-11-14
Consider manifests for pruning based on count

reiddraper:
If the `riak_cs` application env `max_scheduled_delete_manifests` is set
to an integer (default is `unlimited`), manifests will not only be
pruned based on `scheduled_delete_time`, but based on the number of
`scheduled_delete` manifests there are. If there are more than
`max_scheduled_delete_manifests`, they will be sorted by time (more
recent first), and then cut-off at `max_scheduled_delete_manifests`.
This means that in cases where a key is overwritten frequently, we have
the option of pruning manifests more frequently than just the leeway
time.

### DIFF
--- a/src/riak_cs_gc.erl
+++ b/src/riak_cs_gc.erl
@@ -37,6 +37,7 @@
          gc_specific_manifests/5,
          epoch_start/0,
          leeway_seconds/0,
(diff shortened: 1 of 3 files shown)


kellymclaughlin:
Nice work! +1 to merge.

PR #733, opened 2013-11-14
Fix path substitution code to fix Solaris source builds

jaredmorrow:
Addresses: basho/riak#439

The same issue will also be an issue with riak-cs.


jaredmorrow:
+1'd from @andrewjstone in chat

PR #717, opened 2013-11-15
Only a single access archiver can run at a time

reiddraper:
The `riak_cs_access_archiver` module incorrectly uses a local process name when starting which effectively restricts the maximum number of archiver processes that can be running to 1.


kellymclaughlin:
:+1:

PR #738, opened 2013-11-20
Improve handling of block read request timeouts

kellymclaughlin:
Block read requests that take longer than the timeout results in an
`{error, disconnected}` tuple being returned instead of `{error, timeout}`. 
The current code is written such that when `n_val=1` requests are enabled and
a timeout occurs that returns `{error, disconnected}`, the request is retried
again with `n_val=1` set. This means that the request is sent to the same
vnode again. In the case of a severely backed-up vnode queue this likely
results in another timeout and `{error, disconnected}` result from the pb
client.  This continues until the retry limit is reached and an eventual
`{error, notfound}` is returned to the process that called the block server.
Change the block server code to add `disconnected` as a reason to trigger a
read attempt that does not use the `n_val=1` option. This should allow the
read request to complete and not cause downloads to stall in the case of one
or more very slow responding Riak vnodes.

### DIFF
--- a/src/riak_cs_block_server.erl
+++ b/src/riak_cs_block_server.erl
@@ -276,14 +276,14 @@ try_local_get(RiakcPid, FullBucket, FullKey, GetOptions1, GetOptions2,
             ProceedFun(Success);
         {error, {insufficient_vnodes,_,need,_}} ->
             RetryFun(NumRetries + 2);
-        {error, Why} when Why == notfound; Why == timeout;
-                          Why == <<"{insufficient_vnodes,0,need,1}">> ->
+        {error, Why} when Why == notfound;
+                          Why == timeout;
+                          Why == disconnected;
+                          Why == <<"{insufficient_vnodes,0,need,1}">>;
+                          Why == {insufficient_vnodes,0,need,1} ->
             handle_local_notfound(RiakcPid, FullBucket, FullKey, GetOptions2,
                                   ProceedFun, RetryFun, NumRetries, UseProxyGet,
                                   ProxyActive, ClusterID);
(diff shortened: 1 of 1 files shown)


slfritchie:
Kelly, is there any possibility that this (or closely related error) could also affect ZenDesk # 6509?  @ksauzz has been working on that ticket.

src/riak_cs_block_server.erl
```diff
@@ -276,14 +276,13 @@ try_local_get(RiakcPid, FullBucket, FullKey, GetOptions1, GetOptions2,
             ProceedFun(Success);
         {error, {insufficient_vnodes,_,need,_}} ->
             RetryFun(NumRetries + 2);
```
slfritchie, line 278: Hrm, should `handle_local_notfound()` be used here on line 278 also?

kellymclaughlin:
> Kelly, is there any possibility that this (or closely related error) could also affect ZenDesk # 6509? @ksauzz has been working on that ticket.

I do not think it is directly related to this, but could be a similar scenario. I'll comment directly on that ticket with some thoughts.

src/riak_cs_block_server.erl
kellymclaughlin, line 278: I'm almost certain this clause will never match because the pb client sends back these errors as stringified binaries as shown on [L282](https://github.com/basho/riak_cs/pull/738/files#diff-605b96365a09fe99001357092feca21bR282) below. I propose to either remove this clause completely or simply add an additional guard to the next clause.

reiddraper:
Code looks good, and tests pass. +1

PR #705, opened 2013-11-20
Don't delete current object version on bad md5

reiddraper:
It [looks likes](https://github.com/basho/riak_cs/blob/bc99be0d4c62947a36cc7932f8db8173b54a4d68/src/riak_cs_put_fsm.erl#L290-L303) we delete the currently active manifest on an upload if the md5 is wrong, instead of just never putting this one into the active state.

### DIFF
--- a/client_tests/python/boto_tests/boto_test.py
+++ b/client_tests/python/boto_tests/boto_test.py
@@ -620,6 +620,27 @@ class ContentMd5Test(S3ApiVerificationTestBase):
         with self.assertRaises(S3ResponseError):
             key.set_contents_from_string('this is different from the md5 we calculated', md5=x)
 
+    def test_bad_md5_leaves_old_object_alone(self):
+        '''Github #705 Regression test:
+           Make sure that overwriting an object using a bad md5
+           simply leaves the old version in place.'''
+        key_name = str(uuid.uuid4())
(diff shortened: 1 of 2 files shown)


reiddraper:
Confirmed this is an issue. Due to some other code organization, it might be easier now to simply leave the manifest in the writing state, rather than pro-actively garbage-collect it. I'm continuing to investigate.

reiddraper:
What do you all think about getting this in 1.4.3?

kellymclaughlin:
- [x] Pulse tests pass
- [x] Client tests pass
- [x] Int tests pass
- [x] riak_test tests pass

kellymclaughlin:
Verified the additional python test fails with current `release/1.4` code and passes with this change.

reiddraper:
> Verified the additional python test fails with current release/1.4 code and passes with this change.

Marginally related, any thoughts on pros/cons of having a bugfix and the accompanying test in the same commit, or doing two commits?

kellymclaughlin:
> Marginally related, any thoughts on pros/cons of having a bugfix and the accompanying test in the same commit, or doing two commits?

I think it's fine since they are very much related. The only thing I would consider is amending the commit message to mention the addition of the test. Other than that I have no problem with it.

kellymclaughlin:
Looks good! +1 to merge.

PR #753, opened 2013-12-10
Add composer.lock file

reiddraper:
Composer recommends that the lockfile be checked into version control
[1]. This ensures that we're all testing the same dependencies, and that
new installations don't run into version conflicts.

[1] http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file

### DIFF
--- a/client_tests/php/.gitignore
+++ b/client_tests/php/.gitignore
@@ -1,3 +1,2 @@
 phpunit
-composer.lock
 vendor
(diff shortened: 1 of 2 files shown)


kuenishi:
+1. Looks harmless. Do you have any ideas on how do we update composer.lock to catch up with the latest version? When bumping the version would be nice but we have to try not to forget upgrading this.

reiddraper:
> Do you have any ideas on how do we update composer.lock to catch up with the latest version?

I haven't tried it yet, but the [documentation](http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file) suggests you can just run:

``` bash
$ php composer.phar update
```

I agree we should be trying to keep it up to date.

PR #755, opened 2013-12-12
Ignore Owner with an empty ID in AccessControlPolicy blob

arekinath:
Some clients (eg. Cyberduck) are a bit special and can send AccessControlPolicy with an Owner element that has an empty ID element inside, when they are only meant to be updating the ACL. Currently we handle the case where an empty Owner element is sent, but not if it has an empty ID inside it (results in a 500 error at present, due to a badmatch).

I propose that we simply ignore an Owner with an empty element inside, instead of crashing out -- it's what the Amazon S3 seems to do, and it seems vaguely sensible.

For completeness' sake this patch handles any empty child inside the Owner, not just an ID (although I've only ever seen an empty ID sent on its own in the wild so far). It will still crash out in the case that we get an owner with a non-element child or a child with more than one child inside it, which I think is fine for now (as these cases are very clearly malformed).

### DIFF
--- a/src/riak_cs_acl_utils.erl
+++ b/src/riak_cs_acl_utils.erl
@@ -298,23 +298,27 @@ process_owner([], Acl, _) ->
     {ok, Acl};
 process_owner([HeadElement | RestElements], Acl, RiakPid) ->
     Owner = Acl?ACL.owner,
-    [Content] = HeadElement#xmlElement.content,
-    Value = Content#xmlText.value,
-    ElementName = HeadElement#xmlElement.name,
-    case ElementName of
-        'ID' ->
-            _ = lager:debug("Owner ID value: ~p", [Value]),
-            {OwnerName, _, OwnerKeyId} = Owner,
-            UpdOwner = {OwnerName, Value, OwnerKeyId};
-        'DisplayName' ->
-            _ = lager:debug("Owner Name content: ~p", [Value]),
-            {_, OwnerId, OwnerKeyId} = Owner,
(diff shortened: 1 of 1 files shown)


arekinath:
P.S. this applies cleanly to 1.4 as well, and it might be suitable to do it there too (I'm running with it applied to the 1.4 branch at the moment)

reiddraper:
@kuenishi @shino either of you have time to take a look at this?

kuenishi:
@rieddraper , OK, I'll take a look and try to put into 1.4.5 if possible.

@arekinath I tried Cyberduck and try changing object and bucket's ACL, but the XML Cyberduck sent to CS was not empty. What client did you exactly tested?

The content of `Content` has been always

``` erl
{xmlText,[{'DisplayName',2},{'Owner',1},{'AccessControlPolicy',1}],1,[],"ed336f04c62d1790b3a9d223d2b75e68b8827dd546719fd20a859a36140a0314",text}
```

arekinath:
@kuenishi I originally had this problem on Cyberduck 4.3.1 on OSX. It looks like it's no longer occurring on Cyberduck 4.4.3, so maybe this issue is not so much worth looking at any more...

kuenishi:
@arekinath ha. As long as I tested, the only region which accepts empty owner is us-east-1 and other region such as ap-northeast-1 does not accept empty owner (returns 403). Probably that's why Cyberduck fixed their software. Anyway, thank you so much for reporting and contributing. As feedback I opened a new pull request #790 where a logging line is added.

PR #757, opened 2013-12-20
Add check for presence of eqc in rebar test plugin

kellymclaughlin:


shino:
rebar seems to define `EQC` macro properly.
https://github.com/basho/rebar/blob/117c0f7e698f735acfa73b116f9e38c5c54036dc/src/rebar_erlc_compiler.erl#L233

Just removing defining EQC macro works?

kellymclaughlin:
> rebar seems to define EQC macro properly.

It does, but that is just for the `eunit` and `qc` commands which place the beam files in a certain place. This plugin is a helper for our client test and riak_test files. In these cases we want the beams to live outside of `.eunit`. This change is to avoid failures running the `external_client_tests` riak_test module solely for the reason that eqc is not present.

shino:
@kellymclaughlin Thanks for clarification. I will try that pattern.

shino:
Confirmed
1.   `make test-erlang` executes EQC when EQC is in ERL_LIBS
2.   `make test-erlang` executes no tests when EQC is not in ERL_LIBS

:+1:

reiddraper:
woot, thanks @kellymclaughlin!

Issue #743, opened 2013-11-25
Improve handling of storage calculation request timeouts


ksauzz:
riak_cs_storage_d doesn't handle request timeouts well like #738.  [here?](https://github.com/basho/riak_cs/blob/213aeee94f2cd3e0cf3fef53c37bf52bbd4abe73/src/riak_cs_storage.erl#L72)

crash.log

```
2013-11-13 14:58:51 =ERROR REPORT==== 
** State machine riak_cs_storage_d terminating 
** Last event in was continue 
** When State == calculating 
** Data == {state,[],undefined,{{2013,11,13},{5,29,58}},undefined,<0.7057.679>,{{2013,11,13},{5,29,58}},2,0,[<<"xxxxxxxxxxxxxxx-xxxxxx">>,<<"xxxxxxxxxxxxxxxxxxxxxxxx">>,<<"xxxxxxxxxxxxxxxxxxxxxxx">>,<<"xxxxxxxxxxxxxxxxxxxx">>,<<"xxxxxxxxxxxxxxxxxxxxxxx">>],true} 
** Reason for termination = 
** {json_encode,{bad_term,{error,{timeout,[]}}}} 
2013-11-13 14:58:51 =CRASH REPORT==== 
crasher: 
initial call: riak_cs_storage_d:init/1 
pid: <0.29342.291> 
registered_name: riak_cs_storage_d 
exception exit: {{json_encode,{bad_term,{error,{timeout,[]}}}},[{gen_fsm,terminate,7,[{file,"gen_fsm.erl"},{line,611}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,227}]}]} 
ancestors: [riak_cs_sup,<0.29198.291>] 
messages: [{28046540,{error,<<"timeout">>}},{23190669,{error,<<"timeout">>}},{62182351,{error,<<"timeout">>}},{121450601,{error,<<"timeout">>}},{94032563,{error,<<"timeout">>}},{22873166,{error,<<"timeout">>}},{7996964,{error,<<"timeout">>}},{90730175,{error,<<"timeout">>}}] 
links: [<0.29202.291>,<0.7057.679>] 
dictionary: [] 
trap_exit: false 
status: running 
heap_size: 28657 
stack_size: 24 
reductions: 21761 
neighbours: 
[...]

Issue #703, opened 2013-10-11
Add tests for access and storage statistics


reiddraper:
We've had a couple of regressions recently -- paste due time to add some tests.

shino:
Some tests for access stats are added at #747

shino:
storage tests : https://github.com/basho/riak_cs/pull/800/files#diff-1

kuenishi:
I thought this was meant to be some eqc testing for more safety.

shino:
EQC over riak_test? Possible (reasonably) ?

shino:
reopen for the place of discussion

kuenishi:
+1 to close this and eqc testing might be future work or we totally update from scratch.

PR #759, opened 2013-12-26
handle timeout error on sum_bucket/2

kuenishi:
addresses #743.
from commit message:

```
add error handling when sum_bucket failed with timeout, which
causes dirty error message and no way to know which bucket was
failed in usage calculation mapreduce. This commit adds two ways
to know that:

- outputs an error log that prints user key id and bucket name
  with error reason.
- stores the result into moss.storage with reason like:
    { "bucket_name" : "{error,{timeout,[]}}" }

The latter log proves and indicated there existed a bucket at
that time even though its usage size was unknown. Better than
ignored, which operator can't tell non-existent from calculation
failure.
```

### DIFF
--- a/src/riak_cs_storage.erl
+++ b/src/riak_cs_storage.erl
@@ -46,10 +46,27 @@ sum_user(Riak, User) when is_binary(User) ->
 sum_user(Riak, User) when is_list(User) ->
     case riak_cs_utils:get_user(User, Riak) of
         {ok, {?RCS_USER{buckets=Buckets}, _UserObj}} ->
-            {ok, [ {B, sum_bucket(Riak, B)}
-                   || ?RCS_BUCKET{name=B} <- Buckets ]};
+            BucketUsages = [maybe_sum_bucket(Riak, User, B)
+                            || ?RCS_BUCKET{name=B} <- Buckets],
+            {ok, BucketUsages};
         {error, Error} ->
             {error, Error}
+    end.
+
+%% @doc Output a log when calculating total usage of a bucket.
+%%      This log is *very* important because unless this log
+%%      there are no other way for operator to know a calculation
+%%      which riak_cs_storage_d failed.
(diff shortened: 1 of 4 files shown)


andrewjstone:
Code looks good. 
- [x] cs743_regression_test riak_test
- [x] eunit + eqc
- [x] dialyzer
- [x] xref

andrewjstone:
I'm seeing the following dialyzer errors related to this change:

```
riak_cs_storage.erl:60: Invalid type specification for function riak_cs_storage:maybe_sum_bucket/3. The success typing is (pid(),[binary() | maybe_improper_list(any(),binary() | []) | byte()],binary() | [byte()]) -> {binary() | [byte()],binary()}
riak_cs_storage.erl:64: The pattern {'ok', BucketUsage} can never match the type {'error',_} | {'struct',[{_,_},...]}
```

kuenishi:
Hooray! It was totally a bug found by dialyzer.
This regression could be found if #703 were done.

andrewjstone:
:+1:

src/riak_cs_storage.erl
```diff
@@ -46,12 +46,30 @@ sum_user(Riak, User) when is_binary(User) ->
...
-                   || ?RCS_BUCKET{name=B} <- Buckets ]};
+            BucketUsages = [maybe_sum_bucket(Riak, User, B)
+                            || ?RCS_BUCKET{name=B} <- Buckets],
+            {ok, BucketUsages};
         {error, Error} ->
             {error, Error}
     end.
 
+%% @doc Output a log when calculating total usage of a bucket.
+%%      This log is *very* important because unless this log
```
reiddraper, line 57: s/unless/without

kuenishi, line 57: Oh crap. Thank you for pointing it out.

Issue #721, opened 2013-10-29
Comment on cs_version is obsolete


shino:
https://github.com/basho/riak_cs/blob/develop/rel/files/app.config#L82

http://docs.basho.com/riakcs/1.4.2/cookbooks/Rolling-Upgrades-For-Riak-CS/ is correct.

PR #760, opened 2013-12-26
more precise description of cs_version

kuenishi:


kuenishi:
addresses #721.

shino:
+1 from another non-native English speaker

PR #762, opened 2013-12-27
Don't run unnecessary node at every riak_test

kuenishi:
Not all riak_test at Riak CS have to run at four nodes, esp. CS node 
No.2,3,4 are rarely used. Also Riak node dev2,dev3,dev4 participates
into the cluster. The time to setup unused nodes are getting longer
as number of tests increase.
This commit changes node number of most tests from 4 to 1, which
improved the time to run all tests from 22min => 15min in my laptop.

Save 7 minutes!

### DIFF
--- a/riak_test/tests/cs296_regression_test.erl
+++ b/riak_test/tests/cs296_regression_test.erl
@@ -30,7 +30,7 @@
 -define(TEST_BUCKET, "riak-test-bucket").
 
 confirm() ->
-    {UserConfig, {_RiakNodes, _CSNodes, _Stanchion}} = rtcs:setup(4),
+    {UserConfig, {_RiakNodes, _CSNodes, _Stanchion}} = rtcs:setup(1),
 
     lager:info("User is valid on the cluster, and has no buckets"),
     ?assertEqual([{buckets, []}], erlcloud_s3:list_buckets(UserConfig)),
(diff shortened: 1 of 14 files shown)


shino:
- [x] riak_test successful execution
- [x] riak_test shorter execution time
- [ ] source diff

shino:
I agree to use just single node cluster for simple functional tests to
shorten test interval (hmm, as long as we do them by hand?).

Just a random thought but how about to define default node number by
macro or add `rtcs:setup/0` ?
- A cluster with multiple nodes is, in principle, better than
  single node because it is little bit more similar to real cases.
  (Multiple nodes are on single machine yet...)
- If we could introduce automated testing environment (buildbot?),
  then it would be nice to use single node cluster in local environment
  to shorten test interval and use multiple nodes in automated environment.

kuenishi:
I prefer explicit way of naming, so I like naming as `rtcs:single_setup/0` and `rtcs:setup/0` test writers are smart enough to use suitable one. Probably adding default setting should be in `.riak_test.config` ? To configure the number of nodes at runtime I prefer it being explicit, to being implicit by macros.

shino:
You convinced me :)

So I want following two tests have 4-node-setting at least:
- list_objects_v2_test uses coverage operations special to CS 
- external_client_test does many many kinds of tests (i don't know all
  the details but some list operations are executed) and takes ~ 5 min
  in my environment. Overhead by multiple nodes is relatively small.

shino:
:+1:

PR #470, opened 2014-01-03
Allow objects to be reaped sooner than leeway interval.

kellymclaughlin:
Change how manifest delete marking works so that objects may be cleaned up prior to the expiration of the leeway interval. The current scheme works by writing the deleted or overwritten manifest version to a garbage collection bucket with a key that is representative of a timestamp in the future when the blocks of the object represented by the manifest are eligible to actually be deleted. The garbage collection daemon periodically checks the garbage collection bucket for keys that represent a time that is in the past and judges any such key as eligible to be reaped.  

The problem with this is approach is that there is not a way to override the leeway interval in the case of a disk space shortage, bloated manifests, or situations where the system may be under stress. To improve this situation the process should be upended so that the keys written to the GC bucket are representative of the current time and not a future point in time and the enforcement of the `leeway_interval` should be done by the GC daemon. This would allow early cleanup of garbage to be accomplished even after a manifest is written to the GC bucket.

The goals of this work are as follows:
- [x] Change the keys written to the GC bucket to be representative of the current time instead of a time in the future
- [x] Move enforcement of the `leeway_interval` to the GC daemon
- [x] Provide a way to override the default `leeway_interval` from the `riak-cs-gc` tool

### DIFF
--- a/src/riak_cs_gc.erl
+++ b/src/riak_cs_gc.erl
@@ -38,7 +38,7 @@
          epoch_start/0,
          leeway_seconds/0,
          max_scheduled_delete_manifests/0,
-         move_manifests_to_gc_bucket/3,
+         move_manifests_to_gc_bucket/2,
          timestamp/0]).
 
 %% export for repl debugging and testing
@@ -179,11 +179,6 @@ handle_mark_as_pending_delete({ok, RiakObject}, Bucket, Key, UUIDsToMark, RiakcP
     Manifests = riak_cs_utils:manifests_from_riak_object(RiakObject),
     PDManifests = riak_cs_manifest_utils:manifests_to_gc(UUIDsToMark, Manifests),
     MoveResult = move_manifests_to_gc_bucket(PDManifests, RiakcPid),
-    %% riak_cs_gc.erl:185: The pattern [{UUID, _} | _] can never match
-    %% the type [] Oi, this is a stumper.  Just overwrite a file once,
-    %% and it's obvious that PDManifests is a non-empty list and
-    %% indeed has 2-tuples.  And the spec for
-    %% riak_cs_manifest_utils:manifests_to_gc/2 appears correct.
     PDUUIDs = [UUID || {UUID, _} <- PDManifests],
     handle_move_result(MoveResult, RiakObject, Bucket, Key, PDUUIDs, RiakcPid);
 handle_mark_as_pending_delete({error, _Error}=Error, _Bucket, _Key, _UUIDsToMark, _RiakcPid) ->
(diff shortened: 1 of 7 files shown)


kellymclaughlin:
I plan to squash all 4 commits and any that result from code review, but I have left them for now to show the breakdown of the work.

test/riak_cs_gc_d_eqc.erl
```diff
@@ -155,7 +155,7 @@ idle(_S) ->
      {history, {call, ?GCD_MODULE, resume, []}},
      {history, {call, ?GCD_MODULE, set_interval, [infinity]}},
      {{paused, idle}, {call, ?GCD_MODULE, pause, []}},
-     {fetching_next_batch, {call, ?GCD_MODULE, manual_batch, [[testing]]}}
+     {fetching_next_fileset, {call, ?GCD_MODULE, manual_batch, [[testing]]}}
```
kellymclaughlin, line 158: The state transitions have changed a bit since this work began due to the addition of optional 2I pagination for GC.

src/riak_cs_gc.erl
```diff
@@ -179,11 +179,6 @@ handle_mark_as_pending_delete({ok, RiakObject}, Bucket, Key, UUIDsToMark, RiakcP
     Manifests = riak_cs_utils:manifests_from_riak_object(RiakObject),
     PDManifests = riak_cs_manifest_utils:manifests_to_gc(UUIDsToMark, Manifests),
     MoveResult = move_manifests_to_gc_bucket(PDManifests, RiakcPid),
-    %% riak_cs_gc.erl:185: The pattern [{UUID, _} | _] can never match
```
kellymclaughlin, line 182: Mwhahahaha, mystery solved!

kellymclaughlin:
This branch has been freshly rebased from `develop` to pull in some other GC-related changes such as optional 2I pagination for gathering items to GC. 

There is still the issue of confusing and inconsistent terminology (_e.g._ batch, fileset), but I am deferring that to the forthcoming PR to add concurrency to the GC process.

reiddraper:
One thing I've been working through in my head is what happens as you upgrade to this change. I wanted to make sure that as you upgrade, no garbage will be collected _early_. As I've worked it out, it is safe. In fact, one 'leeway' of garbage will be collected _late_ (2x leeway). Here's how I came up with this:

Assume 24-hour leeway. Before upgrading, some garbage is created on day 0, and written to key 1 (lets assume keys with day resolution). Now, upgrade Riak CS. Keys written to day 1 won't be collected until day 2. Which means there will be a 48-hour leeway, but just for those keys written _before_ the upgrade. Keys written after the upgrade will be written to `now()`. So after 24-hours, garbage will go back to being collected after 1x leeway. Does this seem correct?

kellymclaughlin:
> Why are we adding Leeway? Should it be subtracted?

Yes it should. Fixed.

kellymclaughlin:
> Quite nitpicky, but what about turning this into an error-tuple, instead of passing the 'EXIT' to later be pattern matched? Not something I feel strongly about.

Yes, I see your point, but I think I'd rather do that as a specific refactor of that script instead of throwing it in here.

kellymclaughlin:
> Assume 24-hour leeway. Before upgrading, some garbage is created on day 0, and written to key 1 (lets assume keys with day resolution). Now, upgrade Riak CS. Keys written to day 1 won't be collected until day 2. Which means there will be a 48-hour leeway, but just for those keys written before the upgrade. Keys written after the upgrade will be written to now(). So after 24-hours, garbage will go back to being collected after 1x leeway. Does this seem correct?

This does seem correct. A possible workaround is to temporarily set a negative leeway time if that extra interval poses a real problem, but I suspect in most cases it will not be an issue.

reiddraper:
Great work: :+1:

PR #763, opened 2014-01-03
Improve handling of poolboy timeouts during ping requests

kellymclaughlin:
The design of the riak_cs_wm_ping webmachine resource is that it blocks for a
timeout period while attempting to checkout a Riak connection from the poolboy
request_pool. If the timeout expires before a pool connection is checked out
it is designed to attempt to establish a direct connection to Riak and then
attempt the ping.

The reality of what occurs is that the process handling the ping request 
crashes after the timeout expires. This is due to the fact that 
poolboy:checkout calls gen_fsm:sync_send_event with a timeout parameter
specified. The timeout expiration results in a call to exit(timeout) that
causes the request process to crash and return a 500 error to the client.

This change modifies the riak_cs_wm_ping resource to catch the exit on timeout
and return the atom full. This allows the direct connection logic to execute
and either return success to the user or a 503 error indicating the system is
too heavily loaded.

### DIFF
--- a/src/riak_cs_wm_ping.erl
+++ b/src/riak_cs_wm_ping.erl
@@ -40,28 +40,10 @@ init(_Config) ->
     riak_cs_dtrace:dt_wm_entry(?MODULE, <<"init">>),
     {ok, #ping_context{}}.
 
--spec service_available(term(), term()) -> {boolean(), term(), term()}.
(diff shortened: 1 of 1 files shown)


reiddraper:
+1

PR #761, opened 2013-12-27
Refactor/bucket object in state

shino:
Some independent logics fetch bucket object from Riak, bucket policy, bucket ACL
and bucket owner retrieval. In object access, e.g. GET Object, we can/should suppress
extra bucket object fetch.

Refactoring plan is simple. Let object-related webmachine callback
resources keep bucket objects (riak objects) in their state.

This PR is motivated mainly by https://github.com/basho/riak_cs/issues/701 .
Also this will be the basis for https://github.com/basho/riak_cs/issues/723 .

In testing this PR, two things found, both are reproduced at current develop branch.
1. Dialyzer outputs some warnings:
2. make test fails at `riak_cs_get_fsm_eqc`
These may/should be treated separated issues.

1 Dialyzer

```
riak_cs_gc.erl:187: The pattern [{UUID, _} | _] can never match the type []
riak_cs_gc_d.erl:173: The call riak_cs_gc_d:fetch_eligible_manifest_keys(RiakPid::'undefined' | pid(),BatchStart::non_neg_integer(),'undefined') will never return since it differs in the 3rd argument from the success typing arguments: (pid(),non_neg_integer(),binary())
[...]

PR #768, opened 2014-01-07
Bugfix/eqc process leak

shino:
Fix process leak found in #761 
- Stop riak_cs_get_fsm and riak_cs_dummy_reader processes after single command sequence
- Refactoring

Increasing beam.smp's memory size can be seen by, for instance:

```
$ ./rebar eunit skip_deps=true compile_only=true && \
    ERL_LIBS=$ERL_LIBS:deps erl -pa .eunit
> spawn(fun() -> riak_cs_get_fsm_eqc:test(1000) end).
```

### DIFF
--- a/src/riak_cs_get_fsm.erl
+++ b/src/riak_cs_get_fsm.erl
@@ -289,6 +289,8 @@ perhaps_send_to_user(From, #state{got_blocks=Got,
             end
     end.
 
+waiting_chunks(stop, State) ->
+    {stop, normal, State};
 waiting_chunks(timeout, State = #state{got_blocks = Got}) ->
     GotSize = orddict:size(Got),
     _ = lager:debug("starting fetch again with ~p left in queue", [GotSize]),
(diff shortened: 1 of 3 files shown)


reiddraper:
I'm not able to see any leaking processes, seen via: `length(erlang:processes())`. Once the test has been run once, it stays at 60 processes, with and without this change. Am I missing something?

shino:
The word "leak" may be misleading, sorry.
Number of processes continue to grow while test _is running_.

I found growth of memory (RES) of beam.smp and process count when doing:

```
> spawn(fun() -> riak_cs_get_fsm_eqc:test(1000) end).
```

If your machine has more RAM, 2000 or 3000 tests might make the situation more obvious instead of 1000.
In my environment (Linux with 8GB RAM), 1000 tests was sufficient that OOMKiller killed beam.smp
when it reached to about 5GB RES :P

reiddraper:
Aha! Yes, I see that now. And the branch does fix it. I'll take a look at the test refactoring next.

reiddraper:
Great work, :+1:

PR #769, opened 2014-01-07
Add cluster-info command

ksauzz:
`cluster-info` is so helpful for debugging. So I'd like to add it to riak-cs.
And also this commit includes `riak-cs-admin` command like `riak-admin`.
I'm wondering if riak-cs-gc, riak-cs-storage, riak-cs-access and riak-cs-stanchion merge into riak-cs-admin that calls riak_cs_xxxx_console. If it makes sense I can refactor that later. if not, I'll rename these file names.


reiddraper:
+1

ksauzz:
Addressed  #826

PR #784, opened 2014-01-30
Feature/list objects eqc addition (squashed)

shino:
This PR adds two EQC properties for listing all objects.
- Two EQC properties for listing objects (GET Bucket),
  one without delimiter and one with delimiter "/".
  Those properties assert following two are equal:
  1. keys from listing objects by repetitive usage of
     riak_cs_list_objects_fsm_v2
  2. active keys from generated [{State, Key},...]
- When counterexample is found, states and keys are
  written to file `$PWD/.riak_cs_list_objects_fsm_v2_eqc.txt`.
  Each line consists of line number, status and key.
  Example output:
  
  ```
  1,active,0123456/10
  2,scheduled_delete,0123456/1000
  ```
- To generate manifests, generator of whole manifests is avoided
  and generate only states and keys of manifests.
  The property list-all-objects generates several thousands of nested tuples,
  #lfs_manifest_v3 and #acl_v2 in this case.
  But generators of nested tuples seem not so fast.

TODOs remaining (not this PR's TODOs but ones for future)
- prefix of request parameter
- non-default max-keys

### DIFF
--- a/src/riak_cs_list_objects_fsm_v2.erl
+++ b/src/riak_cs_list_objects_fsm_v2.erl
@@ -37,7 +37,8 @@
 -endif.
 
 %% API
--export([start_link/2]).
+-export([start_link/2,
+         start_link/3]).
 
 %% Observability
 -export([]).
@@ -73,7 +74,7 @@
 -record(state, {riakc_pid :: pid(),
                 req :: list_object_request(),
                 reply_ref :: undefined | {pid(), any()},
-                key_multiplier :: float(),
+                fold_objects_batch_size :: pos_integer(),
                 object_list_req_id :: undefined | reference(),
                 reached_end_of_keyspace=false :: boolean(),
                 object_buffer=[] :: list(),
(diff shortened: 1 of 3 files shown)


reiddraper:
> - non-default max-keys

Added in 0cdb828.

reiddraper:
This has uncovered a new Riak CS bug. When a `delimiter` is included, S3 returns an XML node called `NextMarker`, which is to be used by the client for pagination. Riak CS currently does not support this. Instead, we mistakenly assume the client will just use the last key. Further, when a request is made with both a `marker` and `delimiter`, we should be treating the marker as a common delimiter and 'skipping past' it, like we do internally. Once the `NextMarker` XML node is returned, we should be able to update the EQC test to take advantage of it, and continue running the test. We also may need to update _erlcloud_, if it doesn't already support this node.

reiddraper:
Update, _erlcloud_ does not support the `NextMarker` XML node.

reiddraper:
I've fixed the lists:last error in 0d2caedc69aefe13b7aad65671b3775a60164806, by supporting the NextMarker element. The test is still failing, but it's now failing for a different reason. Progress.

reiddraper:
Smallest failing case for `with_prefix` is now something like:

```
1,writing,ZZZZZZZZ/1
2,writing,ZZZZZZZZ/2
3,active,ZZZZZZZZ/3
```

reiddraper:
The bug appears to be related to what @shino initially suspected: we incorrectly skip past a prefix, even if it's deleted. When it's deleted, it won't be included in the result set, and we might be skipping past other keys that fall under that prefix but are not deleted.

reiddraper:
Removing the `skip_past_prefix_and_delimiter` optimization makes the test pass. We can't to keep the optimization, but need some way of not using it if the only object that fits under the delimiter is _not_ in the `active` state.

reiddraper:
Need to clean things up and squash, but I think I've addressed all of the bugs [here](https://github.com/basho/riak_cs/compare/feature%2Flist-objects-eqc-addition-squashed...feature%2Flist-objects-eqc-addition-squashed-reid). In addition to supporting `NextMarker`, I've fixed the `skip_past_prefix_and_delimiter` edge-case here: 3424892d0d26028b5d1f516e1b9d6876bcbadec1.

shino:
> supporting NextMarker

I cherry-picked eqc part of `NextMarker` support and pushed some prefix generator modification.

shino:
At the current HEAD (ed88f35), I confirmed
- eqc tests passes with test number 10,000 each (with/without delimiter) with [fixes](https://github.com/basho/riak_cs/compare/feature%2Flist-objects-eqc-addition-squashed...feature%2Flist-objects-eqc-addition-squashed-reid) and
- without those fixes (partly), eqc tests generates conterexamples we faced so far (original 1003 keys, common prefixes but no keys in result, false prefix skipping).

Everything seems nice. I will +1 to that branch after squash! :joy:

reiddraper:
Some of the NextMarker support is missing in the cherry-picking. I'll add it and then re-order the commits.

shino:
Close in favor of #788

PR #737, opened 2014-01-27
Make fold_objects the default in Riak CS

reiddraper:

PR #785, opened 2014-01-31
Make fold_objects the default in Riak CS, part 2

shino:
Additional changes to #737 
- Change default value in case of no configuration
- Fix riak_test cases to use proper versions


reiddraper:
+1. Thanks for catching this.

PR #790, opened 2014-02-05
add debug log when empty owner in PUT ACL's XML

kuenishi:
in favour of #755, rebased to develop branch and added a log line. This change is too trivial to add test, so I believe few review needed.


shino:
The original bug reporter of #755 says "It looks like it's no longer occurring on Cyberduck 4.4.3, so maybe this issue is not so much worth looking at any more..." at https://github.com/basho/riak_cs/pull/755#issuecomment-34137264 .
Adding new case branch seems somewhat needless for me, doesn't it?

kuenishi:
Agree, lacking `Owner` will be handled correctly even in current code.

Issue #782, opened 2014-01-28
Whitespace in PUT bucket ACL XML payload causes errors


hectcastro:
Below is some `ngrep` output that demonstrates a failure (with whitespace) and a success (without whitespace).

Also, if `ngrep` doesn't appear to be working on OS X 10.9, go [here](https://gist.github.com/jfarcand/8302675).

With whitespace and indentation:

```
T 127.0.0.1:59582 -> 127.0.0.1:8080 [AP]
PUT /mybucket/?acl HTTP/1.1.
User-Agent: fog/1.19.0.
Content-MD5: JLrUS1Nw01RLsz21YDEohA==.
Content-Type: application/json.
Date: Tue, 28 Jan 2014 18:44:49 +0000.
Authorization: AWS MOGMGYGVRHC7OYQHTLYS:6qtIOlQSP7qEPZW2sa7c67HgXsQ=.
Host: joker.dev:8080.
Content-Length: 529.
TE: trailers, deflate, gzip.
Connection: TE, close.
.


T 127.0.0.1:59582 -> 127.0.0.1:8080 [AP]
<AccessControlPolicy>
  <Owner>
    <ID>eb874c6afce06925157eda682f1b3c6eb0f3b983bbee3673ae62f41cce21f6b1</ID>
    <DisplayName>admin</DisplayName>
  </Owner>
  <AccessControlList>
    <Grant>
      <Grantee xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:type="CanonicalUser">
        <ID>eb874c6afce06925157eda682f1b3c6eb0f3b983bbee3673ae62f41cce21f6b1</ID>
        <DisplayName>admin</DisplayName>
      </Grantee>
      <Permission>FULL_CONTROL</Permission>
    </Grant>
  </AccessControlList>
</AccessControlPolicy>

T 127.0.0.1:8080 -> 127.0.0.1:59582 [AP]
HTTP/1.1 500 Internal Server Error.
Server: Riak CS.
Date: Tue, 28 Jan 2014 18:44:49 GMT.
Content-Type: text/html.
Content-Length: 170.
.


T 127.0.0.1:8080 -> 127.0.0.1:59582 [AP]
<html><head><title>500 Internal Server Error</title></head><body><h1>Internal Server Error</h1>The server encountered an error while processing this request</body></html>
```

Without whitespace:

```
T 127.0.0.1:60336 -> 127.0.0.1:8080 [AP]
[...]

PR #793, opened 2014-02-05
Bugfix/gh782 acl xml format

kuenishi:
The root cause for #782 was misuse of `xmerl_scan:string/2`, which doesn't emit spaces, comments between actual xml elements. It leaves them as `#xmlText{}` and `#xmlComment{}` - This pull request commit changes to ignore them at walking around parsed dom tree.
This diffset also includes a small refactoring that wraps all `xmerl_scan:string/2` to one `riak_cs_xml` funcsion to control its usage.

### DIFF
--- a/src/riak_cs_acl_utils.erl
+++ b/src/riak_cs_acl_utils.erl
@@ -499,6 +499,8 @@ process_acl_contents([#xmlElement{content=Content,
         {error, _}=Error ->
             Error
     end;
+process_acl_contents([#xmlComment{} | RestElements], Acl, RiakPid) ->
+    process_acl_contents(RestElements, Acl, RiakPid);
 process_acl_contents([#xmlText{value=" "} | RestElements], Acl, RiakPid) ->
     %% skip normalized space
     process_acl_contents(RestElements, Acl, RiakPid).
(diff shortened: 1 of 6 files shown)


kuenishi:
Or maybe I'd open this against 1.4 branch?

shino:
Agree. > 1.4

reiddraper:
Yes, let's try and get this into 1.4.5.

src/riak_cs_xml.erl
```diff
@@ -42,10 +44,34 @@
 -type internal_node() :: {atom(), [internal_node() | external_node()]} |
                          {atom(), attributes(), [internal_node() | external_node()]}.
 
+%% these types should be defined in xmerl.hrl or xmerl.erl
+%% for now they're defined here for convenience.
+-type xmlElement() :: #xmlElement{}.
+-type xmlText() :: #xmlText{}.
```
reiddraper, line 50: `xmlElement` and `xmlText` should be exported types (thanks dialyzer)

src/riak_cs_acl_utils.erl
```diff
@@ -471,35 +473,40 @@ name_for_canonical(CanonicalId, RiakPid) ->
...
     case UpdAclRes of
         {ok, UpdAcl} ->
             process_acl_contents(RestElements, UpdAcl, RiakPid);
         {error, _}=Error ->
             Error
-    end.
+    end;
+process_acl_contents([#xmlComment{} | RestElements], Acl, RiakPid) ->
+    process_acl_contents(RestElements, Acl, RiakPid);
+process_acl_contents([#xmlText{value=" "} | RestElements], Acl, RiakPid) ->
```
reiddraper, line 504: should we be skipping newlines too somewhere? Or is it only spaces that are causing trouble?

kuenishi, line 504: This is what `{space, normalize}` does. Every space-like characters (newlines and tabs) are folded into single space " " by `erl_scan:string/2`. So we care only single space.

kuenishi:
Reopened an issue against release/1.4 => #795.

src/riak_cs_xml.erl
kuenishi, line 50: This is fixed in #795

src/riak_cs_acl_utils.erl
reiddraper, line 504: Cool, thanks.

Issue #787, opened 2014-02-03
Failed to upload in the case I use multipart upload


KOBA789:
Using the filename including multibyte chars, it fails to upload it.

I found the content of Initiate Multipart Upload XML is broken like this:

```
<?xml version="1.0" encoding="UTF-8"?><InitiateMultipartUploadResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Bucket>isidai</Bucket><Key>ãã¹ãã ã¬/æ¥æ¬èªãã¹ã.m4v</Key><UploadId>d16oiXpDTBCmJhsZpDoRWA==</UploadId></InitiateMultipartUploadResult>
```

The value of `Key` element is corrupted.

Some S3 compatible clients use this element value, so it failed to upload files containing multibyte chars.

The cause is [here](https://github.com/basho/riak_cs/blob/develop/src/riak_cs_wm_object_upload.erl#L110). 

```
    case riak_cs_mp_utils:initiate_multipart_upload(Bucket, list_to_binary(Key),
                                                    ContentType, User, Opts,
                                                    RiakcPid) of
        {ok, UploadId} ->
            XmlDoc = {'InitiateMultipartUploadResult',
                       [{'xmlns', "http://s3.amazonaws.com/doc/2006-03-01/"}],
                       [
                        {'Bucket', [binary_to_list(Bucket)]},
                        {'Key', [Key]},  %% !!HERE!!
                        {'UploadId', [binary_to_list(base64url:encode(UploadId))]}
                       ]
                     },
            Body = riak_cs_xml:export_xml([XmlDoc]),
            RD2 = wrq:set_resp_body(Body, RD),
            {true, RD2, Ctx};
```

In L.110, `Key` is not a binary but a list. Because `riak_cs_xml:export_xml` encodes lists into utf8, `Key` will be broken, but `Key` should be a binary. Binaries are not encoded by `riak_cs_xml:export_xml`.

isidai:
Related to [this issue](https://github.com/basho/riak_cs/issues/628)
It will resolve this problem.

Cyberduck is effected by this problem.
Cyberduck uses XML's value of response when request last finalizing POST.
so the last request's URL different from the first request and part object requests.

```
[...]

Issue #628, opened 2013-07-29
Support multibyte username on xml response.


ksauzz:
List users and show user API doesn't support multibyte on xml response. So name is garbled.
### create user

```
[vagrant@localhost ~]$ nkf --utf8 user.json
{"name": "ほげほげ", "email":"test@example.com"}
[vagrant@localhost ~]$ curl -i http://localhost:8080/riak-cs/user -H "Content-Type: application/json" -d @user.json
HTTP/1.1 201 Created
Vary: Accept
Server: Riak CS
Location: http://localhost:8080/riak-cs/user
Date: Mon, 29 Jul 2013 05:58:22 GMT
Content-Type: application/json
Content-Length: 263

{"email":"test@example.com","display_name":"test","name":"\u307b\u3052\u307b\u3052","key_id":"X9Z01FRHMO1YJWZT-1B7","key_secret":"ueFnnV1pwc1VNg_vHvnLKC6Ah1nw888rO90PnQ==","id":"6d0e53854eed0b7f3df6605aac08723855a89091bc84848f9346b0cfd39cd08b","status":"enabled"}[vagrant@localhost ~]$
```
#### XML response

```
[vagrant@localhost ~]$ ./s3curl.pl --id admin -- -x localhost:8080 http://riak-cs.s3.amazonaws.com/users

--OP3SKSJCRRKG6UJN3NZ50NKLFIJ47Y3
Content-Type: application/xml

<?xml version="1.0" encoding="UTF-8"?><Users><User><Email>test@example.com</Email><DisplayName>test</DisplayName><Name>ã»ãã»ã</Name><KeyId>X9Z01FRHMO1YJWZT-1B7</KeyId><KeySecret>ueFnnV1pwc1VNg_vHvnLKC6Ah1nw888rO90PnQ==</KeySecret><Id>6d0e53854eed0b7f3df6605aac08723855a89091bc84848f9346b0cfd39cd08b</Id><Status>enabled</Status></User></Users>
--OP3SKSJCRRKG6UJN3NZ50NKLFIJ47Y3
Content-Type: application/xml

[...]

PR #791, opened 2014-02-05
Improve access log handler to handle HEAD method.

ksauzz:
This fix prevents to include BytesOut element on Head Object(KeyStat) request the same as Head Bucket(BucketStat).

Requires the latest erlcloud commit: https://github.com/basho/erlcloud/pull/12/commits
Fix #789

Issue #789, opened 2014-02-05
KeyStat in access usage should include only a count element.


ksauzz:
It seems to be desirable that KeyStat in access usage has no BytesOut element like BucketStat.
### BucketStat

```
 "BucketStat": {
           "Count": 1
  },
```
### KeyStat

```
"KeyStat": {
          "BytesOut": 1326,
           "Count": 1
},
```

ksauzz:
Addressed #791.

PR #794, opened 2014-02-05
Bugfix/non ascii in xml

shino:
This PR addresses two issues #787 and #628.

The plan to fix them and possible other XML-with-multibyte(non-ASCII)-character
issues are as follows:
- `riak_cs_xml:format_value(Val)` treats conversion between lists/binaries and
  Unicode strings
  (xmerl accepts only Unicode strings but not binary or latin-1-ish strings).
- In the function clause `when is_list(Val)`, use list_to_binary and call
  `format_value` again with converted binary.
- In the function clause `when is_binary(Val)` use `unicode:characters_to_list`
  to produce Unicode strings.
- Remove all other `unicode:characters_to_list` around XML output creation.

Some misc notes:
- When we take into account Unicode, lists created by `binary_to_list` should not
  be considered as "strings". This conversion is NOT Unicode-aware, so lists and
  binaries have byte-wise correspondence.
- On the other hand, `unicode:characters_to_list` is Unicode-aware, so converted
  lists are not necessarily `[0..255]`. Each element of a converted list is
  correspond to Unicode codepoint (not so precise but it's almost a character).
- The talk [1] is good reference for Latin-1 and Unicode in Erlang.
  Codepoints, encodings, string literals, binaries and more.

```
> A = <<16#E3, 16#81, 16#82>>. % A binary of Japanese Hiragana "A" in UTF-8
<<227,129,130>>
> binary_to_list(A).
[227,129,130]                  % Just byte-wise
> unicode:characters_to_list(A).
[12354]                        % A list of single element (single character)
```

[1] http://www.erlang-factory.com/conference/ErlangUserConference2013/speakers/PatrikNyblom

### DIFF
--- a/riak_test/tests/access_stats_test.erl
+++ b/riak_test/tests/access_stats_test.erl
@@ -26,6 +26,7 @@
 -export([confirm/0]).
 
 -include_lib("erlcloud/include/erlcloud_aws.hrl").
+-include_lib("xmerl/include/xmerl.hrl").
 -include_lib("eunit/include/eunit.hrl").
 
 -define(BUCKET, "access-stats-test-1").
@@ -37,7 +38,8 @@ confirm() ->
     {UserConfig, {_RiakNodes, _CSNodes, _Stanchion}} = rtcs:setup(4, Config),
     {Begin, End} = generate_some_accesses(UserConfig),
     flush_access_stats(),
-    assert_access_stats(UserConfig, {Begin, End}),
+    assert_access_stats(json, UserConfig, {Begin, End}),
+    assert_access_stats(xml, UserConfig, {Begin, End}),
     pass.
 
 generate_some_accesses(UserConfig) ->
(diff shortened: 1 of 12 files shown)


shino:
memo: This PR has conflict with #791. (almost trivial to merge)

reiddraper:
All tests passing. Reviewing the code now.

reiddraper:
Double-checking my Erlang and unicode: is this true? For-all utf-8 encoded binaries, `B`:

``` erlang
B =:= list_to_binary(binary_to_list(B)).
```

riak_test/tests/access_stats_test.erl
```diff
@@ -66,46 +68,69 @@ flush_access_stats() ->
...
-                                      end, ListOfNodeStats),
-    rtcs:json_get(<<"Samples">>, NodeStats).
+    NodeBin = list_to_binary(Node),
+    [NodeStats | _] = lists:dropwhile(
+                        fun(NodeStats) ->
+                                rtcs:json_get(<<"Node">>, NodeStats) =/= NodeBin
+                        end, ListOfNodeStats),
+    rtcs:json_get(<<"Samples">>, NodeStats);
+node_samples_from_content(xml, Node, Content) ->
+    {Usage, _Rest} = xmerl_scan:string(unicode:characters_to_list(Content, utf8)),
```
reiddraper, line 108: In a lot of the other code, we simply `binary_to_list/1` UTF-8 encoded binaries into lists. Does the `xmerl` API demand we give it a list of character points instead?

reiddraper:
I can't think of 'a better way', but one thing that is confusing (and already exists in the codebase) is the mixed use of binaries (presumably UTF-8 encoded), lists (constructed from `binary_to_list/1`) and lists of character points. Is there any way to be more consistent, or are we doing the best we can already?

riak_test/tests/user_test.erl
```diff
@@ -34,38 +34,52 @@ confirm() ->
...
     pass.
 
-user_listing_json_test_case(UserCreds, UserConfig, Node) ->
-    user_listing_test(UserCreds, UserConfig, Node, ?JSON).
+japanese_aiueo() ->
+    %% To avoid dependency on source code encoding, create list from chars.
+    %% These five numbers represents "あいうえお" (A-I-U-E-O in Japanese).
+    %% unicode:characters_to_binary([12354,12356,12358,12360,12362]).
+    Chars = [12354,12356,12358,12360,12362],
+    binary_to_list(unicode:characters_to_binary(Chars)).
```
reiddraper, line 61: Personal note: this is equal to `<<"あいうえお"/utf8>>`.

reiddraper, line 61: Minus the `binary_to_list` part.

src/riak_cs_xml.erl
```diff
@@ -222,7 +245,7 @@ format_value(Val) when is_binary(Val) ->
 format_value(Val) when is_integer(Val) ->
     integer_to_list(Val);
```
reiddraper, line 246: nice, this is much cleaner than scattering `integer_to_list` around the codebase.

shino:
> Double-checking my Erlang and unicode: is this true? For-all utf-8 encoded binaries, B:

True (or I believe it's true).

riak_test/tests/user_test.erl
shino, line 61: I must use it! Will fix.

shino:
> Does the xmerl API demand we give it a list of character points instead?

xmerl's document says it accepts binary as element of deep list (direct binary is not accepted).
Sample code of list of code points and binary > https://gist.github.com/shino/3c5ab84980dde4e7ab24
It seems xmerl converts a binary to byte-wise list (same as `binary_to_list`).

riak_test/tests/user_test.erl
reiddraper, line 61: Well, as your comment mentions, your code doesn't depend on UTF-8 encoding in the actual file, which might be desirable. That being said, I assume most (all?) of us are using text-editors that use UTF-8.

shino:
One trial branch : https://github.com/basho/riak_cs/compare/release%2F1.4...bugfix%2Fnon-ascii-in-xml-2
Difference from `bugfix/non-ascii-in-xml` is only this commit: https://github.com/basho/riak_cs/commit/6a13a83ee74e0d478ed563b93c84860999d1d9e9

No Unicode code points appears in generating XML.
The trick behind is that  `xmerl:export_simple` convert list of binaries to b2l'd lists.
(I wish it accepted direct binary()).

Anyway, XML generation is now encapsulated in `riak_cs_xml`, so switching
these two ways (or another?) is rather easy.

riak_test/tests/user_test.erl
shino, line 61: Yeah, editor's default encoding is UTF-8 :)
Another file encoding issue is about compiler. Erlang compiler thinks files are Latin-1 encoded (until R15, also in R16 but can be overwritten magic comments on files).

`"あ"` is three-byte in utf-8 encoded file, compiler sees it as if latin-1 and convert to three latin-1 characters. Then applies `/utf8` conversion so it results a binary of six bytes (oops).

shino:
Because of conflict with #789, rebased and created new PR #796 .

shino:
(reply to myself)

Avoiding `unicode` module at `bugfix/non-ascii-in-xml-2` seems not good for me...
`unicode` can be avoided in generating XML by `xmerl:export_simple` but parsing XML
by `xmerl_scan:string` returns text value of Unicode code points. So they are not symmetric. :see_no_evil:

shino:
Continue in #796 or #807

shino:
memo: #807  is finally merged to release branch `release/1.4`.

PR #795, opened 2014-02-06
Handle well-indented XML including comments

kuenishi:
The root cause for #782 was misuse of xmerl_scan:string/2, which doesn't emit spaces, comments between actual xml elements. It leaves them as #xmlText{} and #xmlComment{} - This commit changes to ignore them at walking around parsed dom tree. This diffset also includes a small refactoring that wraps all xmerl_scan:string/2 to one riak_cs_xml funcsion to control its usage.

This also adds regression test as simple eunit test case.

backport of #793

### DIFF
--- a/test/cs782_regression_test.erl
+++ b/test/cs782_regression_test.erl
@@ -1,0 +1,51 @@
+%% ---------------------------------------------------------------------
+%%
+%% Copyright (c) 2007-2014 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% ---------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


test/cs782_regression_test.erl
```diff
@@ -0,0 +1,51 @@
...
+        "    <Grant>"
+        "      <Grantee xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xsi:type=\"CanonicalUser\">"
+        "        <ID>eb874c6afce06925157eda682f1b3c6eb0f3b983bbee3673ae62f41cce21f6b1</ID>"
+        "        <DisplayName>admin</DisplayName>"
+        "      </Grantee>"
+        "      <Permission>FULL_CONTROL</Permission>"
+        "    </Grant>"
+        "  </AccessControlList>"
+        "</AccessControlPolicy>",
+    Xml.
```
reiddraper, line 45: Really nitpicky, but why assign this to `Xml`, instead of just returning the string?

reiddraper:
Tests, code and manual testing all pass. +1

test/cs782_regression_test.erl
kuenishi, line 45: Without binding `Xml` the indentation by erlang-mode.el becomes like this:

``` erlang
    "<!-- comments here --> <AccessControlPolicy> <!-- comments here -->"
        "  <Owner> <!-- blah blah blah -->"
        "    <ID>eb874c6afce06925157eda682f1b3c6eb0f3b983bbee3673ae62f41cce21f6b1</ID>"
        " <!-- c -->   <DisplayName>admin</DisplayName> <!--"
        " \n --> </Owner>"
```

Also it's easy to put some `~p` to do some formatting like `io_lib:format(Xml, [ Blah, Blah, ... ])` for future improvement. Yes, I was trying that.

PR #796, opened 2014-02-06
Bugfix/non ascii in xml (rebased)

shino:
This PR addresses two issues #787 and #628 (rebased version of #794 ).

The plan to fix them and possible other XML-with-multibyte(non-ASCII)-character
issues are as follows:
- `riak_cs_xml:format_value(Val)` treats conversion between lists/binaries and
  Unicode strings
  (xmerl accepts only Unicode strings but not binary or latin-1-ish strings).
- In the function clause `when is_list(Val)`, use list_to_binary and call
  `format_value` again with converted binary.
- In the function clause `when is_binary(Val)` use `unicode:characters_to_list`
  to produce Unicode strings.
- Remove all other `unicode:characters_to_list` around XML output creation.

Some misc notes:
- When we take into account Unicode, lists created by `binary_to_list` should not
  be considered as "strings". This conversion is NOT Unicode-aware, so lists and
  binaries have byte-wise correspondence.
- On the other hand, `unicode:characters_to_list` is Unicode-aware, so converted
  lists are not necessarily `[0..255]`. Each element of a converted list is
  correspond to Unicode codepoint (not so precise but it's almost a character).
- The talk [1] is good reference for Latin-1 and Unicode in Erlang.
  Codepoints, encodings, string literals, binaries and more.

```
> A = <<16#E3, 16#81, 16#82>>. % A binary of Japanese Hiragana "A" in UTF-8
<<227,129,130>>
> binary_to_list(A).
[227,129,130]                  % Just byte-wise
> unicode:characters_to_list(A).
[12354]                        % A list of single element (single character)
```

[1] http://www.erlang-factory.com/conference/ErlangUserConference2013/speakers/PatrikNyblom

### DIFF
--- a/client_tests/python/boto_tests/boto_test.py
+++ b/client_tests/python/boto_tests/boto_test.py
@@ -54,8 +54,8 @@ def upload_multipart(bucket, key_name, parts_list, metadata={}):
     upload = bucket.initiate_multipart_upload(key_name, metadata=metadata)
     for index, val in enumerate(parts_list):
         upload.upload_part_from_file(val, index + 1)
-    upload.complete_upload()
-    return upload
+    result = upload.complete_upload()
+    return upload, result
 
 class S3ApiVerificationTestBase(unittest.TestCase):
     host="127.0.0.1"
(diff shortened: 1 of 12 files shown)


shino:
Rebased (again) and force pushed in order to clear cluttered diff.

reiddraper:
Should this be closed in favor of #807?

shino:
Thank you for review! Bug fixed by #807. Close this PR without merge.

PR #801, opened 2014-02-07
Default Make target just compiles

reiddraper:
Including tools.mk at the top of the Makefile changed the default
target. This resets the default target back to `all`.


kuenishi:
+1

PR #803, opened 2014-02-10
Add timeouts to ensure that the list_objects FSM eventually stops.

evanmcc:
Since this FSM has no timeout and the remote index FSM has no timeout
by default, we can get stuck with both FSMs partially completed and
holding on to a lot of memory.

This fix adds a timeout to the remote FSM so it should terminate in a
timely fashion, but adds a timeout to the connection request, just in
case the other end stops talking to us for some reason.


andrewjstone:
@evanmcc Just realized it, but this PR should go to release/1.4

andrewjstone:
superseded by #806

Issue #802, opened 2014-02-08
List objects FSM v2 has poor performance with prefix requests.


reiddraper:
The list objects v2 FSM does not use the `prefix` field in the request to skip past the lesser (lexicographically) set of results. I haven't dug in too much, but it may also not stop once a key has been seen that is _greater_ than the prefix. It should. [The code for constructing the start key](https://github.com/basho/riak_cs/blob/develop/src/riak_cs_list_objects_fsm_v2.erl#L366-L385) should take the prefix into account.

reiddraper:
@shino Any thoughts on how to test this with the EQC test? The bug was _not_ a correctness issue, but a performance/efficiency one.

kuenishi:
Closing as #804 merged. Nice catch.

PR #804, opened 2014-02-11
Performance optimization for object listing requests using a prefix

kellymclaughlin:
- Use prefix in start key calculation when appropriate to avoid
  iterating over unnecessary keys that lexicographically sort less
  than the prefix. The make_start_key_from_marker function in
  riak_cs_list_objects_fsm_v2 is renamed to
  make_start_key_from_marker_and_prefix.
- Terminate processing early when a prefix is specified and a key is
  reached that lexicographically sorts greater than the prefix, but
  does not begin with the prefix. The results come back in sorted key
  order so it can be assumed that once any keys that begin with the
  prefix are processed it is appropriate to cease processing without
  requesting any more data. This is accomplished by the addition of
  the reached_end-of_keyspace/4 function to the
  riak_cs_list_objects_fsm_v2 module.
- Silence extraneous console logging in riak_cs_list_objects_fsm_v2_eqc test

### DIFF
--- a/src/riak_cs_list_objects_fsm_v2.erl
+++ b/src/riak_cs_list_objects_fsm_v2.erl
@@ -217,14 +217,40 @@ handle_done(State=#state{object_buffer=ObjectBuffer,
 
     {NewManis, NewPrefixes} =
     riak_cs_list_objects_utils:filter_prefix_keys(ObjectPrefixTuple, Request),
-    ReachedEnd = ObjectBufferLength < NumKeysRequested,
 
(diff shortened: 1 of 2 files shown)


andrewjstone:
- [x] riak_test
- [x] eunit + eqc
- [x] dialyzer
- [x] xref

andrewjstone:
Nice work @kellymclaughlin. :+1:

shino:
Is this #802 related?

andrewjstone:
Yep. It's the fix for that.

reiddraper:
Thanks for taking care of this, great work!

PR #806, opened 2014-02-11
Add timeouts to ensure that the list_objects FSM eventually stops. -- 1.4

evanmcc:
Since this FSM has no timeout and the remote index FSM has no timeout
by default, we can get stuck with both FSMs partially completed and
holding on to a lot of memory.

This fix adds a timeout to the remote FSM so it should terminate in a
timely fashion, but adds a timeout to the connection request, just in
case the other end stops talking to us for some reason.


evanmcc:
release/1.4 version of #803

andrewjstone:
- [x] eunit + eqc
- [x] dialyzer
- [x] xref
- [x] riak_test

andrewjstone:
:+1:

reiddraper:
Why are we adding a call timeout? The call is always local.

reiddraper:
Unless we suspect a bug in the riakc_pb_socket gen_server, I don't see why a TCP timeout isn't sufficient.

evanmcc:
I don't understand.  A TCP timeout can be several hours.  The call timeout was an overabundance of caution (since the default call timeout has been overridden to be infinity) to make sure that we'd have a timely halt in case the index_fsm stopped in some way without sending a message, since the pb code doesn't seem to monitor the FSM.

reiddraper:
> I don't understand. A TCP timeout can be several hours.

Sorry I was unclear, I was remembering that we had some sort of timeout on reading bytes from the socket. We don't do this with gen_tcp:recv/3, but we do something similar. When a request is queued up, we [create a timer](https://github.com/basho/riak-erlang-client/blob/1.4.1/src/riakc_pb_socket.erl#L1555-L1558) that is used to [disconnect us from the socket if the timer is received before data](https://github.com/basho/riak-erlang-client/blob/1.4.1/src/riakc_pb_socket.erl#L1059-L1070). The problem is that the `cs_bucket_fold` [hardcodes this timeout to infinity](https://github.com/basho/riak-erlang-client/blob/1.4.1/src/riakc_pb_socket.erl#L904). Therefore, we will never time out based on not receiving data from the socket. Even if you pass in a value of 1ms, it's hardcoded to infinity. This is a bug.

> The call timeout was an overabundance of caution (since the default call timeout has been overridden to be infinity) to make sure that we'd have a timely halt in case the index_fsm stopped in some way without sending a message, since the pb code doesn't seem to monitor the FSM.

Since there is a facility for timing out on not reading data from the socket, I'd argue this is preferable to using a gen_server:call timeout. Further, the gen_server call timeout exception is not handled in this changeset. The only reason I can think of that we'd like to use the call timeout is if we're allowing requests to be queued to `riakc_pb_socket`, which we're not. A single process owns the pid at a time, and makes requests in serial.

evanmcc:
You're right, it should be removed, and `cs_bucket_fold` rewritten.  I was
confused by its variance with the rest of the api (which is not high on
consistency in any case), which typically hard-codes the gen-server timeout
to infinity rather than the request timeout.

For example, get/4:

```
get(Pid, Bucket, Key, Options, Timeout)
->
    Req = get_options(Options, #rpbgetreq{bucket = Bucket, key =
Key}),
    gen_server:call(Pid, {req, Req, Timeout},
infinity).

```

If I were do a PR for the socket changes, what branch would I be targeting?

reiddraper:
> You're right, it should be removed, and `cs_bucket_fold` rewritten.  I was
> confused by its variance with the rest of the api (which is not high on
> consistency in any case), which typically hard-codes the gen-server timeout
> to infinity rather than the request timeout.

Yeah, no doubt there is some cruft in here. I also stumbled across [this oddity](https://github.com/basho/riak-erlang-client/issues/155) while reading the code yesterday.

> If I were do a PR for the socket changes, what branch would I be targeting?

Looks like both the 1.4.1 and 1.4.2 tag are contained in the [1.4 branch](https://github.com/basho/riak-erlang-client/tree/1.4).

evanmcc:
Yes, re 155, that seems to be a common pattern in our streaming APIs for
some reason.  That was how the coverage FSM timeout worked as well, before
I changed it when working on streaming buckets, etc.

I'll try to get those PRs out today.  My plan is to add a `request_timeout`
option (leaving `call_timeout` in case someone has come to rely on it).
Change `call_timeout` to `request_timeout` on the CS side.  Should I fix
155 while I am in there, or, pace the last time I changed that behavior and
irritated existing users, leave it as it is?  I am in favor of
deterministic timeouts, but some users will likely have come to rely on the
older behavior.

reiddraper:
> I'll try to get those PRs out today.  My plan is to add a `request_timeout`
> option (leaving `call_timeout` in case someone has come to rely on it).
> Change `call_timeout` to `request_timeout` on the CS side.  Should I fix
> 155 while I am in there, or, pace the last time I changed that behavior and
> irritated existing users, leave it as it is?  I am in favor of
> deterministic timeouts, but some users will likely have come to rely on the
> older behavior.

My vote would be to leave basho/riak-erlang-client#155 as-is for now. Changing [this line](https://github.com/basho/riak-erlang-client/blob/1.4.1/src/riakc_pb_socket.erl#L904):

``` diff
diff --git a/src/riakc_pb_socket.erl b/src/riakc_pb_socket.erl
index 6a05965..20d69a4 100644
--- a/src/riakc_pb_socket.erl
+++ b/src/riakc_pb_socket.erl
@@ -901,7 +901,7 @@ cs_bucket_fold(Pid, Bucket, Opts) when is_pid(Pid), is_binary(Bucket), is_list(O
                           continuation=Continuation,
                           timeout=Timeout},
     ReqId = mk_reqid(),
-    Call = {req, Req, infinity, {ReqId, self()}},
+    Call = {req, Req, Timeout, {ReqId, self()}},
     gen_server:call(Pid, Call, CallTimeout).

 %% @doc Return the default timeout for an operation if none is provided.
```

Would be enough, I think. And then reverting just the `call_timeout` option from the list objects v2 fsm.

evanmcc:
You don't see a need for a separation between the request timeout and the FSM timeout?

reiddraper:
> You don't see a need for a separation between the request timeout and the FSM timeout?

You mean as described in basho/riak-erlang-client#155? If so, I think it'd be nice, but I'd like to wait to hear back from @jonmeredith (the original author), and likely tackle it as a separate issue, where _all_ requests would be fixed at once.

PR #798, opened 2014-02-20
The history of bucket activity for a user is never pruned

reiddraper:
The [records](https://github.com/basho/riak_cs/blob/develop/include/riak_cs.hrl#L50) stored for Riak CS users maintain a list of [bucket records](https://github.com/basho/riak_cs/blob/develop/include/riak_cs.hrl#L67). This list of buckets provides a causal history of bucket creation and deletion for a user and allows us to select a proper resolution when the user record has sibling values that contain conflicting views of the user's bucket ownership. 

The problem is that this history is never pruned. If a user creates and deletes many different buckets then the set continuously grows. In the case when there are no siblings for the user record this poses little problem; however, in the case of siblings this has a negative impact on latencies because for each the lists of buckets for each sibling value must be compared and resolved. As the number of siblings and buckets contained in those siblings grows so does the processing time required to do the resolution. 

The reason this really becomes an issue is that there are a very few cases involving explicit API actions that we update the user record (_e.g._ updating the user's status). As a result if siblings are generated for a user record the bucket resolution happens for every API request by that user until the record is updated.


reiddraper:
It turns out [we do have code to prune deleted buckets](https://github.com/basho/riak_cs/blob/1.4.4/src/riak_cs_utils.erl#L1265-L1272). This happens if the `KeepDeleted` argument to [resolve_buckets](https://github.com/basho/riak_cs/blob/1.4.4/src/riak_cs_utils.erl#L1378) is `false`, which _should_ happen if we were able to read the user record from all primary vnodes. However, [fetch_user](https://github.com/basho/riak_cs/blob/1.4.4/src/riak_cs_utils.erl#L1286) returns [true](https://github.com/basho/riak_cs/blob/1.4.4/src/riak_cs_utils.erl#L1290) in this case. This seems to be a bug.

reiddraper:
It's important to note, however, that the user will only be saved back to Riak when the user record is actually edited (not just the buckets being pruned). However, this will happen as new buckets are created.

reiddraper:
Ugh, made this against master instead of release/1.4. And of course github has no way of undoing. Closing the PR and opening a new one against the correct branch.

Issue #799, opened 2014-02-07
Skip deleted buckets in storage calculation to reduce execution time


ksauzz:
Storage calculation seems to run on deleted buckets. To reduce its execution time on large data in Riak, deleted buckets should be filtered out on this batch.

reiddraper:
+1

kuenishi:
I think this is fixed via #812.

ksauzz:
Exactly.

PR #812, opened 2014-02-20
Bugfix: prune user buckets

reiddraper:
Fixes #798.

### DIFF
--- a/src/riak_cs_config.erl
+++ b/src/riak_cs_config.erl
@@ -46,7 +46,9 @@
          use_t2b_compression/0,
          trust_x_forwarded_for/0,
          gc_key_suffix_max/0,
-         set_gc_key_suffix_max/1
+         set_gc_key_suffix_max/1,
+         user_buckets_prune_time/0,
+         set_user_buckets_prune_time/1
         ]).
 
 %% OpenStack config
@@ -262,6 +264,16 @@ gc_key_suffix_max() ->
 set_gc_key_suffix_max(MaxValue) when is_integer(MaxValue) andalso MaxValue > 0 ->
     application:set_env(riak_cs, gc_key_suffix_max, MaxValue);
 set_gc_key_suffix_max(_MaxValue) ->
+    {error, invalid_value}.
+
+-spec user_buckets_prune_time() -> pos_integer().
+user_buckets_prune_time() ->
+    get_env(riak_cs, user_buckets_prune_time, ?USER_BUCKETS_PRUNE_TIME).
+
+-spec set_user_buckets_prune_time(integer()) -> ok | {error, invalid_value}.
+set_user_buckets_prune_time(PruneTime) when is_integer(PruneTime) andalso PruneTime > 0 ->
+    application:set_env(riak_cs, user_buckets_prune_time, PruneTime);
+set_user_buckets_prune_time(_PruneTime) ->
     {error, invalid_value}.
 
 %% ===================================================================
--- a/src/riak_cs_config.erl
+++ b/src/riak_cs_config.erl
@@ -270,7 +270,7 @@ set_gc_key_suffix_max(_MaxValue) ->
 user_buckets_prune_time() ->
     get_env(riak_cs, user_buckets_prune_time, ?USER_BUCKETS_PRUNE_TIME).
 
--spec set_user_buckets_prune_time(integer()) -> ok | {error, invalid_value}.
+-spec set_user_buckets_prune_time(pos_integer()) -> ok | {error, invalid_value}.
 set_user_buckets_prune_time(PruneTime) when is_integer(PruneTime) andalso PruneTime > 0 ->
     application:set_env(riak_cs, user_buckets_prune_time, PruneTime);
 set_user_buckets_prune_time(_PruneTime) ->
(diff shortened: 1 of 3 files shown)


kuenishi:
- [x] r_t
- [x] test / pulse
- [x] xref
- [x] dialyzer
- [x] review

src/riak_cs_config.erl
```diff
@@ -264,6 +266,16 @@ set_gc_key_suffix_max(MaxValue) when is_integer(MaxValue) andalso MaxValue > 0 -
 set_gc_key_suffix_max(_MaxValue) ->
     {error, invalid_value}.
 
+-spec user_buckets_prune_time() -> pos_integer().
+user_buckets_prune_time() ->
+    get_env(riak_cs, user_buckets_prune_time, ?USER_BUCKETS_PRUNE_TIME).
+
+-spec set_user_buckets_prune_time(integer()) -> ok | {error, invalid_value}.
```
kuenishi: `pos_integer()` instead of `integer()` like above?

src/riak_cs_utils.erl
```diff
@@ -1287,12 +1290,12 @@ fetch_user(Key, RiakPid) ->
     StrongOptions = [{r, all}, {pr, all}, {notfound_ok, false}],
     case riakc_pb_socket:get(RiakPid, ?USER_BUCKET, Key, StrongOptions) of
         {ok, Obj} ->
-            {ok, {Obj, true}};
+            {ok, {Obj, false}};
```
kuenishi: I would suggest mentioning the meaning of this `boolean()` as it's easy to get complicated which is which. Probably, my understanding is -

``` erlang
%% bucket deletion indicators can be cleaned up
```

src/riak_cs_utils.erl
```diff
@@ -1287,12 +1290,12 @@ fetch_user(Key, RiakPid) ->
...
         {error, _} ->
             WeakOptions = [{r, quorum}, {pr, one}, {notfound_ok, false}],
             case riakc_pb_socket:get(RiakPid, ?USER_BUCKET, Key, WeakOptions) of
                 {ok, Obj} ->
-                    {ok, {Obj, false}};
+                    {ok, {Obj, true}};
```
kuenishi: same as above:

``` erlang
%% bucket deletion indicatiors must not be cleaned up
```

kuenishi:
Also I would suggest improving spec of `fetch_user/1` to: 

``` erlang
-spec fetch_user(KeyId::binary(), pid()) ->
                        {ok, {riakc_obj(), ShouldKeepDeletedBuckets::boolean()}} |
                        {error, term()}.
```

For clearer understanding and preventing misuse of boolean.

kuenishi:
Looks like this also fixes #799. `riak_cs_storage:sum_user/2` also uses `get_user`.

kuenishi:
+1 to merge regardless of above nitpick comments. Also confirmed that creating bucket after deleting another bucket removes old `deleted` bucket.

Eventually we'd add riak_test about this?

src/riak_cs_utils.erl
```diff
@@ -1263,13 +1264,15 @@ bucket_sorter(?RCS_BUCKET{name=Bucket1},
...
+%% is deleted and the action occurred over the configurable
+%% maximum prune-time.
 -spec cleanup_bucket(cs_bucket()) -> boolean().
 cleanup_bucket(?RCS_BUCKET{last_action=created}) ->
     false;
 cleanup_bucket(?RCS_BUCKET{last_action=deleted,
                             modification_time=ModTime}) ->
-    timer:now_diff(os:timestamp(), ModTime) > 86400.
+    timer:now_diff(os:timestamp(), ModTime) >
+    riak_cs_config:user_buckets_prune_time().
```
shino, line 1279: Sorry for stepping in and I don't execute this fix ...
`timer:now_diff/2` returns time interval in micro seconds and `riak_cs_config:user_buckets_prune_time()` seems to be represented by seconds. Any mismatch??

src/riak_cs_config.erl
reiddraper: Addressed in 908f373f445089f45f7952e51decd91cab3bb05c.

src/riak_cs_utils.erl
reiddraper: Addressed in 908f373f445089f45f7952e51decd91cab3bb05c.

src/riak_cs_utils.erl
reiddraper: Addressed in 908f373f445089f45f7952e51decd91cab3bb05c.

reiddraper:
- Docs/specs addressed in 908f373f445089f45f7952e51decd91cab3bb05c.
- Microsecond mixup addressed in 3ffac5d979d73cf052bfcb97ce102adf596d0ed4.

src/riak_cs_utils.erl
reiddraper, line 1279: Nice catch! Fixed in 3ffac5d979d73cf052bfcb97ce102adf596d0ed4.

reiddraper:
I'm re-running all tests now.

reiddraper:
> Eventually we'd add riak_test about this?

Yes. I'd like to try and think of a way of testing it that doesn't resort to `timer:sleep/1`, but maybe that's what we have to do?

andrewjstone:
code changes look good. dialyzer is clean regarding changes. tests running now.

andrewjstone:
:+1:

kuenishi:
+1, yay!

PR #816, opened 2014-02-26
Multi-cluster support

shino:
# [WIP] DO NOT MERGE

needs basho/stanchion#76

This PR is for early-stage review.
The branch may be force-pushed because of rebase/ammend/etc.

---

Currently all Riak CS's internal data types are stored into single
Riak cluster. The idea is to distribute the keys among clusters
according to the data types.

From client, it looks single riak cs system as it is currently.
## Data Structure Change

Add **clusterid** entry into each bucket record, which indicates in
which cluster the object manifest is stored.  Add (or Reuse)
**clusterid** entry into each object manifest, which indicates in
which cluster the blocks are stored.

This enables Riak CS to store object manifests and blocks into
separate clusters.
## How to choose cluster to store blocks

When PUT request came from users, CS node gets the bucket record
and retrieves clusterid to store the object manifest.

(Beforehand, when bucket being created Riak CS chooses the cluster to
store object manifests by the same logic from choosing cluster to
store the blocks => open question?)

Then Riak CS puts the object manifest into the cluster indicated by
the clusterid. Inside the object manifest, there are another clusterid
to put blocks, decided by the following manner:
- get estimated free space of each Riak cluster(*)
- filter out the cluster which doesn't have enough free space to store more objects
- probabilistically choose a Riak cluster depending on the amount of free space

Then kick a riak_cs_block_server.

(*) "free space of Riak cluster" can be defined in several ways.
One candidate is as follows:
- For each Riak node, decide `free percentage` from `disksup:get_disk_data()` by using the partition containing `paltform_data_dir`.
- For cluster's `free percentage`, take `minimum` of each node's `free percentage`.
  Taking `minimum` rather than `average` is for safety to avoid disk full.


src/riak_cs_bag.erl
```diff
@@ -0,0 +1,215 @@
...
+-type allocate_type() :: manifest | block.
+-type pool_type() :: request_pool | bucket_list_pool.
+%% Use "bag ID" instead of "cluster ID".
+%% There are more than one clusters in case of MDC.
+-type bag_id() :: binary().
+-type pool_key() :: {pool_type(), bag_id()}.
+-type weight_info() :: #weight_info{}.
+
+-spec is_multi_bag_ebabled() -> boolean().
+is_multi_bag_ebabled() ->
```
kuenishi, line 58: typo!!

src/riak_cs_bag.erl
```diff
@@ -0,0 +1,215 @@
...
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% ---------------------------------------------------------------------
+
+%% @doc Support multi Riak clusters in single Riak CS system
```
kuenishi, line 21: It'd be better to have high level idea of 'bag', what it exactly means, difference from other concepts such as cluster id.

kuenishi:
The term `clusterid` changed to `bag` ?

src/riak_cs_bag_worker.erl
```diff
@@ -0,0 +1,259 @@
...
+
+schedule(State) ->
+    Interval = refresh_interval_msec(),
+    Ref = erlang:send_after(Interval, self(), refresh_by_timer),
+    State#state{timer_ref = Ref}.
+
+json_to_weight_info_list({struct, JSON}) ->
+    json_to_weight_info_list(JSON, []).
+
+json_to_weight_info_list([], WeightInfoList) ->
```
kuenishi, line 157: With jsonx we could have created simple decode/1 function ...

kuenishi, line 157: Like [this](https://github.com/kuenishi/meddatum/blob/master/src/hl7.erl#L111).

``` erlang
encoder() ->
    jsonx:encoder([{weight_info, record_info(fields, weight_info)}],
                  [{ignore, [null]}]).

to_json(#weight_info{}=W) ->
    E = encoder(),
    case E(W) of
        Bin when is_binary(Bin) -> Bin;
        Other -> error(Other)
    end.
```

Maybe we can write a simple record<->JSON mapper with `lists:zip` .

src/riak_cs_bag.erl
```diff
@@ -0,0 +1,215 @@
...
+%% This function assumes that it is called ONLY ONCE at initialization.
+%% TODO: return specs from ETS after initialization?
+-spec pool_specs(term()) -> [{atom(), {non_neg_integer(), non_neg_integer()}}].
+pool_specs(MasterPoolConfig) ->
+    init_ets(),
+    case application:get_env(riak_cs, multi_bag) of
+        undefined ->
+            [];
+        {ok, BagConfig} ->
+            register_pools(BagConfig, BagConfig, MasterPoolConfig, [])
```
kuenishi, line 71: I prefer this style for readability and testability:

``` erlang
 {ok, BagConfigs} ->
   PoolRecords = [bag_config_to_pool_record(BagConfigs, MasterPoolConfig)
                          || BagConfig <- BagConfigs ],
   true = ets:insert(?ETS_TAB, PoolRecords),
   PoolSpecs = [bag_config_to_pool_spec(BagConfig, MasterPoolConfig)
                          || BagConfig <- BagConfigs ]
 end.

bag_config_to_pool_record({BagIdStr, Address, Port}, MasterPoolConfig) ->
    blah blah blah
    #pool{ key = ... }.

bag_config_to_pool_spec({BagIdStr, Address, Port}, MasterPoolConfig) ->
    Name = list_to_atom(....),
             ...
    {Name, ...}.
```

kuenishi:
Just want to make sure, is these correct? :
- user data is in default bag
- bucket data are in default bag
- whole manifests in a bucket are in a same bag, whose bag id is indicated in the bucket data
- whole blocks of an object are in a same bag, whose bag id is indicated in the manifest

shino:
@kuenishi Yes. All are correct.

kuenishi:
- [x] get_fsm path
- [x] put_fsm path
- [x] list objects path
- [x] bucket CRUD path
- [x] user CRUD path
- [x] storage/access calculation path
- [x] balancing system
- [ ] ~~stanchion review again~~
- [ ] ~~heartful hand-made tests~~
- [ ] ~~r_t review~~

I would be happy if a tiny documentation on basic design is written in wiki (or maybe TODO for me).

shino:
I have not mention TODOs, sorry.

Two items are not implemented yet:
- GC (only collect objects in the default riak cluster only)
- MDC proxy_get

src/riak_cs_storage.erl
```diff
@@ -57,19 +56,36 @@ sum_user(Riak, User) when is_list(User) ->
...
+            {Name, iolist_to_binary(io_lib:format("~p", [E]))}
+    end.
+
+-spec sum_bucket_with_pool(pid(), cs_bucket()) -> term() | {error, term()}.
+sum_bucket_with_pool(DefaultRiakc, ?RCS_BUCKET{name=Name} = Bucket) ->
+    case riak_cs_bag:pool_name(request_pool, Bucket) of
+        undefined ->
+            sum_bucket(DefaultRiakc, Name);
+        PoolName ->
+            %% TODO: riak_cs_utils:with_riak_connection(PoolName, Fun) is useful?
```
kuenishi, line 80: I don't think so.

src/riak_cs_bag.erl
kuenishi, line 21: like this:

``` erlang
%% - user records are in default bag
%% - bucket records are in default bag
%% - whole manifests in a bucket are in a same bag, whose bag id is indicated in the bucket data
%% - whole blocks of an object are in a same bag, whose bag id is indicated in the manifest
%%
%% - storage/access usage are in default buckets
```

kuenishi, line 21: I also prefer there should be `app.config` example here; while it is a bit early to put there. So maybe here's the best place to put configuration example.

src/riak_cs_bag_server.erl
```diff
@@ -0,0 +1,171 @@
...
+status() ->
+    gen_server:call(?SERVER, status).
+
+init([]) ->
+    {ok, #state{}}.
+
+handle_call({allocate, Type}, _From, #state{initialized = true} = State)
+  when Type =:= block orelse Type =:= manifest ->
+    Decision = case Type of
+                block ->
```
kuenishi, line 67: weird indentation

src/riak_cs_bag_server.erl
```diff
@@ -0,0 +1,171 @@
...
+                    decide_bag(State#state.blocks);
+                manifest ->
+                    decide_bag(State#state.manifests)
+            end,
+    case Decision of
```
kuenishi, line 72: I prefer this style:

``` erlang
WeightInfo = case Type of
     block -> State#state.blocks;
     manifest -> State#state.manifests
   end,
 case decide_bag(WeightInfo) of
   ...
```

src/riak_cs_bag_server.erl
```diff
@@ -0,0 +1,171 @@
...
+%% bag4   30        60                   31..60
+-spec decide_bag([{riak_cs_bag:pool_key(), riak_cs_bag:weight_info()}]) ->
+                        {ok, riak_cs_bag:bag_id()} |
+                        {error, no_bag}.
+decide_bag([]) ->
+    {error, no_bag};
+decide_bag(WeightInfoList) ->
+    %% TODO: SumOfWeights can be stored in state
+    SumOfWeights = lists:sum([Weight || #weight_info{weight = Weight} <- WeightInfoList]),
+    Point = random:uniform(SumOfWeights),
```
kuenishi, line 118: Just and idea: what if we can make this deterministic ? Say, using `crypto:bytes_to_integer(crypto:md5(term_to_binary({Bucket, Key, UUID}))) rem SumOfWeights` would enable us reproduce specific behavior.

kuenishi, line 118: Though `#weight_info{}` is not so deterministic ...

src/riak_cs_bag_server.erl
```diff
@@ -0,0 +1,171 @@
...
+%% Decide bag to allocate block/manifest randomly regarding weights
+%% bag    weight    cummulative-weight   point (1..60)
+%% bag1   20        20                    1..20
+%% bag2   10        30                   21..30
+%% bag3    0        30                   N/A
+%% bag4   30        60                   31..60
+-spec decide_bag([{riak_cs_bag:pool_key(), riak_cs_bag:weight_info()}]) ->
+                        {ok, riak_cs_bag:bag_id()} |
+                        {error, no_bag}.
+decide_bag([]) ->
```
kuenishi, line 113: It's also not good to use three different words "decode", "allocate" and "assign" for the same purpose. It would be confusing for people who reads the code in future.

src/riak_cs_bag_worker.erl
```diff
@@ -0,0 +1,259 @@
...
+handle_weight_info_list({error, notfound}, State) ->
+    lager:debug("Bag weight information is not found"),
+    {ok, [], State#state{failed_count = 0}};
+handle_weight_info_list({error, Reason}, #state{failed_count = Count} = State) ->
+    lager:error("Retrieval of bag weight information failed. Reason: ~p", [Reason]),
+    {error, Reason, State#state{failed_count = Count + 1}};
+handle_weight_info_list({ok, Obj}, State) ->
+    %% TODO: How to handle siblings
+    [Value | _] = riakc_obj:get_values(Obj),
+    Weights = binary_to_term(Value),
```
kuenishi, line 145: we have to use `binary_to_term(Value, [safe])` and handle badarg.

src/riak_cs_bag_server.erl
reiddraper, line 118: If possible, anything random we do should have an option to pass a random number seed in, so we can make deterministic tests.

src/riak_cs_bag_worker.erl
```diff
@@ -0,0 +1,259 @@
...
+    end.
+
+handle_weight_info_list({error, notfound}, State) ->
+    lager:debug("Bag weight information is not found"),
+    {ok, [], State#state{failed_count = 0}};
+handle_weight_info_list({error, Reason}, #state{failed_count = Count} = State) ->
+    lager:error("Retrieval of bag weight information failed. Reason: ~p", [Reason]),
+    {error, Reason, State#state{failed_count = Count + 1}};
+handle_weight_info_list({ok, Obj}, State) ->
+    %% TODO: How to handle siblings
```
kuenishi, line 143: According to the weight data structure, I think we can use CRDT-style conflict resolution. Also, it would be much better if we have an interface to change weight info incrementally, because in usual operation cluster expansion are always incremental like adding nodes, removing nodes and not refreshing nodes.

src/riak_cs_bag_server.erl
shino, line 118: I prefer deterministic way. @kuenishi 's suggestion sounds nice.

slfritchie:
Closing, pending re-review.

Issue #716, opened 2013-10-24
Add concurrency to GC


kellymclaughlin:
Currently the cleanup of deleted or overwritten object versions is done by the singleton GC daemon process. The daemon gathers the set of deletion-eligible keys from the GC bucket and iterates over the keys in a serial manner. Make changes so that the GC daemon starts and manages multiple worker processes. Also provide a way to dynamically add and remove worker processes during a GC run. The goal is to add more concurrency and control to the GC process in order to allow GC to better handle cases where there is large amount of cleanup work to do.

reiddraper:
Remind me, is there work on a branch for this? If not, I say we push it to 1.6.

kellymclaughlin:
Yeah it's [here](https://github.com/basho/riak_cs/tree/feature/gc-concurrency), though it is in need of rebase from current `develop`.

shino:
@kellymclaughlin How about the status of the branch? Almost finished or need some more work?
I want this functionality in multi-cluster because it's possible that more garbage are generated.

I can take over the branch if you don't mind :)

kellymclaughlin:
@shino Current status is the code should be working, but the `riak_cs_gc_d_eqc` test needs to be updated since it's currently not working. A similar test for the new gc worker module also needs to be added. If you want to pick it up that would be great.

shino:
Thanks for the information.
The branch just rebased to current develop is [here](https://github.com/basho/riak_cs/tree/feature/gc-concurrency-rebased).

shino:
Addressed by #830

PR #830, opened 2014-04-03
GC concurrency (rebased)

shino:
Add concurrency to GC execution.
This issue addresses #716  and based on @kellymclaughlin 's work (h/t).

GC worker processes are introduced.
`riak_cs_gc_d` executes 2i reuqest and pass keys to workers.

A unit of work is
- all keys in a single 2i response when paginated, or
- keys divided to groups which have `gc_batch_size` number of keys each.


kuenishi:
Quick result on `make test`

```
/home/kuenishi/src/riak_cs/rebar eunit skip_deps=true
Building riak-cs-ee
==> rel (eunit)
==> riak_cs (eunit)
Compiled src/riak_cs_access_console.erl
Compiled src/riak_cs_wm_bucket_versioning.erl
src/riak_cs_gc_d.erl:710: function eligible_manifest_key_sets/3 undefined
src/riak_cs_gc_d.erl:711: function eligible_manifest_key_sets/3 undefined
src/riak_cs_gc_d.erl:712: function eligible_manifest_key_sets/3 undefined
src/riak_cs_gc_d.erl:713: function eligible_manifest_key_sets/3 undefined
src/riak_cs_gc_d.erl:714: function eligible_manifest_key_sets/3 undefined
src/riak_cs_gc_d.erl:716: function eligible_manifest_key_sets/3 undefined
ERROR: eunit failed while processing /home/kuenishi/src/riak_cs: rebar_abort
make: *** [test] エラー 1
(v)kuenishi@debian:~/src/ri
```

shino:
Oops. I pushed the fix.

src/riak_cs_gc_worker.erl
```diff
@@ -0,0 +1,285 @@
...
+    {ok, Pid} = riak_cs_delete_fsm_sup:start_delete_fsm(node(), Args),
+
+    %% Link to the delete fsm, so that if it dies,
+    %% we go down too. In the future we might want to do
+    %% something more complicated like retry
+    %% a particular key N times before moving on, but for
+    %% now this is the easiest thing to do. If we need to manually
+    %% skip an object to GC, we can change the epoch start
+    %% time in app.config
+    link(Pid),
```
kuenishi, line 145: Probably this is not the scope of the review, but starting fsm and linking is not atomic, and `link/1` fails if `Pid` does not exist. Then what happens to the `RestKeys` ? Will be deleted at next batch?

src/riak_cs_gc_worker.erl
```diff
@@ -0,0 +1,285 @@
...
+    {reply, {StateName, State}, StateName, State};
+handle_sync_event({change_state, NewStateName}, _From, _StateName, State) ->
+    ok_reply(NewStateName, State);
+handle_sync_event(stop, _From, _StateName, State) ->
+    {stop, normal, ok, State};
+handle_sync_event(_Event, _From, StateName, State) ->
+    ok_reply(StateName, State).
+
+handle_info(_Info, StateName, State) ->
+    {next_state, StateName, State}.
```
kuenishi, line 184: Maybe we have to handle something that comes from delete_fsm here. `'EXIT'` and `'DOWN'` ?

src/riak_cs_gc_worker.erl
shino, line 145: Good question. If delete fsm finished normally or died accidnetaly before `link/1`, worker dies at `link/1` call by `noproc`. So as `riak_cs_gc_d`. Then it will restart.

kuenishi, line 145: So the whole batch fails involving other delete_fsm ? That might be insane situation, so I think we need some logs for better operation, at least at `riak_cs_gc_d`.

kuenishi:
mm, no `riak_test` ... that would be another bunch of work.

kuenishi:
Does the manual batch (esp. `riak_cs_gc_d:manual_batch/1` ) block until the batch finishes? The behavior has been changed from old one. I don't think the operator does not wait until the whole batch finishes (=batches empty, 0 workers).

src/riak_cs_gc_d.erl
```diff
@@ -175,132 +216,115 @@ fetching_next_batch(continue, #state{batch_start=undefined,
...
-    %% finished with this batch
+    {next_state, feeding_workers, NewStateData};
+fetching_next_batch(continue, ?STATE{active_workers=0,
+                                     batch=[],
+                                     continuation=undefined,
+                                     batch_caller=Caller} = State) ->
+    %% finished with this GC run
+    _ = case Caller of
+            undefined -> ok;
+            _ -> Caller ! {batch_finished, State}
```
kuenishi, line 235: `noproc` have to be handled here, because possibly operator may shut down the console while waiting for  `riak-cs-gc batch` command finish. Or at least following lager message output must be come before this line.

shino:
> Does the manual batch (esp. riak_cs_gc_d:manual_batch/1 ) block until the batch finishes? 

Really? I thing the command returns after `riak_cs_gc_d`'s state is changed to `fetching_next_batch`. If not, it would be a bug.

shino:
I should add one commnet. `batch_caller` was added for testing. It is used in `riak_cs_gc_single_run_eqc`.

src/riak_cs_gc_d.erl
```diff
@@ -175,132 +216,115 @@ fetching_next_batch(continue, #state{batch_start=undefined,
...
+idle({manual_batch, Options}, {CallerPid, _Tag}=_From, State) ->
     Leeway = leeway_option(Options),
-    ok_reply(fetching_next_fileset, start_manual_batch(
-                                      lists:member(testing, Options),
-                                      Leeway,
-                                      State));
+    ok_reply(fetching_next_batch, start_manual_batch(
+                                    lists:member(testing, Options),
+                                    Leeway,
+                                    State?STATE{batch_caller=CallerPid}));
```
kuenishi, line 324: I would write like this, if `batch_caller` is meant for testing:

``` erlang
NewState = case lists:member(testing, Options) of
    true -> start_manual_batch(true, Leeway, State?STATE{batch_caller=CallerPid});
    false -> start_manual_batch(false, Leeway, State)
end
```

kuenishi:
One thing I worry about is that we already have task-queue-and-parallel-workers style   access archiver. Storage calculation is yet single-threaded but it might be better done in parallel. It might be good time to stop repeating ourselves, this time or next time.

Code looks correct except above mentions.

src/riak_cs_gc_worker.erl
kuenishi, line 145: Maybe this doesn't happen because starting delete_fsm is enough slow so that `link/1` gets in time before delete_fsm terminates. OK.

kuenishi:
> Really? I thing the command returns after riak_cs_gc_d's state is changed to fetching_next_batch. 

Okay. I didn't find that function.

kuenishi:
OK, I'm persuaded. +1 from me. I think we'd better have original author's review - @kellymclaughlin do you have any seconds to take a glance at this?

kuenishi:
+1  b6ce119

src/riak_cs_gc_worker.erl
shino, line 184: TCP errors are handled by `{error, Reason}`, not by exceptions. `'EXIT'` of delete_fsm may be really disastrous, such as riak dies. So I think it would be nice `riak_cs_gc_d` and friends crash and restart (by supervisor).

src/riak_cs_gc_d.erl
shino, line 235: `!` does not throw `noproc`.

shino, line 235: But I should mention `batch_caller` is used only by `riak_cs_gc_single_run_eqc` (at least currently). I will do that.

shino:
Remaining: add comment about `batch_caller`, fix dialyzer warning.

shino:
Remaining items at the above comments were addressed at bf89876 and 9431489.

kuenishi:
+1 9431489

shino:
@borshop merge

PR #829, opened 2014-04-03
Add cluster-info

ksauzz:
Add riak-cs-cinfo as a command-line tool to get cluster-info. 

_This PR was extracted from #826._

PR #826, opened 2014-03-26
WIP: Introduce riak-cs-admin

ksauzz:
RiakCS has various CLI tools and they are expected to increase number of them gradually in future. To avoid increasing the number of them I suggest introducing riak-cs-admin into RiakCS as administration CLI like riak-admin for Riak. Changes on this PR include follows:
- riak-cs-admin
  - move the implementation of existing tools into riak-cs-admin
  - leave existing tools as adopter to riak-cs-admin for a while, just in case.
  - add cluster-info to riak-cs-admin
- new stats format
  - add new stats to riak-cs-admin with the same format of `riak-admin status` and also `/stats`. 
  - add config `stats_format` to switch a stats format and keep backward compatibility.
- riak-cs-debug
  - add riak-cs-debug

TODO
- [x] how to deal with current `/stats` API's format ?
  - option1: switch it by config `{cs_stats_version, v2}`
  - option2: switch it by url param `/stats?v2` or `/stats/v2
- [x] how to add riak-cs-debug, this subcommand or not ? It looks like too large code base to add riak-cs-admin as its subcommand.
- [ ] regression test.
#### riak-cs-admin

```
% ./riak-cs-admin
Usage: riak-cs-admin { status | cluster-info | gc | access | storage | stanchion }
```
##### sub commands

```
% ./riak-cs-admin status | head
1-minute stats for 'riak-cs@127.0.0.1'
-------------------------------------------
block_get_total : 0
block_get_rate : 0.0
block_get_time_mean : 0.0
block_get_time_median : 0.0
block_get_time_95 : 0.0
block_get_time_99 : 0.0
block_get_retry_total : 0
block_get_retry_rate : 0.0
```

```
% ./riak-cs-admin cluster-info
Usage: riak-cs-admin cluster-info <output_file>
```

```
% ./riak-cs-admin gc
Usage: riak-cs-admin gc { batch [<leeway_seconds>] | status | pause | resume | cancel |
                          set-interval <interval_seconds> | set-leeway <leeway_seconds> }
```

```
% ./riak-cs-admin access
Usage: riak-cs-admin access  { flush }
```

```
% ./riak-cs-admin storage
Usage: riak-cs-admin storage  { batch | status | pause | resume | cancel }
```

```
% ./riak-cs-admin stanchion
Usage: riak-cs-admin stanchion  { switch HOST PORT | show }
```
#### v0

```
% ./riak-cs-admin status
           block_get:   0       0.0     0.0     0.0     0.0     0.0
[...]

PR #497, opened 2013-03-16
rebar config script to isolate ee deps

kellymclaughlin:

PR #831, opened 2014-04-03
unintentionally pkg.vars.config was in .gitignore, recovering

kuenishi:
At b79d88596281513d658f4d pkg.var.config was put into .gitignore, this commit saves it from ignorance. What makes me wonder is why packaging system has been working correctly through several 1.4.x releases - probably `git archive` command let it resurrect in the archive. Whatever, without this file we can't add any commands in #826, #829 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@ current_counterexample.eqc
 .int_test/
 .client_test/
 priv/riak_cs_drv.so
-pkg.vars.config
 client_tests/python/ceph_tests/s3-tests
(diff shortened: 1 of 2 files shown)


kuenishi:
@kellymclaughlin if you have any memory on b79d885 , please let me stop merging this.

shino:
pkg.vars.config is not a source but generated by rebar.config.script.
https://github.com/basho/riak_cs/pull/497

kuenishi:
oops.

Issue #475, opened 2013-03-08
MP initiated objects remains after delete/create buckets


shino:
commit 304c7396d7b64cb2a11fe126efa65067b40fba2a (HEAD, origin/release/1.3, release/1.3)

steps to reproduce:
1. Create a bucket
2. Initiate MP upload
3. Confirm list of MP uploads: 1 object
4. Delete the bucket
5. Confirm list of MP uploads: 1 object

Expectation: no objects listed at step 5. (as s3 does)

```
$ s3cmd mb s3://test-tmp
Bucket 's3://test-tmp/' created
$ s3curl.pl --id local_rcs \
>     --post \
>     -- -s -v -x localhost:8080 \
>     'http://test-tmp.s3.amazonaws.com/test1.txt?uploads'
> POST http://test-tmp.s3.amazonaws.com/test1.txt?uploads HTTP/1.1
> User-Agent: curl/7.27.0
> Host: test-tmp.s3.amazonaws.com
> Accept: */*
> Proxy-Connection: Keep-Alive
> Date: Fri, 08 Mar 2013 05:38:57 +0000
> Authorization: AWS RUY9FSJHGMBQMFU7CITP:vKN22Wd8g/jfe56TrK2uwVDmIZY=
>
< HTTP/1.1 200 OK
< Server: Riak CS
< Date: Fri, 08 Mar 2013 05:38:57 GMT
< Content-Type: application/xml
< Content-Length: 239
<

$ s3curl.pl --id local_rcs \
>     -- -s -v -x localhost:8080 \
>     'http://test-tmp.s3.amazonaws.com/?uploads'
> GET http://test-tmp.s3.amazonaws.com/?uploads HTTP/1.1
> User-Agent: curl/7.27.0
> Host: test-tmp.s3.amazonaws.com
> Accept: */*
> Proxy-Connection: Keep-Alive
> Date: Fri, 08 Mar 2013 05:39:26 +0000
> Authorization: AWS RUY9FSJHGMBQMFU7CITP:sjyDY/WUHBL7IOB0W1SnRH4Nhao=
>
< HTTP/1.1 200 OK
< Server: Riak CS
< Date: Fri, 08 Mar 2013 05:39:26 GMT
< Content-Type: application/xml
< Content-Length: 674
<
[...]

PR #842, opened 2014-04-18
Bugfix/cs475 check mpuploads

kuenishi:
addresses #475 . With Stanchion pull request: https://github.com/basho/stanchion/pull/77

These pull requests let Riak CS do:
- delete all writing multipart manifests in usual Riak CS way just before deleting the bucket
- delete all writing multipart manifests in usual Riak CS way just before creating the bucket
- Just before creating bucket, Stanchion checks whether there are no writing multipart manifests and if it's true, then creates the bucket - this is done inside the critical section.

Note: this won't work well with multi bag, as stanchion accesses the object manifest bucket.


reiddraper:
Thanks @kuenishi. I'll finish reviewing today and tomorrow. This is really reiterating to me how much we need a layer of abstraction between human bucket names and the buckets stored in Riak. Each time a new bucket 'foo' is created, we should be using a new bucket in Riak with just a UUID. The mapping of 'foo' -> UUID would exist in the bucket record. That way, if two users owned the same bucket (over time), they never actually store data in the same Riak bucket. This is a large-ish change, however, and is outside the scope of 1.5.0.

kuenishi:
> how much we need a layer of abstraction between human bucket names and the buckets stored in Riak

Exactly. TODOs gonna be much more piled up.

I hope that abstraction layer will be cut out in some sort of external library and code duplication with object manifests be avoided.

kuenishi:
After yesterday's discussion with Shino, we concluded to change the design a bit
- don't delete all writing multipart manifests in usual Riak CS way just before creating the bucket
- instead add a cleanup API into Stanchion, where Stanchion cleans up whole multipart uploads without active bucket
- to detect inactive buckets are easy, just iterate over `moss.buckets` and find deleted bucket and start cleanup.

With this way, users can cleanup existing buckets without stopping stanchion service.

kuenishi:
#842 and stanchion/#77 are ready for review again. ping @shino @reiddraper

shino:
I will jump in as a 2nd reviewer :smile: 
The stanchion issue @kuenishi  mentioned above is  CS's #77.
Stanchion's one is here > https://github.com/basho/stanchion/pull/77

src/riak_cs_bucket.erl
```diff
@@ -0,0 +1,690 @@
...
+            riak_cs_utils:key_exists(RiakcPid, Bucket, Key)
+    end.
+
+%% @doc Check if a bucket exists in the `buckets' bucket and verify
+%% that it has an owner assigned. If true return the object;
+%% otherwise, return an error.
+%%
+%% @TODO Rename current `bucket_exists' function to
+%% `bucket_exists_for_user' and rename this function
+%% `bucket_exists'.
```
shino: It seems natural for me that the function with name `bucket_exists` returns boolean.

src/riak_cs_mp_utils.erl
```diff
@@ -221,6 +218,22 @@ list_multipart_uploads(Bucket, {_Display, _Canon, CallerKeyId} = Caller,
...
+        {ok, ?INDEX_RESULTS{keys=Names}} ->
+            {ok, list_multipart_uploads2(Bucket, RiakcPid,
+                                         Names, Opts)};
+
+        Else2 ->
+            Else2
+    end.
+
+
+
```
shino: Three blank lines.

src/riak_cs_bucket.erl
kuenishi: See also: https://github.com/basho/riak_cs/pull/761/files#diff-7c73ce393c6dde0096f84131eae5aa7bL1116

shino: :see_no_evil:

src/riak_cs_bucket.erl
```diff
@@ -0,0 +1,690 @@
...
+         get_bucket_acl_policy/3,
+         maybe_log_bucket_owner_error/2,
+         resolve_buckets/3,
+         update_bucket_record/1
+        ]).
+
+-include("riak_cs.hrl").
+-include_lib("riak_pb/include/riak_pb_kv_codec.hrl").
+-include_lib("riakc/include/riakc.hrl").
+-include_lib("xmerl/include/xmerl.hrl").
```
shino: Where is this include used?

shino: Compile succeeded only with "riak_cs.hrl"

src/riak_cs_bucket.erl
```diff
@@ -0,0 +1,690 @@
+%% ---------------------------------------------------------------------
+%%
+%% Copyright (c) 2007-2013 Basho Technologies, Inc.  All Rights Reserved.
```
shino: It's 2014 now.

src/riak_cs_bucket.erl
```diff
@@ -0,0 +1,690 @@
...
+                    true -> {true, ok};
+                    false -> {false, {error, bucket_not_empty}}
+                end;
+            false -> {true, ok}
+        end,
+    case AttemptDelete of
+        true ->
+            %% TODO: output log if failed in cleaning up existing uploads.
+            %% The number of retry is hardcoded.
+            retry_delete_bucket(User, UserObj, Bucket, RiakPid, 5);
```
shino: Just out of curiosity, what is the reason to retry here?

kuenishi: To reduce the probability of conflicting against concurrent mp initiate. Riak CS has 5 chances to clean up the bucket.

src/riak_cs_bucket.erl
```diff
@@ -0,0 +1,690 @@
...
+            case riak_cs_gc:gc_specific_manifests(
+                   [M?MANIFEST.uuid], Obj, Bucket, Key, RiakPid) of
+                {ok, _NewObj} ->
+                    fold_delete_uploads(Bucket, RiakPid, Ds);
+                E ->
+                    lager:debug("cannot delete multipart manifest: ~p ~p (~p)",
+                                [{Bucket, Key}, M?MANIFEST.uuid, E]),
+                    E
+            end;
+        _E ->
```
shino, line 229: How about make `_E` more explicit, that is, just `false`?

shino, line 229: ah... I was wrong... MP uploads with state =/= writing can enter this branch. Please ignore the above comment.

src/riak_cs_bucket.erl
```diff
@@ -0,0 +1,690 @@
...
+        false ->
+            LocalError
+    end.
+
+retry_delete_bucket(_, _, Bucket, _, 0) ->
+    lager:error("All retry on deleting bucket '~s' failed "
+                "because multipart upload still exists (or newly created).",
+                [Bucket]),
+    %% TODO: needs discussion,or "please reduce your request rate".
+    {error, multipart_delete_retry_exhausted};
```
shino: Why not respond with 409 like stanchion?

kuenishi: Because no clients can't handle this (it's not defined in S3).

src/riak_cs_bucket.erl
shino: It's certainly user-friendly that riak cs does automatic retry.

I'm afraid that the interval between retries can be ~ minutes, because there are stanchion calls and list keys operations (pb timeout is 60sec=1min). It widen the possibility of race (another DELETE/PUT bucket).

kuenishi:
mmm, that dialyzer.

```
==> riak_cs (dialyzer)
riak_cs_bucket.erl:235: Function fold_all_buckets/3 has no local return
riak_cs_bucket.erl:236: The call riak_cs_bucket:iterate_csbuckets(Pid::any(),Acc0::any(),Fun::fun(),'undefined') will never return since it differs in the 4th argument from the success typing arguments: (pid(),any(),fun(),binary())
tools.mk:35: recipe for target 'dialyzer' failed
make: *** [dialyzer] Error 1
```

kuenishi:
r_t, xref, dialyzer, eunit, pulse all passed. Ready for review, again.

riak_test/tests/buckets_test.erl
```diff
@@ -0,0 +1,156 @@
...
+
+%% keys for non-multipart objects
+-define(TEST_BUCKET,        "riak-test-bucket").
+-define(KEY_SINGLE_BLOCK,   "riak_test_key1").
+-define(KEY_MULTIPLE_BLOCK, "riak_test_key2").
+
+%% keys for multipart uploaded objects
+-define(KEY_MP_TINY,        "riak_test_mp_tiny").  % single part, single block
+-define(KEY_MP_SMALL,       "riak_test_mp_small"). % single part, multiple blocks
+-define(KEY_MP_LARGE,       "riak_test_mp_large"). % multiple parts
```
shino: More appropriate names? :smiley:

src/riak_cs_console.erl
```diff
@@ -46,6 +51,73 @@ cluster_info([OutFile]) ->
...
+    _ = io:format("cleaning up with timestamp ~s", [Timestamp]),
+    Fun = fun(BucketName, GetResult, Acc0) ->
+                  _ = maybe_cleanup_csbucket(Pid, BucketName, GetResult, Timestamp),
+                  Acc0
+          end,
+    _ = riak_cs_bucket:fold_all_buckets(Fun, [], Pid),
+
+    ok = riak_cs_riakc_pool_worker:stop(Pid),
+    _ = lager:info("~nall old unaborted orphan multipart uploads has deleted.~n", []),
+    _ = io:format("~nall old unaborted orphan multipart uploads has deleted.~n", []).
```
shino: s/all/All/, s/has/have been/ ?

src/riak_cs_s3_response.erl
```diff
@@ -88,6 +89,7 @@ error_message(invalid_range) -> "The requested range is not satisfiable";
 error_message(invalid_bucket_name) -> "The specified bucket is not valid.";
 error_message(unexpected_content) -> "This request does not support content";
 error_message(canned_acl_and_header_grant) -> "Specifying both Canned ACLs and Header Grants is not allowed";
+error_message(remaining_multipart_upload) -> "Concurrent multipart upload initiation detected. Please stop it to delete or create bucket.";
```
shino, line 92: Does `riak_cs_oos_response` need similar changes?

shino, line 92: When creating bucket, user can not stop initiating mp uploads. So remove "or create" ?

src/riak_cs_s3_response.erl
```diff
@@ -169,6 +172,7 @@ status_code(invalid_bucket_name) -> 400;
 status_code(unexpected_content) -> 400;
 status_code(canned_acl_and_header_grant) -> 400;
 status_code(malformed_acl_error) -> 400;
+status_code(remaining_multipart_upload) -> 409;
```
shino, line 175: 409 is nice when a user attempt to delete a bucket. It seems a little bit strange for creating buckets.

src/riak_cs_s3_response.erl
kuenishi, line 92: I don't think so. Swift API doesn't support multipart uploads.

src/riak_cs_s3_response.erl
kuenishi, line 175: Changed to 503, by changing return value of `riak_cs_bucket:handle_stanchion_response/3`.

src/riak_cs_s3_response.erl
kuenishi, line 92: See also: https://wiki.openstack.org/wiki/Swift/APIFeatureComparison#Amazon_S3_REST_API_Compatability

shino, line 92: Wow. I understand.

src/riak_cs_s3_response.erl
kuenishi, line 175: No, 500.

src/riak_cs_s3_response.erl
kuenishi, line 92: removed "or create" in #856

src/riak_cs_s3_response.erl
kuenishi, line 175: fixed in #856

kuenishi:
continues to #856

Issue #849, opened 2014-05-07
Storage calculation doesn't handle tombstones, nor handle undefined manifest.props


kuenishi:
around here: https://github.com/basho/riak_cs/blob/d0501f50a499ca0b1a4d5e55e850ff30d62ce129/src/riak_cs_storage.erl#L104

This leads to storage calculation skips the objects where `.props`  is  `undefined`.

kuenishi:
Maybe related to #840 .

shino:
just a note: object list (v1) properly uses upgrading:
https://github.com/basho/riak_cs/blob/release/1.4/src/riak_cs_utils.erl#L510

see also : #670 #671

kuenishi:
see also: #524

kuenishi:
#840 identified with this issue. will open another PR.

PR #848, opened 2014-05-06
Feature/xmerl avoid undef rebased

reiddraper:
Subsumes #706. This is #706 rebased and with 0079a4a908757714baebef70e538fea936d324a7 added.

Issue #840, opened 2014-04-17
Storage Statistics Inaccurate in 1.4.5


zschlag:
A customer recently upgraded from CS 1.4.1 to 1.4.5.  Afterwards we noticed a couple users had dropped off our weekly CS Tenant Utilization reports.  It seemed unlikely that these users had suddenly deleted their data, so I investigated into this.  Below is an example from one of the users:

Before the upgrade (3/25), there are 9 Objects and 126877696 bytes, then after (4/12) it shows 0 Objects/Bytes. I get the same results if I check storage via curl or s3cmd.

```
(>o_o)> ~ :: s3cmd get s3://riak-cs/usage/••••••••••••••••••••/b/20140316T170000Z/20140416T170000Z
s3://riak-cs/usage/••••••••••••••••••••/b/20140316T170000Z/20140416T170000Z -> ./20140416T170000Z  [1 of 1]
 1268 of 1268   100% in    1s   758.46 B/s  done

(>o_o)> ~ :: grep -o "{\"StartTime\"[^}]*}}" 20140416T170000Z | sort
{"StartTime":"20140317T090516Z","EndTime":"20140317T090517Z","anonymous-bucket":{"Objects":9,"Bytes":126877696}}
{"StartTime":"20140318T090803Z","EndTime":"20140318T090803Z","anonymous-bucket":{"Objects":9,"Bytes":126877696}}
{"StartTime":"20140322T090224Z","EndTime":"20140322T090231Z","anonymous-bucket":{"Objects":9,"Bytes":126877696}}
{"StartTime":"20140324T091106Z","EndTime":"20140324T091106Z","anonymous-bucket":{"Objects":9,"Bytes":126877696}}
{"StartTime":"20140325T091022Z","EndTime":"20140325T091022Z","anonymous-bucket":{"Objects":9,"Bytes":126877696}}
{"StartTime":"20140412T090937Z","EndTime":"20140412T090938Z","anonymous-bucket":{"Objects":0,"Bytes":0}}
{"StartTime":"20140412T225305Z","EndTime":"20140412T225306Z","anonymous-bucket":{"Objects":0,"Bytes":0}}
{"StartTime":"20140413T091507Z","EndTime":"20140413T091508Z","anonymous-bucket":{"Objects":0,"Bytes":0}}
{"StartTime":"20140414T090303Z","EndTime":"20140414T090303Z","anonymous-bucket":{"Objects":0,"Bytes":0}}
[...]

Issue #837, opened 2014-04-15
log stacktrace in try/catch


rzezeski:
While doing a bunny tour @cdahlqvist came across an customer issue
where he observed the following in the log:

```
2014-04-08 01:05:05.314 [error] <0.28335.192> Riak CS object list map failed for <<48,111,58,159,41,117,103,93,157,237,246,106,105,56,228,6,81,180,210>>:<<"FL_30146370-ff4b-46c2-9761-c92f77ff6d4f_07088df2a63ee578e125f71126d42150f33b4209">> with reason error:badarg
```

Here is the relevant code: https://github.com/basho/riak_cs/blob/1.4.5/src/riak_cs_utils.erl#L507

It wraps several function calls into a `try/catch` and logs an error
if any type of error occurs whether it be throw, error, or
exit. Unfortunately, the stacktrace is not gathered (via
`erlang:get_stacktrace`) and logged and thus you have no way to
determine which line of code actually produced the `badarg`. There is
a good chance the `term_to_binary` is responsible but you cannot be
100% certain without a trace.

In general, even if only calling one function, you should log the
stacktrace when using `try/catch`. The error could come from that
function or any function it calls or any function that functions
calls, and so forth. Without the stacktrace you are left to guess at
the problem. If you are lucky you have access to the environment and
can run the code by hand to determine the source of failure but it
would be simpler just to log the stacktrace in the first place.

kuenishi:
~~Closing~~ as #864 and #865 merged, where printing stacktraces and removeing tombstones included. There seems to exist few try/catch sentences with deep call stack inside.

kuenishi:
No, not closing yet.

PR #850, opened 2014-05-07
handle tombstones and undefined manifest props

kuenishi:
addresses #849 .

### DIFF
--- a/src/riak_cs_storage.erl
+++ b/src/riak_cs_storage.erl
@@ -146,9 +146,14 @@ get_usage(Riak, User, Start, End) ->
     {ok, Period} = archive_period(),
     rts:find_samples(Riak, ?STORAGE_BUCKET, User, Start, End, Period).
 
+-spec count_multipart_parts([{binary(), lfs_manifest()}]) ->
+                                   {non_neg_integer(), non_neg_integer()}.
 count_multipart_parts(Resolved) ->
     lists:foldl(fun count_multipart_parts/2, {0, 0}, Resolved).
 
+-spec count_multipart_parts([{binary(), lfs_manifest()}],
+                            {non_neg_integer(), non_neg_integer()}) ->
+                                   {non_neg_integer(), non_neg_integer()}.
 count_multipart_parts({_UUID, ?MANIFEST{props=Props, state=writing} = M},
                       {MPparts, MPbytes} = Acc)
   when is_list(Props) ->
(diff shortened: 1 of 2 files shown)


kuenishi:
TODO: write tests by using `delete_mode` .

reiddraper:
@kuenishi any preference on a milestone for this issue?

kuenishi:
Now I am investigating #840 with @zschlag, which is in 1.5.0. This is expected to solve #840 but not sure and some weird things left. For the time being I'll mark this as 1.5.0.

reiddraper:
Thanks @kuenishi.

src/riak_cs_storage.erl
```diff
@@ -102,8 +102,9 @@ object_size_map({error, notfound}, _, _) ->
 object_size_map(Object, _, _) ->
     try
         AllManifests = [ binary_to_term(V)
-                         || V <- riak_object:get_values(Object) ],
-        Resolved = riak_cs_manifest_resolution:resolve(AllManifests),
+                         || V <- riak_object:get_values(Object), V /= <<>> ],
```
shino: How about replacing `/= <<>>` by riak_cs_utils:has_tombstone/1`?

shino:
If we had some warning (or error?) logs at catch block, analysis would be easier.
Is it a good idea  that we add log in this PR?

See : https://github.com/basho/riak_cs/issues/837

kuenishi:

kuenishi:
@shino yeah, logging is good, definitely.

shino:
Do we need some tests for `props=undefined` and/or tombstoned manifest?

kuenishi:
Thanks to @zschlag , this patch will definitely solve #840 . It wasn't `.props=undefined` but old `manifest_v1` . I think adding tests may be another issue. Thoughts?

kuenishi:
Note: Maybe we should change our codes to print stacktrace.

kuenishi:
We saw `function_clause [{proplists,get_value,[multipart,undefined,undefined]` and `reason error:{badrecord,multipart_manifest_v1} [{riak_cs_storage,count_multipart_parts,2,[{file,"src/riak_cs_storage.erl"},{line,160}]}` in the log.

src/riak_cs_storage.erl
```diff
@@ -111,7 +109,14 @@ object_size_map(Object, _, _) ->
             _ ->
                 [{MPparts, MPbytes}]
         end
-    catch _:_ ->
+    catch Type:Reason ->
+            _ = lager:log(error,
+                          self(),
+                          "Riak CS object list map failed for ~p:~p with reason ~p:~p",
```
shino, line 115: list?

kuenishi:
This branch will be rebased.

PR #648, opened 2013-08-09
riak_test for Riak Enterprise 1.4 Replication v3 proxy_get

bookshelfdave:
code around "object_three" has been commented out, as I need a better way to detect when proxy_get is enabled/disabled. If anyone would be so kind as to review and file any issues against this PR, I would be more than happy to revisit after vacation.

Best of luck on your release!

Cheers -
Dave

PR #869, opened 2014-05-28
Feature/repl v3 test #648 rebased

kuenishi:
#648 have been long left, and it just works. So we'd better merge it than nothing until 1.5.0.

### DIFF
--- a/riak_test/src/repl_helpers.erl
+++ b/riak_test/src/repl_helpers.erl
@@ -152,6 +152,44 @@ wait_until_leader(Node) ->
         end),
     ?assertEqual(ok, Res).
 
+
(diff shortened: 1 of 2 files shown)


kuenishi:
Test failure was because of timeout (1200 seconds) of long PLT build due to version change of OTP to R15. dializer in R15 is slow because it runs in single thread.

kuenishi:
rebased to latest develop to include #873 .

reiddraper:
@borshop merge

bookshelfdave:
yay!

shino:
Yay! Great work >  @metadave  @kuenishi and @reiddraper :exclamation:

PR #853, opened 2014-05-08
Feature/ bag support

shino:
To enable multibag feature.

PR #870, opened 2014-05-28
Riak client abstraction layer for bag support

shino:
Instead of using `riakc_pb_socket` directly, add `riak_cs_riak_client` that understands CS specific types like users and buckets. `riak_cs_riak_client` wraps one or more `riakc_pb_socket` processes. All Webmachine resource callbacks and gc/access daemons use them. 

This PR includes also:
- multibag support (this is motivation for introducing `riak_cs_riak_client`)
- bug fix of premature connection release ba13361


reiddraper:
Is this necessary for the multi-cluster work? Is that why we're trying to target it for 1.5?

shino:
Yes. This started as a refactoring work of #853. (maybe it seems like a rewrite?)

shino:
Rebased one #874 
For reviewers, please pick either one :) Then I will close another one.

shino:
Close in favor of #874

PR #872, opened 2014-05-28
Feature/riak_test flavor

shino:
By adding flavor setting to riak_test config, generic test cases which depends only on S3 API and does not on riak cs / riak internal settings can be run in different settings.

Flavors affect setup functions `rtcs:setup/1` and `rtcs:setup/2`.
Other setup functions such as `rtcs:setup2x2` (used by `repl_test`) does not change their behavior.

To specify a flaver, riak_test.config should be 

```
{cs, [
    [snip]
     {flavor, FlavorYouWant}
    ]}.
```

Commit until 99c1d7d are the same as #870 (this PR in on #870 but target to `develop`).


shino:
Rebased on develop after #874 merged. Ready for review.

kuenishi:
Looks good to me, but could you update riak_test/README.md to remind me how and let new riak_test user know this?

shino:
Good point :smile: Added in 5e0bf3d.

riak_test/src/rtcs.erl
```diff
@@ -49,11 +49,9 @@ setup(NumNodes) ->
     setup(NumNodes, default_configs()).
 
 setup(NumNodes, Configs) ->
-    JoinFun = fun(Nodes) ->
-                      [First|Rest] = Nodes,
-                      [join(Node, First) || Node <- Rest]
-              end,
-    setup_clusters(Configs, JoinFun, NumNodes).
+    Flavor = rt_config:get(flavor, basic),
+    lager:info("Use flavor : ~p", [Flavor]),
```
kuenishi: Just `"flavor: ~p"` would be better English, I think.

riak_test/src/rc_helper.erl
```diff
@@ -0,0 +1,64 @@
...
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% ---------------------------------------------------------------------
+
+-module(rc_helper).
+-compile(export_all).
+-include_lib("eunit/include/eunit.hrl").
+
+to_riak_bucket(manifest, CSBucket) ->
```
kuenishi: It would be better to use same atom name "objects" and "blocks" as `riak_cs_utils:to_bucket_name/2` [like in CS](https://github.com/basho/riak_cs/blob/develop/src/riak_cs_utils.erl#L708) .

riak_test/src/rc_helper.erl
```diff
@@ -0,0 +1,64 @@
...
+
+get_riakc_obj(RiakNodes, ObjectKind, CsBucket, CsKey) ->
+    Pbc = rtcs:pbc(RiakNodes, ObjectKind, CsBucket, CsKey),
+    RiakBucket = to_riak_bucket(ObjectKind, CsBucket),
+    RiakKey = to_riak_key(ObjectKind, CsKey),
+    Result = riakc_pb_socket:get(Pbc, RiakBucket, RiakKey),
+    riakc_pb_socket:stop(Pbc),
+    Result.
+
+update_riakc_obj(RiakNodes, ObjectKind, CsBucket, CsKey, NewObj) ->
```
kuenishi, line 49: like in original `to_bucket_name/2` , just `Type` is a bit more comprehensive than ObjectKind  to me.

shino, line 49: The name `to_bucket_name` restricts context well. But this function does not.
At least for me (maybe only for me :P ), only the word `Type` sounds ambiguous.
What kind? Riak nodes or bucket or ... hmm...

kuenishi, line 49: Or maybe annotating with spec would be enough? I don't have strong opinion on the variable name.

shino:
Add one commit to address comments. Thanks.

kuenishi:
+1  18be862

shino:
@borshop merge

PR #873, opened 2014-05-28
Update line number in dialyzer ignore warnings

reiddraper:

PR #874, opened 2014-05-29
Riak client abstraction [rebased1]

shino:
Rebased PR of #870 

Instead of using `riakc_pb_socket` directly, add `riak_cs_riak_client` that understands CS specific types like users and buckets. `riak_cs_riak_client` wraps one or more `riakc_pb_socket` processes. All Webmachine resource callbacks and gc/access daemons use them. 

This PR includes also:
- multibag support (this is motivation for introducing `riak_cs_riak_client`)
- bug fix of premature connection release 60b32b3

For reviewers:
Because this PR (or #870) has large diff, so commits are `rebase -i`ed
to form understandable chunks. Some are trivial (superfluous changes) and
others are adding functionalities or fix bugs.

For example, the compare view [1] without first 4 commits are composed
of `892 additions and 666 deletions`, so isn't it less touble than reviewing
full diff?

[1] https://github.com/basho/riak_cs/compare/8031fd8...feature%2Friak-client-rebased?expand=1


.gitignore
```diff
@@ -22,5 +22,6 @@ riak-cs.png
 .local_dialyzer_plt
 .eqc-info
 .riak_cs_list_objects_fsm_v2_eqc.txt
+dialyzer.ignore-warnings
```
kuenishi, line 25: I don't think we need this line.

dialyzer.ignore-warnings.ee
```diff
@@ -1,6 +1,6 @@
 # Errors
 Function get_block_remote/5 has no local return
-riak_cs_config.erl:203: Function will never be called
+riak_cs_config.erl:208: Function will never be called
```
kuenishi, line 3: This may conflict with #873 . Is this branch rebased on this?

include/riak_cs.hrl
```diff
@@ -19,6 +19,7 @@
 %% ---------------------------------------------------------------------
 
 -define(MANIFEST, #lfs_manifest_v3).
+-define(MANIFEST_REC, lfs_manifest_v3).
```
kuenishi: multipart manifest uses `MULTIPART_MANIFEST_RECNAME` so this one is also good to use `MANIFEST_RECNAME`.

include/riak_cs.hrl
```diff
@@ -82,8 +84,9 @@
                   bucket :: binary(),
                   acl :: 'undefined' | acl(),
                   requested_perm :: acl_perm(),
-                  riakc_pid :: pid(),
-                  riakc_pool :: atom(),
+                  riak_client :: riak_client(),
+                  rc_pool :: atom(),
```
kuenishi: What does rc stand for? It might be good to anotate in comments.

shino: Sure. It stands for `riak_client` :)

include/riak_cs.hrl
```diff
@@ -129,8 +132,9 @@
 -type acl() :: #acl_v1{} | #acl_v2{}.
 
 -type cluster_id() :: undefined | binary(). %% flattened string as binary
-
 -type cs_uuid() :: binary().
+-type bag_id() :: undefined | binary().
+-type riak_client() :: riak_client().
```
kuenishi: tautology?

shino: ( ﾟ∀ﾟ)･∵. ｸﾞﾊｯ!!

include/riak_cs_gc_d.hrl
```diff
@@ -50,13 +53,15 @@
           max_workers :: non_neg_integer(),
           active_workers=0 :: non_neg_integer(),
           %% Used for paginated 2I querying of GC bucket
-          continuation :: undefined | binary(),
+          key_list_state :: undefined | term(),
```
kuenishi: Seems like this should be `gc_key_list_state()` .

include/riak_cs_gc_d.hrl
```diff
@@ -70,6 +75,25 @@
           delete_fsm_pid :: pid()
          }).
 
+-record(gc_key_list_state, {
+          remaining_bags :: [{bag_id(), string(), non_neg_integer()}],
```
kuenishi, line 79: Should this be `[{bag_id(), inet:hostname(), inet:port_number()}]`?

kuenishi, line 79: Also, it'd better to have initial value as `[]` instead of undefined for dialyzer

include/riak_cs_gc_d.hrl
shino: We should have some prefix for record name because it is (in a sense) global/public. But I have no strong opinion for attributes in records.
If we put prefix, should other attributes have? (e.g. `max_gc_workers`, `lfs_manifest_uuid`, etc.) Isn't it annoying?

include/riak_cs_gc_d.hrl
shino, line 79: You are correct and it should be so.
But I want to defer this change as an future issue because there are other same usages of `string()` and `non_neg_integer()` for hostnames and port numbers.

include/riak_cs_gc_d.hrl
kuenishi: I meant the type, not the record element name; so this should be; `key_list_state :: undefined | gc_key_list_state(),`

include/riak_cs_gc_d.hrl
kuenishi, line 79: okay, this is gonna be the first tech dept in 1.5, right?

shino, line 79: Sure, filed at #875

include/riak_cs_gc_d.hrl
shino: I completely misunderstood. Sorry. You commented on `term()`, not `key_list_state`.

src/riak_cs_riak_client.erl
```diff
@@ -0,0 +1,304 @@
...
+
+init([]) ->
+    {ok, fresh_state()}.
+
+handle_call(stop, _From, State) ->
+    _ = do_cleanup(State),
+    {stop, normal, ok, State};
+handle_call(cleanup, _From, State) ->
+    {reply, ok, do_cleanup(State)};
+handle_call({get_bucket, BucketName}, _From, State) ->
```
kuenishi, line 165: These API-specific message handling could be done outside the gen_server. This gen_server just keeps the pid and gives caller pid to run specific API call like `do_get_bucket`, in `get_bucket/3`. This will reduce most `handle_call` heads and simplify this module.

src/rts.erl
```diff
@@ -48,6 +48,7 @@
          iso8601/1
         ]).
 
+-include("riak_cs.hrl").
```
kuenishi: I'd rather prefer preserving `rts.erl` as it is, because this seems to be just a copy-and-paste module. This should rather be put out as a common library.

src/riak_cs_riak_client.erl
shino, line 165: `do_get_bucket/1` is also state changing function. If we devide get_bucket API, then it will look like

```
get_bucket(PcPid, BucketName) ->
    case call(RcPid, master_pbc) of
      {error, _} -> %% Some error handling
      {ok, Pbc} ->
          case get_bucket_with_pbc(Pbc, BucketName) of
              {error, _} -> %% ...
              {ok, BucketObj} ->
                  ok = set_bucket_obj(RcPid) % new API
          end
    end.
```

kuenishi:
Still a few comments left but very great work, @shino !

.gitignore
shino, line 25: I devided dialyzer's ignore-warnings to oss and ee.
`dialyzer.ignore-warnings` is now craeted by rebar.config.script depending on `RIAK_CS_EE_DEPS`.

If we don't expect OSS user to execute dialyzer (and xref) then I will remove this line and recover `dialyzer.ignore_warnings` from EE version.
Thought?

dialyzer.ignore-warnings.ee
shino, line 3: What a brand new PR! I will do just before merge.

src/riak_cs_riak_client.erl
shino, line 165: I want(ed) `riak_cs_riak_client` processes to be a wrappers of PB
connection(s) and hiding underlying complexities (as `riakc_pb_socket`
hides TCP socket).

But I don't stick around to change the direction. Just being holder of
PB connection(s) may be good too.

Is it reasonable to choose the (wishful/future) direction and describe
it in the module comment?
1. Hide PB connections inside `riak_cs_riak_client` processes (close to current impl.)
2. Just manage PB connections and give it to other processes (close to your suggestion).
3. Other (please add your idea)

Currently the module has `master_pbc`, `manifest_pbc` and `block_pbc` that
expose PB connections to outside. If we choose 1. above, these should be
removed and replaced by additional external APIs like `list_keys(BucketName)`,
`mapred(BucketName, ...)` etc.

src/rts.erl
shino: `find_samples/6` uses `riak_client()` now (and `riak_cs.hrl` includes `riak_client()` type def.)
Other functions are "functional" as you say.
Hmm...

include/riak_cs.hrl
shino: This macro is not used... Sorry. Removed at bb2b8f5.

include/riak_cs.hrl
shino: Fixed bb2b8f5

include/riak_cs.hrl
shino: Fixed bb2b8f5

include/riak_cs_gc_d.hrl
shino: Fixed bb2b8f5

shino:
Thanks a lot for review of big diff PR! @kuenishi 

Some comments remain open. Please discuss about them later.

src/riak_cs_riak_client.erl
kuenishi, line 165: I understood your idea after looking at multibag codes. Totally reasonable. So forget my question above, thanks.

.gitignore
kuenishi, line 25: Okay, I understand.

reiddraper:
Added 1.5.0 milestone.

shino:
All comments so far were addressed.
Rebased on develop and force pushed.
(please reset if you checkout this branch locally)

shino:
Micro benchmark on my laptop

# Environment

riak
- 1 node
- ring size: 16
- n_val: 1 (!)
- backend: leveldb for default, bitcask for blocks and memory for manifests (to avoid fluctuation with leveldb)

b_b
- driver: basho_bench_driver_cs (so watch out what latency means..)
- key generator: partitioned_sequential_int (both for preload and get)
- keys: 100,000 (10^5)
- file_size: 10bytes (so small)
- workers: 8

# Comparison
- CS 1.4.5 (commit: 1596a166739554b89e709d04baf10916689b9d3f) versus
- this branch (feature/riak-client-rebased: a032386b0f2487f6a22b1a27335649d58d5bf286)

## PUT (preload)
- CS 1.4.5
  cs145-preload_summary
- this branch
  sb-3-preload_summary

## GET (after preload)
- CS 1.4.5
  cs145-3-get_summary
- this branch
  sb-3-get_summary

shino:
The above comment was wrongly on stanchion's issue. Transcribed.

kuenishi:
+1 0bae83d

kuenishi:
+1 145bbbf

shino:
@borshop merge

kuenishi:
:boom:

reiddraper:
Nice work all. A minor nitpick: we've been trying to keep `tools.mk` synced across repositories, would it be possible to make these changes general enough to apply to all of our repositories?

shino:
@reiddraper I thank a lot to https://github.com/basho/tools.mk , because the changes of `tools.mk` file in this PR was just a copy from it :smile: 

In detail, hipe generated some additional information logs and I struggled with grep expression. But it was already there in `tools.mk` repo :+1:

reiddraper:
Ok great, thanks!

PR #877, opened 2014-06-03
fix request pool size on riak_test

kuenishi:
stats_test has been failing due to the change of default pool size.

### DIFF
--- a/riak_test/src/rtcs.erl
+++ b/riak_test/src/rtcs.erl
@@ -45,6 +45,15 @@
 -define(S3_PORT, 80).
 -define(DEFAULT_PROTO, "http").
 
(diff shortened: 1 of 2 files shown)


kuenishi:
oh no, this was just for my branch.

kuenishi:
Better way.

shino:
@kuenishi this PR should go to target branch instead of develop?

kuenishi:
@shino Both is okay, but to avoid bad merge conflict we'd better use either of them. I think develop and release/1.5 should have same diff set until 1.5.0 release. So, for now let's use develop.

kuenishi:
@borshop merge

PR #885, opened 2014-06-12
Reduce execution time of repl test (v2 and v3)

shino:
The most time consuming part is five retries for getting blocks.
After making retry number configurable, use small number `1` for testing.

There are also some misc refinement such as removing sleep or replacing sleep with `wait_until`.

One example of execution times are as follows.

Before
- repl_test  20.78s user 8.88s system 7% cpu 6:26.88 total
- repl_v3_test  21.25s user 9.68s system 6% cpu 7:44.34 total

After
- repl_test  21.09s user 10.34s system 41% cpu 1:15.17 total
- repl_v3_test  20.61s user 9.04s system 23% cpu 2:04.00 total

### DIFF
--- a/riak_test/src/repl_helpers.erl
+++ b/riak_test/src/repl_helpers.erl
@@ -3,6 +3,10 @@
 -include_lib("eunit/include/eunit.hrl").
 
 verify_sites_balanced(NumSites, BNodes0) ->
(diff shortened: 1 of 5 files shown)


kuenishi:
+1 f7abe97
great

kuenishi:
@borshop merge

PR #890, opened 2014-06-16
update deps to the latest

kuenishi:
- eper to 0.78, same as riak develop
- riak-erlang-client to 1.4.2, which includes [cs bucket list fix](https://github.com/basho/riak-erlang-client/commit/01cdc125c1c8f14a40309f8fdb45587a3b28e37f)
- ~~webmachine to 1.10.6~~
- poolboy to 0.8.1p1
- folsom to 0.8.1

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -36,15 +36,16 @@
 
 {deps, [
         {node_package, "1.3.8", {git, "git://github.com/basho/node_package", {tag, "1.3.8"}}},
+        {getopt, ".*", {git, "git://github.com/jcomellas/getopt.git", {tag, "v0.4.3"}}},
         {webmachine, ".*", {git, "git://github.com/basho/webmachine", {tag, "1.10.3"}}},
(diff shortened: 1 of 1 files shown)


kuenishi:
Webmachine 1.10.6 doesn't work well because it requires ibrowse 4.0.1. On the other hand erlcloud uses ibrowse 4.0.2.

erlcloud in upstream is not using ibrowse any more (lhttpc, which now has timeouts), but it is not easy to rebase our ibrowse to latest erlcloud upstream. Or it may be good to back to original erlcloud.

kuenishi:
The conflict in 5daa487 is ibrowse 4.0.2 in erlcloud vs ibrowse 4.0.1 in webmachine 1.10.6 . Both are used in only tests, so there should be non-critical way.

kuenishi:
I think it's ready for review. riak_test running. => riak_test ok.

kuenishi:
@borshop merge

Issue #855, opened 2014-05-12
Inconsistent ETag on complete multipart upload POST response


hectcastro:
When completing a multipart upload against Amazon S3, the XML payload contains a specialized `ETag`:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<CompleteMultipartUploadResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
   <Location>http://joker-test.s3.amazonaws.com/25.txt</Location>
   <Bucket>joker-test</Bucket>
   <Key>25.txt</Key>
   <ETag>"6ce4f6859aaecc0e7f23fc487c31f8db-2"</ETag>
</CompleteMultipartUploadResult>
```

According to a [response](https://forums.aws.amazon.com/thread.jspa?messageID=456442) from an Amazon employee on their support forums, this `ETag` is:

> The checksum in a multipart digest is the MD5 of the concatenated MD5 bytes of each of the parts affixed with the number of parts
> 
> Basically:
> 
> ETag = MD5(Sum(p \in numberParts, MD5(PartBytes(p))) + "-" + numberParts

In this example, I uploaded a `25MB` file in chunks of `20MB`, so the `numberParts` part of that adds up. I also used an [S3 client for Go](https://github.com/rlmcpherson/s3gof3r) that calculates the MD5 of the concatenated MD5 bytes for each part, and that is consistent as well.

When completing a similar multipart upload against Riak CS, the `ETag` attribute looks different (it actually looks like the `UploadId`):

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<CompleteMultipartUploadResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
   <Location>http://isos.s3.amazonaws.com/25.txt</Location>
   <Bucket>isos</Bucket>
   <Key>25.txt</Key>
   <ETag>oD_XfZorRXOY6ouY61cFMA==</ETag>
</CompleteMultipartUploadResult>
```

Response from multipart upload initialization to show that the `ETag` is the `UploadId`:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<InitiateMultipartUploadResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
   <Bucket>isos</Bucket>
   <Key>25.txt</Key>
   <UploadId>oD_XfZorRXOY6ouY61cFMA==</UploadId>
</InitiateMultipartUploadResult>
```

[...]

Issue #876, opened 2014-05-29
Avoid unnecessary logging on anonymous access.


ksauzz:
We shouldn't output no_user_key error on proper anonymous access such as getting a public object.

```
2014-05-29 18:08:35.835 [error] <0.2515.0>@riak_cs_wm_common:maybe_create_user:218 Retrieval of user record for s3 failed. Reason: no_user_key
```

shino:
For example, if one use riak_cs for file download with public read ACL, this log can happen heavily. Lager, by default setting, calls `file:datasync/1` for error or higher severity [1]. So this issue may increase disk load.

[1] https://github.com/basho/lager/blob/master/src/lager_file_backend.erl#L228

shino:
Move to 1.5.0

shino:
Resolved by #883.

PR #891, opened 2014-06-16
Add release note on #855 fix

kuenishi:


shino:
Sorry that I forgot to add #876 . Could you please add it?

shino:
> Also, all CS nodes which don't run GC should be upgraded before CS
>   node which does run GC upgraded to let leeway second system work
>   properly.

Is it danger that GC-enabled node has old version?
DELETEd objects may be collected just after deletion.

```
                    DELETE object
  CS w/o GC --------X--------------
  (1.5.x)           |
                    |
                    V
  riak-cs-gc  ----------------------
                       ^
                       |
                       |
  CS w/ GC  -----------O-----------
  (1.4.x)              GC triggered
```

When GC-enabled node is new version, it is safer.

```
                    DELETE object
  CS w/o GC --------X--------------
  (1.4.x)           |
                    +--------+
                             V
  riak-cs-gc  ----------------------
               ^
               +-------+
                       |
  CS w/ GC  -----------O-----------
  (1.5.x)              GC triggered
```

shino:
Thanks for adding the line :smile: 
Diff looks nice.

kuenishi:
@borshop merge

PR #892, opened 2014-06-17
Create any number of CS nodes by devrel or stagedevrel

shino:
Like Riak's `DEVNODES=10 make devrel`

This PR also
- changes `cs_port` of `dev/dev?`, for example dev1 has 15018 from 8071, and
- removes `rel/vars/rtdev?_vars.config` because they are not used.


shino:
Wrong target branch `develop`.

PR #896, opened 2014-06-20
make rebar output package_version to pkg.vars.config file

binarytemple:
The first step towards an easier riak/riak_cs setup. 

The goal is to remove the following manual configuration step. 

In the `/etc/riak/app.config` file, you must specify the following, `{add_paths, ["/usr/lib/riak-cs/lib/riak_cs-1.4.5/ebin"]},`.

That path changes on every release, which complicates configuration management. 

With the version information in the pkg.vars.config file perhaps we can do a copy or symlink the directory 
`/usr/lib/riak-cs/lib/riak_cs-<version>/ebin` to `/usr/lib/riak-cs/lib/riak_cs-latest/ebin`.

### DIFF
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -12,6 +12,9 @@ HashMacroAdded =
             end
     end,
 
+{ok, VsnString} = rebar_utils:sh("git describe --always --tags", [{cd, "."}, {use_stdout, false}]),
+Version = string:strip(VsnString, right, $\n),
+
 {IsEE, Package} = case os:getenv("RIAK_CS_EE_DEPS") of
                       false -> {false, "riak-cs"};
                       _     -> {true, "riak-cs-ee"}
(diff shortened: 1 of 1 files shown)


kuenishi:
We can do it without changing `pkg.var.config` by just symlinking outside the packages, and put a line like `{add_paths, ["/some/symlink/path/to/riak-cs-ebin"]}`. 

```
# dpkg -i riak-cs-1.5.0beta1-xxxx.deb
# ln -sf /usr/lib/riak-cs/lib/riak-cs-1.5.0beta1/ebin /some/symlink/path/to/riak-cs-ebin
# emacs /etc/riak/app.config
```

Or could you show me the whole picture?

binarytemple:
This stuff was prompted by a couple of things, that I experienced, and that customers experienced, and I handled while working tickets.
- The first time anyone installs Riak CS -  they forget to update the path, the system doesn't start properly and it wastes an hour (minimum (not particularly helpful error message) ) while they figure out what went wrong.
- The second reason is Configuraton Management. Although it is a relatively simple thing, it requires manual intervention, if Riak CS requires manual configuration, then it is inconvenient. If it is inconvenient, then it becomes less attractive. 

This is about helping to improve the out-of-the-box experience. Getting to the stage where a user has a working installation, just by installing the appropriate packages, would be very good for us.

shino:
Ping in repo cleanup sweeping.
This PR is inactive for long. Is it worth to keep open?

binarytemple:
Probably not worth keeping it open, nobody seemed interested at the time - and quite some time has gone by since I originally created it.

The original intent was to surface the information about the Riak CS ebin path for use by the package tools (RPM/Deb/etc). 

This would assist in automating the installation of Riak and Riak CS together on a box - perhaps without the operator needing to manually add the Riak CS path to the advanced.conf - and make installation faster and simpler.

PR #898, opened 2014-06-24
Use standard lager for logs of "make test-client"

shino:
The log of `make test-client` in `external_client_tests` currently
goes `/tmp/*` and is removed after the `make` exits.

When executing riak_test in batch (e.g. `-d riak_test/ebin`), I
sometimes want to read such log files to investigate why the test fails.

This PR makes the log be forwarded to lager and logged to cosole or files.

One subtle point is progress indicator by dots `....` are logged after
line break character comes.
This can be handled but I now leave it future refactoring in the future.

### DIFF
--- a/riak_test/tests/external_client_tests.erl
+++ b/riak_test/tests/external_client_tests.erl
@@ -7,50 +7,53 @@
 -define(TEST_BUCKET, "external-client-test").
 
 confirm() ->
-    {UserConfig, {RiakNodes, _CSNodes, _Stanchion}} = rtcs:setup(2, [{cs, cs_config()}]),
-    ?assertEqual(ok, erlcloud_s3:create_bucket("external-client-test", UserConfig)),
-
-    CS_http_port = rtcs:cs_port(hd(RiakNodes)),
-
-    CS_src_dir = rt_cs_dev:srcpath(cs_src_root),
-    lager:debug("cs_src_root = ~p", [CS_src_dir]),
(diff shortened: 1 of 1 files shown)


riak_test/tests/external_client_tests.erl
```diff
@@ -7,50 +7,53 @@
...
-                lager:error("Expected 'Res 0', got: ~s", [Res]),
-                error(fail)
-        end
-    after
-        os:cmd("rm -f " ++ StdoutStderr)
+get_cmd_result(Port) ->
+    WaitTime = rt_config:get(rt_max_wait_time),
+    receive
+        {Port, {data, {Flag, Line}}} when Flag =:= eol orelse Flag =:= noeol ->
+            lager:info(Line),
```
kuenishi, line 46: I prefer `debug` here as I don't want to see details mostly, except in case investigation on client tests needed.

kuenishi:
Other part are quite nice and I'll give future +1 to this pull request.

riak_test/tests/external_client_tests.erl
kuenishi, line 46: Either is fine, and this will be improved when we afford our time to reduce the info log.

shino:
After approval, but I note some ideas:
- accumulate log lines from `make client-test` and output them if error occurs, otherwise discard them

A variant is:
- Output recent 50 or 100 lines if error occurs.

shino:
@borshop merge

Issue #900, opened 2014-06-25
500 return when trying to copy an object uploaded by multipart


kuenishi:
on `release/1.5` branch with e9c9e16a62db126dd

```
2014-06-25 09:56:31.103 [error] <0.123.0> Lager event handler error_logger_lager_h exited with reason {'EXIT',{{badmatch,["/buckets/test2/objects/100MB",{error,{error,function_clause,[{base64,encode_to_string,[{<<63,106,166,120,163,119,42,196,219,217,165,110,39,131,247,68>>,"-7"}],[{file,"base64.erl"},{line,46}]},{riak_cs_copy_object,test_condition_and_permission,4,[{file,"src/riak_cs_copy_object.erl"},{line,44}]},{riak_cs_wm_object,handle_copy_put,4,[{file,"src/riak_cs_wm_object.erl"},{line,366}]},{riak_cs_wm_common,accept_body,2,[{file,"src/riak_cs_wm_common.erl"},{line,346}]},{webmachine_resource,...},...]}},...]},...}}
```

shino:
Addressed by #901

PR #901, opened 2014-06-25
Bugfix/copy from multipart

shino:
- Fix bug that PUT copy from MP uploaded source results in 500 error
  (expected : successfully copied)
- Fix bug that PUT copy from absent key results in 503
  (expected : 404 error) 

Includes tests by boto for both fixes and some refactoring.

Addresses #900


shino:
@borshop merge

Issue #880, opened 2014-06-09
update s3-tests to latest (esp. include put copy code)


kuenishi:
after #868 merged.

shino:
Addressed by #906

PR #903, opened 2014-06-26
Delete Multiple Objects should be also RESTful as usual Delete

kuenishi:
Found by s3-tests when working on #880 .  Integration tests will be followed by #880 .


shino:
@borshop merge

PR #904, opened 2014-06-26
Fix zero-size object copy failure

kuenishi:
Found via work on #880 . Tests will follow.

### DIFF
--- a/src/riak_cs_copy_object.erl
+++ b/src/riak_cs_copy_object.erl
@@ -159,6 +159,14 @@ copy(PutFsmPid, SrcManifest, ReadRcPid) ->
                   {ok, DstManifest::lfs_manifest()} | {error, term()}.
 copy(_, _, _, fail) ->
     {error, bad_request};
+copy(PutFsmPid, ?MANIFEST{content_length=0} = _SrcManifest,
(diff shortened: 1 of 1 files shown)


src/riak_cs_copy_object.erl
```diff
@@ -159,6 +159,14 @@ copy(PutFsmPid, SrcManifest, ReadRcPid) ->
                   {ok, DstManifest::lfs_manifest()} | {error, term()}.
 copy(_, _, _, fail) ->
     {error, bad_request};
+copy(PutFsmPid, ?MANIFEST{content_length=0} = _SrcManifest,
+     _ReadRcPid, _) ->
+    %% Zero-size copy will successfully create zero-size object
+    riak_cs_put_fsm:finalize(PutFsmPid, undefined);
+copy(PutFsmPid, _SrcManifest, _ReadRcPid, {Start, End})
+  when Start =:= End ->
```
shino: Start and end of Range are both index, so `Start =:= End` means 1 byte range, not empty.

kuenishi: So, there are no zero-size copy using Range, right?

shino:
@borshop merge

Issue #882, opened 2014-06-10
Suppress and/or warning for bloated manifests #(sibling), value size, #({uuid, manifest})


shino:
In multipart upload with high concurrency (> 30 or > 100) with small part size (~ 5MB),
the number of siblings (in riak_object sense) for the manifest grows to O(1000)
and in consequence the object size grows to nearly 1GB.
I guess riak sibling explosion happened [1].

Riak (riak_kv) also logs for many siblings and/or large object size,
it will be helpful to log warnings/errors in riak cs side because
bucket name in riak_kv's log are hashed.

Some backpressure to such sibling explosion is still under discussion.
Ideas so far are:
1. If #(siblings) exceeds a threashold, respond with 503 and
   hope clients sleep and retry.
2. If #(siblings) exceeds a threashold, sleep 3 seconds for example
   (or random seconds between 1 and 5 sec is better?) and fetch the
   manifest again, resolve and write back it.
3. [Please put your idea here]

As for value size, deserializing 1GB binary consumes considerable memory,
it may be worth to check it.

[1] https://github.com/basho/riak_test/pull/383

reiddraper:
This is the have Riak CS `lager:warning/error` when a manifest reaches a certain size?

shino:
I had to start from background :)

_Comment below were moved to the description_

---

In multipart upload with high concurrency (> 30 or > 100) with small part size (~ 5MB),
the number of siblings (in riak_object sense) for the manifest grows to O(1000)
and in consequence the object size grows to nearly 1GB.
I guess riak sibling explosion happened [1].

Riak (riak_kv) also logs for many siblings and/or large object size,
it will be helpful to log warnings/errors in riak cs side because
bucket name in riak_kv's log are hashed.

Some backpressure to such sibling explosion is still under discussion.
Ideas so far are:
1. If #(siblings) exceeds a threashold, respond with 503 and
   hope clients sleep and retry.
2. If #(siblings) exceeds a threashold, sleep 3 seconds for example
   (or random seconds between 1 and 5 sec is better?) and fetch the
   manifest again, resolve and write back it.
3. [Please put your idea here]

As for value size, deserializing 1GB binary consumes considerable memory,
it may be worth to check it.

[1] https://github.com/basho/riak_test/pull/383

reiddraper:
Hmm, any idea why/how we're creating so many siblings? I would've hoped they would be being resolved as they were being created.

shino:
[...]

Issue #679, opened 2013-09-25
Adress large manifests


reiddraper:
Riak CS objects which are frequently overwritten can cause large manifest objects. Currently, we keep old copies of the manifest around for 24 hours (configurable). I'm wondering if it's worth doing something to what vector-clock pruning does, and prune at the sooner of a time _or_ number of manifests threshold. This would allow us to prune old manifests before 24 hours, if we have $N old copies already around.

reiddraper:
Fixed in #732.

PR #905, opened 2014-06-26
Add exponential back off with randomization to avoid sibling explosion

shino:
Addresses #882 (together with #915).

When multipart uploads are executed in concurrent, get-modify-put manifest objects
may be interleaved and lead to sibling explosion.
This PR add retries with sleep to releive it.

Three knobs are introduced.
- `manifest_update_max_retries` (default: 4)
  
  Max number of reties for GET manifests before update.
- `manifest_siblings_suppress` (default: 5)
  
  The threshold for the number of manifest object siblings to trigger suppression logic.
  If the number of siblings is equal to or greater than this value,
  retry logic is applied.
- `manifest_siblings_force_update_at_last_retry` (default: 15)
  
  The threshold for the number of manifest object siblings to
  force update at the last retry.
  If the number of siblings is less than this value,
  update without gamble (see below) at the last retry.
### Logic

After getting manifest object, check the number of siblings.
If it is less than `manifest_siblings_suppress`,
modify/put it without sleep or retry.
Otherwise, do a simple gamble.
- First, generate random uniform integer from 0 to 9.
- If it is 0, then win! update modify/put manifests.
- Other than that, sleep and retry from getting manifest object.

This randomness is necessary because there should be a chance to
resolve siblings which needs modify/put.

Also, at the last retry attempt, depending on the #(siblings),
the upload forcedly move to upload without gambling.

Another randomness is in sleep intervals.
Sleep intervals are uniformly distributed in the ranges 0.5 +/- 0.25, 1 +/- 0.5,
2 +/- 1, 4 +/- 2, for 1st retry, 2nd, 3rd,... respectively.
### Misc notes
- No dependency on the number of siblings other than threashold for suppression.
  If the current logic not enough, some penalty for sleep interval for large
  number of siblings might help (or not. I don't know).
- Add `manifest_update_retry_sleep` to riak_cs_stats.
  It collects sleep intervals as latency (a bit of abuse).

### DIFF
--- a/src/riak_cs_manifest_fsm.erl
+++ b/src/riak_cs_manifest_fsm.erl
@@ -299,6 +299,17 @@ get_and_delete(RcPid, UUID, Bucket, Key) ->
     end.
 
 get_and_update(RcPid, WrappedManifests, Bucket, Key) ->
+    MaxRetries = riak_cs_config:get_env(riak_cs, manifest_update_max_retries, 5),
+    get_and_update(RcPid, WrappedManifests, Bucket, Key, MaxRetries, 0).
+
+get_and_update(_RcPid, _WrappedManifests, _Bucket, _Key, MaxRetries, Retries)
+  when MaxRetries < Retries ->
+    {{error, retry_exceeded}, undefined, undefined};
+get_and_update(RcPid, WrappedManifests, Bucket, Key, MaxRetries, Retries) ->
+    case Retries of
+        0 -> ok;
+        _ -> sleep_retries(Retries)
+    end,
     %% retrieve the current (resolved) value at {Bucket, Key},
     %% add the new manifest, and then write the value
     %% back to Riak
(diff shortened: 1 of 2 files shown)


kuenishi:
Maybe #679 is related

shino:
Current status of the branch
- rebased on release/1.5 after 1.5.0 release
- remove warning log because it was already added at 1.5.0
- isolate async update error handling (not yet completed), which should be addressed another issue. related to #920 

TODOs
- Increase sleep interval for "far" beyond threshold (e.g. > 2*threshold)
- Add the way to monitor retry by threshold (warning log is too noisy, probably).

kuenishi:
With this code, when # of siblings is more than threshold and there're less than 10 concurrent uploads, there happens unnecessary wait. Then the max retry is 5, so if there are a few concurrent uploads, there will be unnecessary wait and re-getting from Riak. If Riak object larger, the read load will get heavier. Andalso, there is non-trivial possibility to fail for all 5 retries. Its probability is `0.9^5 = 0.590..`

So I would suggest not to throw a dice when `MaxRetries == Retries` which is the last chance.

shino:
> Its probability is 0.9^5 = 0.590..

This is the probability in case all the concurrent uploads fails to resolve siblings.
If one upload can resolve siblings, then other uploads after retry does not see siblings over threshold.

My estimation is
- thresold=5 then there are 5+ uploads
- each uploads retry up to 5 times, them 25+ trials
- 1/10 winning percentage is not so bad to bet

But... more consideration needed (for me)

kuenishi:
Well, then we need discussion on whether `0.9^25 = 0.0717...` is low enough for that. Let's spill water in tern.

shino:
Update description to reflect the last commit 3f51452

shino:
Close this.
Retries in overloaded situation seems bad.  Prefer simple sleep at #959 .

PR #911, opened 2014-07-07
Fix policy version validation in PUT Bucket Policy

kuenishi:
CS 1.4 does not validate the version of policy json when the policy
is being put, but validates when 'using' it (=anytime when you touch
anything in that bucket). This commit adds validation in WM resource
of PUT Bucket policy and related error codes.

There might be invalid policy versions in already upload buckets,
this commit also validates only version "2012-10-17" and "2008-10-17"
which are also valid in S3. Other invalid version strings continue
to be invalid. With existing buckets created at 1.3 and 1.4 with
invalid version strings, any operations on that bucket or on objects
in that bucket can't be touched and refused with 403.

Tests and workaround to fixup those buckets will follow.


kuenishi:
Example in docs are already in 2012 while we still live in 2008. http://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html

shino:
@borshop merge

shino:
Oops, the above mention from the issue 977 was wrong. Please ignore it.

PR #919, opened 2014-07-17
Update release notes

kuenishi:
- take off preview from put copy
- add Delete Multiple Objects

PR #923, opened 2014-07-17
Remove erroneous entry from 1.5.0 release notes

kellymclaughlin:
Remove the entry about the initial garbage collection delay from the
1.5.0 section of the release notes. This feature was added in 1.4.2
and is already documented in the release notes of that section.


kellymclaughlin:
I spoke with @michellep and she agreed we should remove the duplication.

@kuenishi Could you remove the corresponding entry from the Japanese translation if it exists?

kuenishi:
@kellymclaughlin thank you for pointing it out. We haven't even yet started translation.

But we're working against release/1.5 branch, so I removed the above sentence at #919 instead.

PR #924, opened 2014-07-28
Allow multipart orphan cleanup to be run from command line

kellymclaughlin:
In the release notes for 1.5.0 in the comments about upgrading we have the instructions that require attaching to the console and calling an erlang function:

> - run riak_cs_console:cleanup_orphan_multipart/0 or riak_cs_console:cleanup_orphan_multipart/1 in an attached console to cleanup all buckets

I think is easy enough to go ahead and add this to the command line tools in order to make it easier for users. This is just a suggestion and it could be done after 1.5.0 is released, but it should be very easy to do and better for users.

cc: @michellep

### DIFF
--- a/rel/files/riak-cs-admin
+++ b/rel/files/riak-cs-admin
@@ -15,7 +15,7 @@ cd $RUNNER_BASE_DIR
 SCRIPT=`basename $0`
 
 usage() {
-    echo "Usage: $SCRIPT { gc | access | storage | stanchion | cluster-info }"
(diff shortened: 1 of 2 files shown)


kellymclaughlin:
I decided to abandon my `node_package` change and just have the `cleanup-orphan-multipart` command work in the same manner as zero parameter rpc calls in `riak-admin`.

shino:
@borshop merge

PR #932, opened 2014-07-30
Bugfix/kill copy on disconnect

kuenishi:
Copying can't be stopped on client disconnect, because it does not need any socket read/writes once copying started. That makes detecting client disconnect harder - this patch puts a zero-timeout read on socket to find `{error, closed}` and other errors. This is a dirty hack and mochiweb or webmachine should have such interface that handles lower layer connection error handling, but for now we adopt this hack, which should be fixed in the future. 
Copy should be cancelled because orphan copy without clients are likely not required by clients and may cause overloaded cluster.

Important limitation is: if client sends any additional HTTP request to CS during copy being processed, Riak CS reads that request and throws error, and cancels the copy.

### DIFF
--- a/src/riak_cs_copy_object.erl
+++ b/src/riak_cs_copy_object.erl
@@ -31,6 +31,9 @@
          copy/4, copy/5,
          copy_range/2]).
 
+%% Do not use this
+-export([connection_checker/1]).
+
 -include_lib("webmachine/include/webmachine.hrl").
 
 %% @doc tests condition and permission on source object:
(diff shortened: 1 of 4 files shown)


src/riak_cs_copy_object.erl
```diff
@@ -189,15 +196,24 @@ get_content_md5({_MD5, _Str}) ->
...
-            riak_cs_get_fsm:stop(GetPid),
-            riak_cs_put_fsm:finalize(PutPid, MD5);
-        {chunk, Block} ->
-            riak_cs_put_fsm:augment_data(PutPid, Block),
-            get_and_put(GetPid, PutPid, MD5)
+-spec get_and_put(pid(), pid(), list(), fun(() -> boolean()))
+                 -> {ok, lfs_manifest()} | {error, term()}.
+get_and_put(GetPid, PutPid, MD5, ContFun) ->
+    case ContFun() of
+        true ->
```
shino: A trailing white space here.

kuenishi:
Add note: detection only with `inet:peername/1` didn't work for me, with [this server](https://gist.github.com/kuenishi/c1363dd5021c29a0292a) , with my linux. That is why using `gen_tcp:recv/2` for disconnect detection. Maybe peername is cached inside the socket?

kuenishi:
Fixed :)

shino:
@borshop merge

PR #933, opened 2014-07-30
Fix pbc leak in cleaning orphan MPs

shino:
Forgot to close riak_client processes in cleaning orphan MPs.
Key commit of this fix is `fe8ecb0`.

Other two commits are for:
- Propagating errors to shell script caller and exit 1 and
- Adding riak_test for regression.


kuenishi:
@borshop merge

Issue #931, opened 2014-07-30
Update Release Note


shino:
- [x] #924 
- [x] #929 
- [x] `proxy_get` in multibag
- [x] put copy not in oos API

kuenishi:
added two from from #928

PR #934, opened 2014-07-30
Update release notes to the latest bugfix

kuenishi:
Addressed by #931 .

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -36,7 +36,7 @@
 * Fix inconsistent ETag on objects uploaded by multipart [riak_cs/#855](https://github.com/basho/riak_cs/issues/855)
 * Fix policy version validation in PUT Bucket Policy [riak_cs/#911](https://github.com/basho/riak_cs/issues/911)
 * Fix return code of several commands, to return 0 for success [riak_cs/#908](https://github.com/basho/riak_cs/issues/908)
-
+* Fix `{error, disconnected}` repainted with notfound [riak_cs/#929](https://github.com/basho/riak_cs/issues/929)
 
 ## Notes on Upgrading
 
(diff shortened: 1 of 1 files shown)


RELEASE-NOTES.md
```diff
@@ -57,9 +57,13 @@ the bucket with same name. This was fixed by:
 Note that a few operations are needed after upgrading from 1.4.x (or
 former) to 1.5.0.
 
-- run `riak_cs_console:cleanup_orphan_multipart/0` or
-  `riak_cs_console:cleanup_orphan_multipart/1` in an attached console
-  to cleanup all buckets
+- run `riak-cs-admin cleanup-orphan-multipart` to cleanup all
+  buckets. It would be safer to specify timestamp with ISO 8601 format
```
shino, line 61: I don't understand the meaning of "safer". Is there dangerous interval?

If the timestamp argument is preferable, is it useful for users to recommend which time is to be specified. (for example, the time when you finished rolling upgrade all nodes)

RELEASE-NOTES.md
```diff
@@ -115,6 +119,10 @@ make sure GC is not running while upgrading.
   See [#932](https://github.com/basho/riak_cs/pull/932) for further
   information.
 
+* Copying objects in OOS interface is not implemented.
+
+* Multibag is added as Enterprise feature, but it is in early preview
+  status. It won't work well with a cluster with `proxy_get` setup.
```
shino: `s/a cluster/clusters/` ? (because MDC requires more than one clusters)

kuenishi:
I think I fixed the sentences. And I want to ask @lucperkins for 2nd review, to let him know updates in release notes.

RELEASE-NOTES.md
kuenishi, line 61: From hipchat:

```
(15時55分26秒) Shunichi Shinohara: list buckets -> get bucket and confirm delete-marker -> deleting MPs ...(cont) -> user create bucket and init MP (at 1.4.x nodes) -> (cont) delete just inited MP
(15時55分35秒) Shunichi Shinohara: broken, true
```

shino, line 61: This happens in mixed cluster with <1.5.0 and 1.5.0.
Should we add STRONG CAUTION that the script must not be executed in the cluster with < 1.5.0 nodes?

shino, line 61: Again from chat:

```
Kota Uenishi
4:02 PM
fmmm, I don't think so, because a lot of cautions happened before that sentence.
```

shino:
+1 from me :relaxed:

RELEASE-NOTES.md
kuenishi, line 61: Moreover, if cleanup ran in mixed env and things messed up, just upgrade to 1.5 all nodes and re-run the script. Temporarily CS may return 409 or 500 to clients on deletion or creation, if multipart remains in the bucket.

shino, line 61: Yes, only if Stanchion is already upgraded to 1.5.0.

kuenishi, line 61: Yes, and if Stanchion is not yet upgraded, just buckets could be deleted/created. Just upgrade it.

kuenishi:
I changed my plan; if 2nd review had problem, I want the reviewer to open another issue, because there are other updates on release notes merged at other PRs.

kuenishi:
@borshop merge

PR #938, opened 2014-08-04
Clean up edoc generation

reiddraper:
Run `make docs` on the previous commit to see all of the individual
error messages. There are several basic types of changes:
- Fix `foo` syntax to `foo'
- Put specs on top of their functions. Otherwise edoc complains that
- there are multiple spec definitions for a function
- Properly escape code that edoc was interpreting  as XML
- Make sure module documentation comes before `-module(..).`

`rm -rf doc && make docs; echo $?` should now exit with 0, and have no
warnings or errors printed.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -4,6 +4,9 @@
 
 {cover_enabled, true}.
 
+%% EDoc options
+{edoc_opts, [preprocess]}.
+
 {lib_dirs, ["deps", "apps"]}.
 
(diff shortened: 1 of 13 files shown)


reiddraper:
Minor note: I rebase and force-pushed over an older version of this.

kuenishi:
+1 2819121

reiddraper:
@borshop merge

PR #942, opened 2014-08-06
Import release notes correction from Jimmy's review

kuenishi:
Thanks to @jhmullen


kuenishi:
oh, no this should be against release/1.5

Issue #927, opened 2014-07-25
Review bucket restrictions and limitations


kuenishi:
"Each AWS account can own up to 100 buckets at a time."
 And make sure other limitations to protect the system.
http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html

shino:
It may be nice to make the limit as configuration item like `unlimited | non_neg_integer()`.

kuenishi:
Other restrictions seems to be only in bucket names. So bucket number limitation is the only missing.

PR #950, opened 2014-08-12
Add bucket number limit check in Riak CS process

kuenishi:
The size of an user record is proportional to bucket number. If too
many buckets were created, then the user's record gets large and
access to anything owned by the user gets slow. This commit adds
bucket number limitation in bucket creation, to prevent user record to
grow too large. As this is not a strict check in a critical section,
there might be a race condition in multiple bucket creations where
they validate total number of owned buckets and then send requests to
Stanchion. But bucket creation is a rare operation, this won't happen
so frequently.

The case where this commit can't prevent growth of an user object is,
when buckets are frequently created and deleted; in that case the user
record may have many deleted bucket records. But that will be pruned.

Addressed by #927

### DIFF
--- a/dialyzer.ignore-warnings.ee
+++ b/dialyzer.ignore-warnings.ee
@@ -1,6 +1,6 @@
 # Errors
 Function get_block_remote/5 has no local return
-riak_cs_config.erl:211: Function will never be called
+riak_cs_config.erl:212: Function will never be called
 # Warnings
 Unknown functions:
   app_helper:get_prop_or_env/3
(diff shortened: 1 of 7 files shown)


kuenishi:
TODO: add tests

shino:
Can we set `unlimited`?

kuenishi:
`100 =< unlimited` is always true (applies to any integers). I found it jut now.

shino:
Nice :smile:

shino:
Tight corner case.
When  `disable_local_bucket_check=true` and a user has 100 buckets then request to create bucket with existing bucket name results in 400 TooManyBuckets.
If `disable_local_bucket_check=false` (default), the status code is 200.

A workaround is ... delete one bucket :)
If this is hard to fix, then I think the workaround is acceptable.

kuenishi:
Added my workaround. Allow 101st bucket like S3 and error on 102nd bucket creation. Also, I cannot add disable-local-bucket-check `orelse` because we don't do bucket number check in Stanchion.

kuenishi:
Reverted my workaround. If allowing 101st bucket needed, just set `{max_buckets_per_user, 101}` in app.config

shino:
@borshop merge

PR #956, opened 2014-08-20
Use new UUID for seed of canonical ID instead of secret

shino:
Canonical IDs are not hidden information.
Worry about nothing, but better not to use secret (hidden information) to generate it if unnecessary.

This is NOT an security issue because secret is hased by md5 even before this PR.


kuenishi:
@borshop merge

PR #958, opened 2014-08-22
Single user mode

kuenishi:
When all the access goes to single bucket (and single user) then getting user record and bucket object from Riak might be bottleneck, because every single operation is serialized there. This patch makes a new separate mode that does not use `moss.buckets` nor `moss.users`, but takes them from `app.config`. Also,  if single user mode is enabled there is no need for stanchion.

Limitations are: no ACLs, policies can be modified at runtime. No users and no buckets can be created on runtime. Enabled APIs are: PUT,GET,DELETE Object, GET Bucket. And pre-defined canned ACLs. This mode cannot be mixed with existing Riak CS setup (or, if very carefully configured, maybe).

To enable this mode, add lines like this to `app.config`:

``` erlang
              {single_user_mode, enabled},
              {single_users_buckets,
               [
                "su-flat-bucket",
                {"su-acl-public-read", [{canned_acl, "public-read"}]},
                {"su-acl-public-rw", [{canned_acl, "public-read-write"}]}
               ]
              },
```

The only accessible user is admin, defined via `admin_key` and `admin_secret`.

### DIFF
--- a/dialyzer.ignore-warnings.ee
+++ b/dialyzer.ignore-warnings.ee
@@ -1,6 +1,6 @@
 # Errors
 Function get_block_remote/5 has no local return
-riak_cs_config.erl:212: Function will never be called
+riak_cs_config.erl:240: Function will never be called
 # Warnings
 Unknown functions:
   app_helper:get_prop_or_env/3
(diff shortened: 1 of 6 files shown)


kuenishi:
TODO: 1. discuss should this be merged to mainstream, 2. add riak_test, 3. please bors

shino:
Ping in repo cleanup sweeping.
This PR is inactive for long. Is it worth to keep open
or rebase to develop?

PR #959, opened 2014-08-22
Add sleep after update manifests to avoid sibling explosion

shino:
Another candidate solution #882 of based on the @kuenishi 's idea (see also #905).

Idea is simple:
sleep some interval after get-modify-update manifests depending on #(siblings)

knobs
- `manifest_siblings_bp_threashold` (default 5)
  
  threshold to trigger backpressure by sleep
- `manifest_siblings_bp_coefficient` (default 200 [msec])
  
  Coefficient of sleep interval, which is `manifest_siblings_bp_coefficient * #(siblings)`
- `manifest_siblings_bp_max_sleep` (default 30*1000 [msec] = 30sec)
  
  Max sleep interval. Mean sleep interval is determined by minimum of this value and the above.

Actual sleep interval is uniformly randomized between
(0.5 \* mean sleep interval, 1.5 \* mean sleep interval).

### DIFF
--- a/src/riak_cs_manifest_fsm.erl
+++ b/src/riak_cs_manifest_fsm.erl
@@ -299,35 +299,15 @@ get_and_delete(RcPid, UUID, Bucket, Key) ->
     end.
 
 get_and_update(RcPid, WrappedManifests, Bucket, Key) ->
-    %% retrieve the current (resolved) value at {Bucket, Key},
-    %% add the new manifest, and then write the value
-    %% back to Riak
-    %% NOTE: it would also be nice to assert that the
-    %% UUID being added doesn't already exist in the
-    %% dict
     case riak_cs_manifest:get_manifests(RcPid, Bucket, Key) of
         {ok, RiakObject, Manifests} ->
-            NewManiAdded = riak_cs_manifest_resolution:resolve([WrappedManifests, Manifests]),
-            %% Update the object here so that if there are any
-            %% overwritten UUIDs, then gc_specific_manifests() will
-            %% operate on NewManiAdded and save it to Riak when it is
-            %% finished.
-            ObjectToWrite0 = riak_cs_utils:update_obj_value(
-                               RiakObject, riak_cs_utils:encode_term(NewManiAdded)),
-            ObjectToWrite = update_md_with_multipart_2i(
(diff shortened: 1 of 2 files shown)


shino:
Did some measurement.

Details are in [1] and [2]

Setup:
- 4-node riak cluster + 1 riak-cs
- gof3r - riak-cs - riak in single box,
  
  other 3 riak nodes are other 3 boxes
- Scope: investigate behaivior up to 80 or 160 concurrency
- Upload 2GB file by MP, part size 5MB = 5242880 bytes
- S3 client is gof3r (in s3gof3r), with patch to change timeout from 5sec to 300sec
- target branches: `feature/mp-backpressure` and `feature/mp-backpressure-simple-sleep`

Summary:
- Both `feature/mp-backpressure`  and `feature/mp-backpressure-simple-sleep` are
  better behaivior (=smaller siblings) above concurrency 40, than `release/1.5`.
- `feature/mp-backpressure-simple-sleep` is slightly better (shorter) elapsed time
  than `feature/mp-backpressure` at concurrency 80.
- `feature/mp-backpressure`, which has retry logic, managed to retry even
  after client already closed the connection. It leads unnecessary load.
- `feature/mp-backpressure-simple-sleep` with `manifest_siblings_bp_max_sleep`
  changed to 30 sec suppress numbers of siblings at concurrency 160.
  Although, it leads to 50+ siblings.

[1] https://gist.github.com/shino/2c4079d4e489ccd57ab2
[2] https://gist.github.com/shino/4617a628408537ef6a98
[3] https://github.com/rlmcpherson/s3gof3r/tree/master/gof3r

shino:
Extended max sleep interval at f3cbbd9, updated the description of this PR.

kuenishi:
Note that s3gof3r by default has 5 sec retry, which leads to sibling explosion anyway. This patch expects ther clients will retry within longer interval, say, > 30~60 sec.

There is a tradeoff between number of concurrent part uploads, its retry interval. Choose carefully the knobs listed above.

kuenishi:
@borshop merge

shino:
Memo to self:

To disable the backpressure sleep completely:

```
> application:set_env(riak_cs, manifest_siblings_bp_threashold, infinity).
```

PR #963, opened 2014-09-03
Set timeout as infinity, default was 5000ms

kuenishi:
According to [the manpage](http://www.erlang.org/doc/man/gen_server.html#call-2) the default timeout is 5 seconds while Riak PB client could return much slower. That might have been causing unnecessary timeouts under rather high load.

### DIFF
--- a/src/riak_cs_riak_client.erl
+++ b/src/riak_cs_riak_client.erl
@@ -143,17 +143,17 @@ set_manifest_bag(RcPid, ManifestBagId) ->
 %% from this module.
 -spec master_pbc(riak_client()) -> {ok, MasterPbc::pid()} | {error, term()}.
 master_pbc(RcPid) ->
-    gen_server:call(RcPid, master_pbc, infinity).
+    gen_server:call(RcPid, master_pbc).
 
 %% TODO: Also this is cheat
 -spec manifest_pbc(riak_client()) -> {ok, ManifetPbc::pid()} | {error, term()}.
 manifest_pbc(RcPid) ->
-    gen_server:call(RcPid, manifest_pbc, infinity).
+    gen_server:call(RcPid, manifest_pbc).
 
 %% TODO: Also this is cheat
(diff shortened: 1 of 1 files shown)


shino:
Controling actual timeout by riakc_pb_socket look nice :)

There are some calls which do not include riakc_cpb_socket, I wonder how they are controlled.
`infinity` might lead unbound block or too long checkout from pool.

kuenishi:
On `riak_cs_riak_client` ? I believe there might be a case where non-pb_socket-related calls could block 5 seconds and then timeout, especially when the other caller blocked by `riakc_pb_socket` call. The other caller might happen if there are a bad retry logic on top of this.

shino:
riakc_pb_socket calls are (possibly) included in `gen_server:call`s in
there functions:
- get_bucket/2
- set_bucket_name/2
- get_user/2
- save_user/3

Other than these, only process' internal state manipulation and/or
pool checkin/checkout.
Default 5 seconds timeout seems sufficient.

kuenishi:
updated

src/riak_cs_riak_client.erl
```diff
@@ -140,17 +143,17 @@ set_manifest_bag(RcPid, ManifestBagId) ->
 %% from this module.
 -spec master_pbc(riak_client()) -> {ok, MasterPbc::pid()} | {error, term()}.
 master_pbc(RcPid) ->
-    gen_server:call(RcPid, master_pbc).
+    gen_server:call(RcPid, master_pbc, infinity).
```
shino: This does not call riakc_pb_socket, does it?

shino: Same for manifest_pbc and block_pbc

shino:
All riak_test passed.

This PR fixes the bug introduced in 1.5 and brings back the 1.4.x behaivior.
5 seconds is short because user fetch is PR=all, only single slow node make it timeout.

:+1:

shino:
@borshop merge

Issue #827, opened 2014-03-26
Invalid state in GC manifest seen


engelsanchez:
This was observed twice at customer sites. I'm not sure how it got there, but a manifest in the GC bucket did not have the state expected. The following patch was applied to avoid the daemon crashing so that GC could continue:

https://github.com/basho/riak_cs/commit/1c1c6f45d7aac568d8d8d7c125ba28471c10c4e1

jonmeredith:
Customer ticket zd://7336

kuenishi:
This also happened for 1.5.0, upgraded from 1.4. See zd://8612

shino:
memo:

`{stop, normal}` with state `BlockCount=0` replies to a worker (or gc daemon in 1.4.x) with successful message. Then manifests in GC bucket are deleted (precisely speaking `towp_set:delelete_element/2`).
If the manifest with invalid state are actually the one which should be deleted, the blocks belonging to the manifest will become orphan.

shino:
The code to skip invalid state manifests and output warning logs was merged at #964 .
The root cause is not yet identified.

PR #949, opened 2014-08-12
GC may stall due to  riak_cs_delete_fsm deadlock

kellymclaughlin:
GC can stall when a `riak_cs_delete_fsm` worker process encounters a deadlock condition. It is related to the `delete_concurrency` setting, but only has the potential to affect files that were uploaded using multipart upload.

The problem rests on the fact that we treat the `delete_blocks_remaining` field of the object manifest as an ordered set except when that set is initially determined for a manifest that contains metadata indicating it was a multipart upload. The initial assignment for `delete_blocks_remaining` can be seen [here](https://github.com/basho/riak_cs/blob/release/1.5/src/riak_cs_delete_fsm.erl#L247). The code that fails to properly order the list of upload parts is [here](https://github.com/basho/riak_cs/blob/release/1.5/src/riak_cs_lfs_utils.erl#L135).

The reason this is an issue and that `riak_cs_delete_fsm` processes hang around indefinitely is that the `ordsets:del_element` [call is used to remove elements of the `delete_blocks_remaining` set](https://github.com/basho/riak_cs/blob/release/1.5/src/riak_cs_lfs_utils.erl#L333) as the delete workers (whose count is controlled by `delete_concurrency`) respond that their assigned blocks have been deleted. If a call to `ordsets:del_element` is made with an element that does not exist in the set or more imporantly in this case is made on an unordered set where the element for deletion sorts less than any key ahead of it in the the set, the return value is identical to the input set. This can come into play when you have `delete_concurrency` greater than 1 and a block delete worker, call it worker _A_, assigned to delete a block identified by `{UUID, BlockNumber}` return a response to the riak_cs_delete_fsm process before another delete worker, worker _B_, whose assigned `{UUID, BlockNumber}` pair sorts greater than that of worker _A_, but appears before the pair of worker A int the set. More concisely, if the block assigned to worker _A_ sorts less than that assigned to worker _B_, but appears after the block assigned to worker _B_ in the `delete_blocks_remaining` set in the manifest and worker _A_ responds before worker _B_, then the entry for worker _A_ will not be removed from the `delete_blocks_remaining` set . Since the delete fsm relies on `delete_blocks_remaining` being empty as a termination condition, the process will hang forever and GC stalls out.

Here is a snippet of the return value from a call to `riak_cs_lfs_utils:block_sequences_for_manifest` for a multipart-uploaded file:

```
[{<<65,96,107,167,183,167,76,217,141,81,114,254,144,244,52,133>>,0},
[...]

PR #964, opened 2014-09-05
Bugfix: Skip invalid state manifests in GC bucket

shino:
Addresses #827 partly.

At present, the root cause why active state manifests are entered in GC bucket.
This PR does not process such manifests but just skip and log in order to
GC can proceed to other manifests instead of abort.

It's noteworthy that if `riak_cs_delete_fsm` replies ok with the same total blocks and deleted blocks
to `riak_cs_gc_worker`, then the worker attempt to delete manifest entry in `twop_set`.
Doing it _before_ deleting manfiests and blocks completely poses orphan manifests/blocks.
In this PR, `riak_cs_delete_fsm` replies with total blocks = 1 and deleted_blocks = 0.

To address #827 completely, further investigation on root cause needed. Information of logs
by this PR will help.

### DIFF
--- a/riak_test/tests/gc_tests.erl
+++ b/riak_test/tests/gc_tests.erl
@@ -65,11 +65,11 @@ setup_obj(RiakNodes, UserConfig) ->
     erlcloud_s3:delete_object(?TEST_BUCKET, ?TEST_KEY_BAD_STATE, UserConfig),
     %% Change the state in the manifest in gc bucket to active.
     %% See https://github.com/basho/riak_cs/issues/827#issuecomment-54567839
-    Pbc = rtcs:pbc(RiakNodes, gc_bucket, not_used),
-    {ok, GCKeys} = riakc_pb_socket:list_keys(Pbc, ?GC_BUCKET),
+    GCPbc = rtcs:pbc(RiakNodes, objects, ?TEST_BUCKET),
+    {ok, GCKeys} = riakc_pb_socket:list_keys(GCPbc, ?GC_BUCKET),
     BKey = {list_to_binary(?TEST_BUCKET), list_to_binary(?TEST_KEY_BAD_STATE)},
     lager:info("Changing state to active ~p, ~p", [?TEST_BUCKET, ?TEST_KEY_BAD_STATE]),
-    {ok, GCKey, UUID} = change_state_to_active(Pbc, BKey, GCKeys),
+    {ok, GCKey, UUID} = change_state_to_active(GCPbc, BKey, GCKeys),
 
     %% Put and delete some more objects
     [begin
@@ -79,7 +79,7 @@ setup_obj(RiakNodes, UserConfig) ->
          erlcloud_s3:delete_object(?TEST_BUCKET, Key, UserConfig)
      end || {Suffix, Size} <- [{"Z", 0}, {"Y", 150}, {"X", 1}]],
 
-    riakc_pb_socket:stop(Pbc),
+    riakc_pb_socket:stop(GCPbc),
     {GCKey, {BKey, UUID}}.
 
 change_state_to_active(_Pbc, TargetBKey, []) ->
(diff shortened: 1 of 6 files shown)


shino:
Set milestone 1.5.1. If wrong, please correct me.

shino:
This PR is to branch `release/1.5`, especially based on #949 .

If patch needed, another branch (cherry-picked and conflicts resolved) `support/patch-skip-invalid-state-manifest` might be better, depending on whether #949 is necessary or not.
Compare view: https://github.com/basho/riak_cs/compare/1.5.0...support/patch-skip-invalid-state-manifest

kuenishi:
I'm not sure why, but an eqc test didn't work for me:

``` erlang
...riak_cs_gc_single_run_eqc:84: eqc_test_...*failed*
::{function_clause,
    [{riak_cs_gc_single_run_eqc,dummy_start_delete_fsm,
         [testnode247683@nausicaa,
          [<0.27611.1>,
           {<<"UUIDno_error11">>,
            {lfs_manifest_v3,3,1048576,
                {<<"bucket">>,<<"no_e"...>>},
                undefined,"2014-09-09T03:50:22.000Z",undefined,...}},
           <0.27610.1>,<<"no_error1">>,[]]],
         [{file,"test/riak_cs_gc_single_run_eqc.erl"},{line,250}]},
     {riak_cs_delete_fsm_sup,start_delete_fsm,
         [testnode247683@nausicaa,
          [<0.27611.1>,
           {<<"UUIDno_error11">>,
            {lfs_manifest_v3,3,1048576,
                {<<"buck"...>>,<<...>>},
                undefined,
                [...],...}},
           <0.27610.1>,<<"no_error1">>,[]]],
         []},
     {riak_cs_gc_worker,initiating_file_delete,2,
         [{file,"src/riak_cs_gc_worker.erl"},{line,138}]},
     {gen_fsm,handle_msg,7,[{file,"gen_fsm.erl"},{line,494}]},
     {proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,227}]}]}
```

riak_test/tests/gc_tests.erl
```diff
@@ -0,0 +1,139 @@
...
+
+    riakc_pb_socket:stop(GCPbc),
+    {GCKey, {BKey, UUID}}.
+
+change_state_to_active(_Pbc, TargetBKey, []) ->
+    lager:warning("Target BKey ~p not found in GC bucket", [TargetBKey]),
+    {error, notfound};
+change_state_to_active(Pbc, TargetBKey, [GCKey|Rest]) ->
+    {ok, Obj0} = riakc_pb_socket:get(Pbc, ?GC_BUCKET, GCKey),
+    Manifests = twop_set:to_list(binary_to_term(riakc_obj:get_value(Obj0))),
```
kuenishi, line 90: Can we assume `twop_set` is in the path?

shino:
Nice catch :smile: .
Pushed two commits, which address two comments respectively.

src/riak_cs_delete_fsm.erl
```diff
@@ -160,15 +164,31 @@ deleting_state_result(_, State) ->
...
+                                       gc_key=GCKey}) ->
+    case blocks_to_delete_from_manifest(Manifest) of
+        {ok, {NewManifest, BlocksToDelete}} ->
+            BlockCount = ordsets:size(BlocksToDelete),
+            NewState = State#state{manifest=NewManifest,
+                                   delete_blocks_remaining=BlocksToDelete,
+                                   total_blocks=BlockCount},
+            start_block_servers(NewState);
+        {error, invalid_state} ->
+            _ = lager:warning("Invalid state manifest in GC bucket at ~p: ~p",
```
kuenishi: Thinking possible operations when this log line found, that might be backing up the object and try deleting it again. Or restore the whole historical manifests via inspector. Anyway I think it'd be more friendly to print out bucket and key name explicitly. UUID and Manifest could follow them or in a debug line.

kuenishi: I meant the bucket and key as S3 manner. GCKey should also be in the first log line.

src/riak_cs_delete_fsm.erl
```diff
@@ -160,15 +164,31 @@ deleting_state_result(_, State) ->
...
+                              [GCKey, Manifest]),
+            %% If total blocks and deleted blocks are the same,
+            %% gc worker attempt to delete the manifest in fileset.
+            %% Then manifests and blocks becomes orphan.
+            %% To avoid it, set total_blocks > 0 here.
+            {stop, normal, State#state{total_blocks=1}}
```
kuenishi, line 189: Without no prior knowledge, this comments are insufficient and sounds like this should stop with  `{stop, {error, invalid_state}, ...}` . We also have to not about #827 , which is like this:

```
  #827 prevents all other blocks and manifests being collected. 
  For now delete FSM stops with pseudo-normal termination to
  let other valid manifests be collected, as the root cause of #827
  is still unidentified.
```

kuenishi:
Great work. Code itself looks pretty good to me.

src/riak_cs_delete_fsm.erl
shino, line 189: I will take last thre lines from your recommendation.
The reason why I don't want to mention #827 as patch is that it is an issue but not an PR.
Execuse for imcompleteness is good to encourage future refinement :)

shino:
@borshop merge

PR #966, opened 2014-09-09
Remove riak_cs_s3_response dependency from mp_upload_test

kuenishi:
`mp_upload_test' uses riak_cs_s3_response to in one of its assertion.
Not sure why it has passed since now but copying it every compile
time fixes it.


kuenishi:
I believe it's ready now

kuenishi:
@borshop merge

PR #980, opened 2014-09-29
Add logging when recieved {error, disconnected} to diagnose the cause

kuenishi:
This commit adds logging the result of `riakc_pb_socket:is_connected/1`
when Riak CS recieved {error, disconnected} from Riak. This forms a
part of long campaign of connection problem between Riak and Riak CS.
`riakc_pb_socket:is_connected/1` returns `{false, []}` mostly when
TCP connection was closed by Riak - or by Operating System. Sometimes
increasing `somaxconn` or `pb_backlog` helps. If the second element
of the return value, the list of reasons when connecting to Riak
(technically, return value of `gen_tcp:connect` might help diagnosing
the root cause, which is usually outside Riak and Riak CS.

Those logging lines are added to major places where disconneced error
was observed. `riakc_pb_socket` could return disconneced at any call,
potentially, but these logging lines will cover most of them.

[added] This pull request also includes adding switches to change riakc
reconnect and queuing strategy, choosing `queue_if_disconnected` and
`auto_reconnect`, to do trial and errors to know CS's behaviour. The
default behaviour isn't changed.

RIAK-1242

### DIFF
--- a/src/riak_cs_access_archiver.erl
+++ b/src/riak_cs_access_archiver.erl
@@ -207,9 +207,10 @@ store(User, RcPid, Record, Slice) ->
             ok = lager:debug("Archived access stats for ~s ~p",
                              [User, Slice]);
         {error, Reason} ->
+            IsConnected = riakc_pb_socket:is_connected(MasterPbc),
             ok = lager:error("Access archiver storage failed (~p), "
-                             "stats for ~s ~p were lost",
-                             [Reason, User, Slice]),
+                             "stats for ~s ~p were lost. Connection status: ~p",
+                             [Reason, User, Slice, IsConnected]),
             {error, Reason};
         {'EXIT', {noproc, _}} ->
             %% just haven't gotten the 'DOWN' yet
--- a/src/riak_cs_access_archiver.erl
+++ b/src/riak_cs_access_archiver.erl
@@ -207,10 +207,11 @@ store(User, RcPid, Record, Slice) ->
             ok = lager:debug("Archived access stats for ~s ~p",
                              [User, Slice]);
         {error, Reason} ->
-            IsConnected = riakc_pb_socket:is_connected(MasterPbc),
             ok = lager:error("Access archiver storage failed (~p), "
-                             "stats for ~s ~p were lost. Connection status: ~p",
-                             [Reason, User, Slice, IsConnected]),
+                             "stats for ~s ~p were lost.",
+                             [Reason, User, Slice]),
+            riak_cs_pbc:check_connection_status(MasterPbc,
+                                                "riak_cs_access_archiver:store/4"),
             {error, Reason};
         {'EXIT', {noproc, _}} ->
             %% just haven't gotten the 'DOWN' yet
--- a/src/riak_cs_block_server.erl
+++ b/src/riak_cs_block_server.erl
@@ -387,7 +387,7 @@ constrained_delete(RcPid, RiakObject, BlockId) ->
 secondary_delete_check({error, {unsatisfied_constraint, _, _}}, RcPid, RiakObject) ->
     riakc_pb_socket:delete_obj(block_pbc(RcPid), RiakObject);
 secondary_delete_check({error, Reason} = E, _, _) ->
-    _ = lager:warning("Contrained block deletion failed. Reason: ~p", [Reason]),
+    _ = lager:warning("Constrained block deletion failed. Reason: ~p", [Reason]),
     E;
 secondary_delete_check(_, _, _) ->
     ok.
--- a/src/riak_cs_block_server.erl
+++ b/src/riak_cs_block_server.erl
@@ -386,6 +386,9 @@ constrained_delete(RcPid, RiakObject, BlockId) ->
 
 secondary_delete_check({error, {unsatisfied_constraint, _, _}}, RcPid, RiakObject) ->
     riakc_pb_socket:delete_obj(block_pbc(RcPid), RiakObject);
+secondary_delete_check({error, Reason} = E, _, _) ->
+    _ = lager:warning("Contrained block deletion failed. Reason: ~p", [Reason]),
+    E;
 secondary_delete_check(_, _, _) ->
     ok.
 
--- a/src/riak_cs_config.erl
+++ b/src/riak_cs_config.erl
@@ -51,6 +51,8 @@
          set_user_buckets_prune_time/1,
          riak_host_port/0,
          connect_timeout/0,
+         queue_if_disconnected/0,
+         auto_reconnect/0,
          is_multibag_enabled/0,
          max_buckets_per_user/0
         ]).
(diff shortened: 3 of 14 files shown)


shino:
I don't start the review of this PR, but please let me ask one question.

There are some sparrowed errors. Should those are addressed by this PR or another?
Not comprehensive but just in my head now are
- the errors at first attempt of user fetch with strong options [1]
- both first and second attempt of deleting blocks are not logged [2]

(Off the subject: These operation have "strong" (i.e. PR=all) option,
so onle one slow (v)node makes them failed with timeout. I suspect
using storong option is one of root causes of disconnected.)

[1] https://github.com/basho/riak_cs/blob/release/1.5/src/riak_cs_riak_client.erl#L285
[2] https://github.com/basho/riak_cs/blob/release/1.5/src/riak_cs_block_server.erl#L405 https://github.com/basho/riak_cs/blob/release/1.5/src/riak_cs_block_server.erl#L377

kuenishi:
On deletion, CS doesn't try second deletion unless the error reason is constraint not satisfied and the reason is to be forwarded to delete_fsm, which is expected to output log because it's not `normal` stop.

src/riak_cs_access_archiver.erl
```diff
@@ -207,9 +207,10 @@ store(User, RcPid, Record, Slice) ->
             ok = lager:debug("Archived access stats for ~s ~p",
                              [User, Slice]);
         {error, Reason} ->
+            IsConnected = riakc_pb_socket:is_connected(MasterPbc),
```
shino: Only a grance at riakc_pb_socket so maybe I'm wrong.
If `auto_reconnect=true`, riakc_pb_socket seems to reply `{error, timeout}` to the caller and dies.
If it is correct, the call `is_connected` results in `noproc` exception and `Reason` is swallowed (not output to log).

This may not cause an issue if the switch `auto_reconnect`, which becomes configurable by this PR, is only for developers and not for production.

src/riak_cs_block_server.erl
```diff
@@ -386,6 +386,9 @@ constrained_delete(RcPid, RiakObject, BlockId) ->
 
 secondary_delete_check({error, {unsatisfied_constraint, _, _}}, RcPid, RiakObject) ->
     riakc_pb_socket:delete_obj(block_pbc(RcPid), RiakObject);
+secondary_delete_check({error, Reason} = E, _, _) ->
+    _ = lager:warning("Constrained block deletion failed. Reason: ~p", [Reason]),
```
shino, line 390: I love this! We have had almost no idea for this error, even wherther error happened or not. :clap:

src/riak_cs_config.erl
```diff
@@ -307,6 +309,25 @@ connect_timeout() ->
             10000
     end.
 
+%% @doc choose client connection option: true by default
+-spec auto_reconnect() -> [{auto_reconnect, boolean()}].
```
shino, line 313: Why the return type is list? Will just a 2-tuple suffice?

src/riak_cs_config.erl
```diff
@@ -307,6 +309,25 @@ connect_timeout() ->
...
+auto_reconnect() ->
```
shino, line 314: For the default value, `riak_cs_config:get_env/3` is useful, e.g.

```
{auto_reconnect, riak_cs_config:get_env(riak_cs, riakc_auto_reconnect, true)}
```

src/riak_cs_access_archiver.erl
kuenishi: Then how about this?

``` erlang
    IsConnected = catch riakc_pb_socket:is_connected(MasterPbc),
```

src/riak_cs_config.erl
```diff
@@ -307,6 +309,25 @@ connect_timeout() ->
...
+    end.
+
+%% @doc choose client connection option: undefined by default, let
+%% riak-erlang-client choose the default behaviour
+-spec queue_if_disconnected() -> [{queue_if_disconnected, boolean()}].
+queue_if_disconnected() ->
+    case application:get_env(riak_cs, riakc_queue_if_disconnected) of
+        {ok, true} ->  [{queue_if_disconnected, true}];
+        {ok, false} -> [{queue_if_disconnected, false}];
+        _ -> []
```
shino, line 328: Oh, I understand. This is the reason why the return type is list.

src/riak_cs_config.erl
shino, line 313: I understand the purpose. Please ignore this comment.

src/riak_cs_access_archiver.erl
shino: Other workarounds from chat
- `is_connected` helper 
  
  ```
  case AutoReconnect of true -> riakc_pb_socket:is_connected(..); false -> no_reconnect end.
  ```
- Just add a comment "Use only true in production" at `riak_cs_config:auto_reconnect/0`.

src/riak_cs_manifest.erl
```diff
@@ -107,7 +107,15 @@ get_manifests_raw(RcPid, Bucket, Key) ->
     ManifestBucket = riak_cs_utils:to_bucket_name(objects, Bucket),
     ok = riak_cs_riak_client:set_bucket_name(RcPid, Bucket),
     {ok, ManifestPbc} = riak_cs_riak_client:manifest_pbc(RcPid),
-    riakc_pb_socket:get(ManifestPbc, ManifestBucket, Key).
+    case riakc_pb_socket:get(ManifestPbc, ManifestBucket, Key) of
+        {ok, _} = Result -> Result;
+        {error, disconnected} ->
+            _ = lager:warning("Manifest retrieval failed. Client connection status: ~p",
+                              [riakc_pb_socket:is_connected(ManifestPbc)]),
```
shino: I want the word "disconnected" included in the log line because it can help why the error occurs only from log file.
The log format at https://github.com/basho/riak_cs/pull/980/files#diff-b2a2e87f46c1d16e5a9c7e96944d1d89R223 seems nice in this regard.

src/riak_cs_riak_client.erl
```diff
@@ -282,7 +282,8 @@ get_user_with_pbc(MasterPbc, Key) ->
             %% be able to properly resolve conflicts).
             KeepDeletedBuckets = false,
             {ok, {Obj, KeepDeletedBuckets}};
-        {error, _} ->
+        {error, Reason0} ->
+            _ = lager:warning("Fetching user record with strong option failed: ~p", [Reason0]),
```
shino, line 286: Love this too :fireworks:

src/riak_cs_gc_key_list.erl
```diff
@@ -156,6 +156,15 @@ gc_index_query(RcPid, EndTime, BatchSize, Continuation, UsePaginatedIndexes) ->
                     ?GC_BUCKET, ?KEY_INDEX,
                     riak_cs_gc:epoch_start(), EndTime,
                     Options),
+
+    case QueryResult of
+        {error, disconnected} ->
+            _ = lager:warning("GC index query failed. Client connection status: ~p",
+                              [riakc_pb_socket:is_connected(ManifestPbc)]);
```
shino: "disconnected" wanted

src/riak_cs_access_archiver.erl
shino: Testing `auto_reconnect=false` and timeout.
Log (line breaks inserted):

```
00:33:05.310 [warning] Access archiver stopping with work left to do; logs will be dropped
00:33:05.310 [error] gen_server <0.417.0> terminated with reason: disconnected
00:33:05.310 [error] CRASH REPORT Process <0.417.0> with 0 neighbours exited
  with reason: disconnected in gen_server:terminate/6 line 747
00:33:05.310 [error] Supervisor {<0.290.0>,poolboy_sup} had child
   riak_cs_riakc_pool_worker started with {riak_cs_riakc_pool_worker,start_link,undefined}
   at <0.417.0> exit with reason disconnected in context child_terminated
00:33:05.314 [error] gen_fsm <0.10906.6> in state archiving terminated with reason:
   no such process or port in call to gen_server:call(<0.417.0>, is_connected, infinity)
00:33:05.314 [error] CRASH REPORT Process <0.10906.6> with 0 neighbours exited
   with reason: no such process or port in call to
   gen_server:call(<0.417.0>, is_connected, infinity) in gen_fsm:terminate/7 line 611
```

`Reason=timeout` is lost.

kuenishi:
I added a function `riak_cs_pbc:check_connection_status/2` which is for checking and catching noproc.

kuenishi:
Rebased to release/1.5 due to merge conflict.

PR #987, opened 2014-10-08
Feature/log disconnected rebased

kuenishi:
rebased #980 . See that issue to learn the discussion.

### DIFF
--- a/src/riak_cs_gc_key_list.erl
+++ b/src/riak_cs_gc_key_list.erl
@@ -160,8 +160,6 @@ gc_index_query(RcPid, EndTime, BatchSize, Continuation, UsePaginatedIndexes) ->
 
     case QueryResult of
         {error, disconnected} ->
-            _ = lager:warning("GC index query ~p to ~p failed.",
-                              [EpochStart, EndTime]),
             riak_cs_pbc:check_connection_status(ManifestPbc, gc_index_query);
         _ ->
             ok
--- a/src/riak_cs_gc_key_list.erl
+++ b/src/riak_cs_gc_key_list.erl
@@ -151,16 +151,18 @@ gc_index_query(RcPid, EndTime, BatchSize, Continuation, UsePaginatedIndexes) ->
                       []
               end,
     {ok, ManifestPbc} = riak_cs_riak_client:manifest_pbc(RcPid),
+    EpochStart = riak_cs_gc:epoch_start(),
     QueryResult = riakc_pb_socket:get_index_range(
                     ManifestPbc,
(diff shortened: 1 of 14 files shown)


src/riak_cs_pbc.erl
```diff
@@ -28,7 +28,8 @@
          put/3,
          put_with_no_meta/2,
          put_with_no_meta/3,
-         list_keys/2]).
+         list_keys/2,
+          check_connection_status/2]).
```
shino: One extra white space char.

src/riak_cs_gc_key_list.erl
```diff
@@ -151,11 +151,22 @@ gc_index_query(RcPid, EndTime, BatchSize, Continuation, UsePaginatedIndexes) ->
...
     QueryResult = riakc_pb_socket:get_index_range(
                     ManifestPbc,
                     ?GC_BUCKET, ?KEY_INDEX,
-                    riak_cs_gc:epoch_start(), EndTime,
+                    EpochStart, EndTime,
                     Options),
+
+    case QueryResult of
+        {error, disconnected} ->
+            _ = lager:warning("GC index query ~p to ~p failed.",
```
shino: `riak_cs_pbc:check_connection_status` logs of connection status, so this log seems unnecesarry.
General `{error, Reason}` is logged at `eligible_manifest_keys` 's branch `({{error, Reason}, EndTime}, _) ->`.

src/riak_cs_pbc.erl
```diff
@@ -85,3 +86,18 @@ list_keys(PbcPid, BucketName) ->
         {error, _}=Error ->
             Error
     end.
+
+-spec check_connection_status(pid(), term()) -> no_return().
```
shino: From http://www.erlang.org/doc/reference_manual/typespec.html,

>  Some functions in Erlang are not meant to return; either because they define servers or because they are used to throw exceptions as the function below:
> 
> ```
>   my_error(Err) -> erlang:throw({error, Err}).
> ```
> 
> For such functions we recommend the use of the special no_return() type for their "return", via a contract of the form:
> 
> ```
>  -spec my_error(term()) -> no_return().
> ```

This function return `ok`, so just `ok` or `any()` (if one does not care about return values) seems better.
But, I know, you are far better than me in respect to type theory, so please deny/correct me :sweat_smile:

shino:
Seems nice in general. Only a few tiny nitpicks. Great work :crown:

kuenishi:
Updated

kuenishi:
@borshop merge

PR #988, opened 2014-10-09
Add tests for amendment log, and a bit code reformat

kuenishi:
RIAK-1212
- Add a function `riak_cs_access:flush_to_log/2,3` which outputs one access key and value supposed to be put into `moss.access`.
- `flush_to_log/2` folds around all ets entries and outputs all of them to each line, with JSON encoded
- `flush_to_log/3` outputs single key and value to log.
- The test works as follows:
  - write some data to CS
  - kill all Riak process
  - try to flush the access stats
  - check Riak CS log and verifies whether correct log is updated or not.

### DIFF
--- a/riak_test/tests/access_stats_test.erl
+++ b/riak_test/tests/access_stats_test.erl
@@ -111,7 +111,6 @@ verify_stats_lost_logging(UserConfig, RiakNodes, CSNodes) ->
     %% check logs, at same node with flush_access_stats
     CSNode = hd(CSNodes),
     lager:info("Checking log in ~p", [CSNode]),
-    true = rt:expect_in_log(CSNode, "Access archiver storage failed"),
     ExpectLine = io_lib:format("lost access stat: User=~s, Slice=", [KeyId]),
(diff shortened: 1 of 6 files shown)


shino:
I want some description about the change. One with just one or two lines is fine.

kuenishi:
- Add a function `riak_cs_access:flush_to_log/2,3` which outputs one access key and value supposed to be put into `moss.access`.
- `flush_to_log/2` folds around all ets entries and outputs all of them to each line, with JSON encoded
- `flush_to_log/3` outputs single key and value to log.
- The test works as follows:
  - write some data to CS
  - kill all Riak process
  - try to flush the access stats
  - check Riak CS log and verifies whether correct log is updated or not.
- And some code reformatting added.

shino:
Copied the above comment to the PR description.

ksauzz:
- [x] manual test: https://gist.github.com/ksauzz/115ded58df82d39dba9e
- [x] riak_test

kuenishi:
@borshop merge

PR #993, opened 2014-10-15
Bugfix: storage calc on deleted buckets 1.4

shino:
This fixes a bug related to bucket name re-cycle,
storage calc wrongly includes deleted buckets.
An example of steps to reproduce is:
1. Alice creates a bucket foo
2. Alice deletes the bucket foo
3. Bob creates a bucket with the same name foo
4. Bob PUTs an object to the bucket
5. Then storage calc reports the bucket usage to both Alice and Bob

Condition: (A AND (B-1 OR B-2))

A. a user deletes a bucket and another user creates a bucket with the same name.

B-1. recycle of the bucket name is occured in leeway seconds (default: 1 day)
B-2. some nodes in cluster are down before another change to the user occures.

cf. There are another branch for patch on tag 1.4.5:
https://github.com/basho/riak_cs/compare/1.4.5...bugfix/storage-calc-on-deleted-buckets

### DIFF
--- a/riak_test/tests/storage_stats_test.erl
+++ b/riak_test/tests/storage_stats_test.erl
@@ -39,6 +39,9 @@
 -define(BUCKET6, "storage-stats-test-6").
 -define(BUCKET7, "storage-stats-test-7").
 -define(BUCKET8, "storage-stats-test-8").
+
+-define(BUCKET9, "storage-stats-test-9").
+
 -define(KEY, "1").
 
 -define(HIDDEN_KEY, "5=pockets").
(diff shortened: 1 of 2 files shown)


kuenishi:
@borshop retry

kuenishi:
+1

shino:
build failure is irrelevant one.

shino:
If one wants to make a patch to 1.4.5, should pick two files `riak_cs_storage.beam` and `riak_cs_utils.beam`.

PR #995, opened 2014-10-16
Bugfix: storage calc on deleted buckets (porting to 1.5)

shino:
Porting to `release/1.5` of #993.


shino:
Ouch, this goes to develop. My bad.

PR #1011, opened 2014-11-14
Add read_before_last_manifest_write option to avoid sibling explosion

kellymclaughlin:
Add read_before_last_manifest_write option to help avoid sibling
explosion for use cases involving high churn and concurrency on a
fixed set of keys. When sibling explosion occurs the objects stored in
Riak can become very large and severely impair the functioning of the
system. The trade-off in enabling this option is latency penalty of
doing an extra read before the final write of an object's manifest to
Riak; however, for use cases matching the description the minor
latency penalty is preferable to the system to the consequences of
sibling explosion. This option is not needed for Riak CS installations
using a Riak version that is >= 2.0.0 with DVVs enabled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: erlang
-otp_release: R15B01
+otp_release:
+  - R16B02
+  - R15B01
 notifications:
   webhooks: http://basho-engbot.herokuapp.com/travis?key=7408e8628d324bc4b522deee77bf987ecff67b16
   email: eng@basho.com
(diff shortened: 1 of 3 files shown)


bearcage:
Testing with this applied reduces sibling generation on moderate concurrency write loads significantly -- I went from sustaining 8-12 writers across a test cluster 4 up to 24 without unbounded upward sibling growth. GC seems to have a nontrivial impact on sibling counts under these conditions as well, although I'm not certain there's a great strategy available for addressing that.

slfritchie:
Are there any objections to amending the branch to default to `true` behavior before final review & merge?

kellymclaughlin:
> Are there any objections to amending the branch to default to true behavior before final review & merge?

I am not opposed to this. I'll push a change to that effect shortly.

kellymclaughlin:
FYI, the CI failure is due to a dialyzer bug with Erlang R15B01. The test machine is still using that version and I will work with Paul to get that updated soon.

bearcage:
Reviewed — looks good once the CI failure is resolved.

kellymclaughlin:
@borshop merge

PR #1016, opened 2014-11-18
Add caching on user and bucket records; default is false

kuenishi:
Getting user records and bucket records can be bottleneck
under a workload of single user and single bucket, as all
requests are serialized. This does not affect existing codes
without any caching. The TTL is currently 4 seconds hard
coded.


kuenishi:
I can upload rpm package if needed.

PR #1030, opened 2014-12-11
User creation fails when just ~10 users created

kuenishi:
This pull request just includes the test that reproduces the bug, for reporting. Right fix of code and tests will follow.

Fail log: https://gist.github.com/kuenishi/cf174e2bf3d796fcc378


shino:
Is this a bug of test code or of source code?

kuenishi:
I still don't know. Needs investigation.

ksauzz:
According the fail log, the test failed at line 41. That means admin user test failed before running 10 users test.

shino:
Could not reproduce.

Issue #1036, opened 2014-12-17
Wrong suffix check about hostname


tsubosaka:
Riak CS don't check properly hostname ends with cs_root_host.
Riak CS only check host contains cs_root_host and don't check host ends with cs_root_host.

Here is the example request.

```
[vagrant@vagrant-centos65 riak_cs_1_5_2]$ s3cmd ls s3://tmp
2014-12-17 01:49        13   s3://tmp/hello.txt
[vagrant@vagrant-centos65 riak_cs_1_5_2]$ curl  -H "Host: tmp.s3.amazonaws.com" localhost:15018/hello.txt
Hello world!
[vagrant@vagrant-centos65 riak_cs_1_5_2]$ curl  -H "Host: tmp.s3.amazonaws.com.com" localhost:15018/hello.txt
Hello world!
```

Issue #1035, opened 2014-12-17
Support virtual hosting with custom domain


tsubosaka:
S3 is considered value of host header as bucket name, but Riak CS ignore value of host header if it's not subdomain of cs_root_host.

The following two requests are regarded as same request at S3, but Riak CS aren't.

```
GET /homepage.html HTTP/1.1
Host: www.johnsmith.net
```

```
GET /homepage.html HTTP/1.1
Host: johnsmith.net.s3.amazonaws.com
```

So, Supporting this specification is useful for some application like CDN.

PR #1037, opened 2014-12-17
Support virtual hosting with custom domain.

tsubosaka:
This pull request for supporting virtual-hosting with custom domain #1035 , and also fix bug #1036 .
For backward compatibility, it check new paraeter "support_customdomain_for_bucket" is true.

### DIFF
--- a/src/riak_cs_s3_rewrite.erl
+++ b/src/riak_cs_s3_rewrite.erl
@@ -113,20 +113,50 @@ bucket_from_host(HostHeader, RootHost) ->
                      []    -> HostHeader;
                      [H|_] -> H
                  end,
-    extract_bucket_from_host(HostNoPort,
-                             string:rstr(HostNoPort, RootHost)).
+    case application:get_env(riak_cs, support_customdomain_for_bucket) of
+        {ok , true} -> 
+            extract_bucket_from_host(HostNoPort, RootHost , true);
+        _ -> 
+            extract_bucket_from_host(HostNoPort, RootHost , false)
+    end.
 
 %% @doc Extract the bucket name from the `Host' header value if a
 %% bucket name is present.
--spec extract_bucket_from_host(string(), non_neg_integer()) -> undefined | string().
-extract_bucket_from_host(_Host, 0) ->
-    undefined;
-extract_bucket_from_host(_Host, 1) ->
-    undefined;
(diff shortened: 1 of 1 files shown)


kuenishi:
First of all if we support virtual hosting, there should be name server's CNAME checking - and only hostnames with valid DNS record should be accepted. And we need a little bit more discussion on how to validate. Maybe resolving the CNAME and caching the result with TTL would be the answer. And a few comments:
- This PR should point to develop branch
- Before supporting virtual domains we should introduce `Host` header validation.

tsubosaka:
@kuenishi Sorry I missed your comment.

> This PR should point to develop branch

Ok, I retry PR.
And separate pull request into bug fix and feature enhancement.

> Maybe resolving the CNAME and caching the result with TTL would be the answer.
> ・Before supporting virtual domains we should introduce Host header validation.

Yes, but it add a little latency at system. So if we use this validation, this feature should be optional.

shino:
>  if we support virtual hosting, there should be name server's CNAME checking - and only hostnames with valid DNS record should be accepted

Is it a spec of AWS S3? Are there some design/security concerns?

kuenishi:
Yes, my concern is security, like some bad nameserver a.charlie.com CNAME pointing alice.example.com.s3.amazonaws.com. What happens in this case?

shino:
I think nothing wrong happens.

AWS S3's specification for virtual hosting [1] says only about Host
header.  I guess it does not care whether users either use DNS/CNAME
or just use `cURL` and `-H` option manually.

Roughly speaking, S3 treats Host header as bucket name as whole if
it is not `s3.amazonaws.com`. Then it auth/authz as its way (ACL/Policy).
Using CNAME is _just an example_ of how to combine this S3 spec/behavior
with other services or environment.

[1] Virtual Hosting of Buckets - Amazon Simple Storage Service http://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html#VirtualHostingSpecifyBucket

kuenishi:
ok. I think I was sort of caremad.

NoKriK:
Hi,

I'm interested by this feature, any news?

kuenishi:
Stale. If this is strongly needed again, PR should be open against develop branch.

PR #1038, opened 2014-12-18
Reformat webmachine errors to be lager-compatible

angrycub:
Lager has a built in transformation to pretty print webmachine errors,
or any error_logger event starting with "webmachine error"
(https://github.com/basho/lager/blob/master/src/error_logger_lager_h.erl#L186).
It expects two arguments.  The three argument form was crashing lager with a
badmatch error.


kuenishi:
The reason for crash is not only the bad tuple, but also three `~p`s in the format while the argv only has two elements. Could you get rid of one `~p` and update your branch?

kuenishi:
Although it's still nice catch.  Thanks CV.

angrycub:
Updated.  Thanks for seeing the format string issue.

kuenishi:
@borshop merge

Issue #977, opened 2014-09-25
Keys are doubly URL decoded, possibly key collision and list error


shino:
Keys are doubly URL decoded at [1]

This possibly cause:
1. Key collision for originally different keys
2. List objects error even for ASCII only keys (together with #910)
3. Keys in list objects are different from the ones at PUT Objects

Backward compatibility should be considerd in fixing this bug.

---

To illustrate, consider the case one uses a key `%41`
(3 bytes, the first letter is `%` (ASCII 0x25) at the appliction layer.

Use s3curl.pl here to look into TCP layer partially.
At HTTP layer, `%` should be URL encoded to `%25`, so it becomes like this.

```
% S3CURL=.s3curl.15018.alice s3curl.pl --id cs --put a.txt \
     -- -x 127.0.0.1:15018 -s \
     --trace-ascii - \
     'http://test.s3.amazonaws.com/%2541'
== Info: Hostname was NOT found in DNS cache
== Info:   Trying 127.0.0.1...
== Info: Connected to 127.0.0.1 (127.0.0.1) port 15018 (#0)
=> Send header, 302 bytes (0x12e)
0000: PUT http://test.s3.amazonaws.com/%2541 HTTP/1.1
0031: User-Agent: curl/7.35.0
004a: Host: test.s3.amazonaws.com
0067: Accept: */*
0074: Proxy-Connection: Keep-Alive
0092: x-amz-date: Thu, 25 Sep 2014 01:22:22 GMT
00bd: Authorization: AWS AS2NXQAZUZY2MEORZCBK:Z5yTIGMyc12DkKBmLg83ot5e
00fd: Vfg=
0103: Content-Length: 2
0116: Expect: 100-continue
012c:
<= Recv header, 23 bytes (0x17)
0000: HTTP/1.1 100 Continue
=> Send data, 2 bytes (0x2)
0000: a.
== Info: We are completely uploaded and fine
<= Recv header, 17 bytes (0x11)
0000: HTTP/1.1 200 OK
== Info: Server Riak CS is not blacklisted
<= Recv header, 17 bytes (0x11)
0000: Server: Riak CS
<= Recv header, 42 bytes (0x2a)
0000: ETag: "60b725f10c9c85c70d97880dfe8191b3"
<= Recv header, 37 bytes (0x25)
0000: Date: Thu, 25 Sep 2014 01:22:22 GMT
<= Recv header, 26 bytes (0x1a)
0000: Content-Type: text/plain
[...]

Issue #910, opened 2014-07-07
Improvement to handle irregular strings as keyname.


ksauzz:
This is API compatibility issue. If one uses `test%%00` for keyname:
- S3 handles this as `test%%00`
- CS handles this as `test%0x00`

Then, `s3cmd ls s3://bucketname` fails due to XML response including `<Key>hoge%\x00</Key>`.

shino:
It's useful (for searchability) to copy the stack on this error from #974.

```
01:01:04.857 [error] Lager event handler error_logger_lager_h exited with reason
 {'EXIT',{{badmatch,["/buckets/test/objects",{error,{error,function_clause,
   [{xmerl_lib,expand_attributes,["abc",1,[{error,1},{'Key',1},{'Contents',16},{'ListBucketResult',1}]],
    [{file,"src/xmerl_lib.erl"},{line,234}]},{xmerl_lib,expand_element,4,[{file,"src/xmerl_lib.erl"},{line,179}]},{xmerl_lib,expand_content,4,[{file,"src/xmerl_lib.erl"},{line,224}]},{xmerl_lib,expand_element,4,[{file,"src/xmerl_lib.erl"},{line,187}]},{xmerl_lib,expand_content,4,[{file,"src/xmerl_lib.erl"},{line,224}]},{xmerl_lib,...},...]}},...]},...}}
```

And another possibility, `xmerl_lib:expand_element` on top of stack.

```
10:00:44.885 [error] Lager event handler error_logger_lager_h exited with reason
 {'EXIT',{{badmatch,["/buckets/test2/objects",{error,{error,function_clause,
   [{xmerl_lib,expand_content,[<<"ÿ">>,1,[{error,1},{'Key',1},{'Contents',7},{'ListBucketResult',1}],false],
    [{file,"src/xmerl_lib.erl"},{line,216}]},{xmerl_lib,expand_element,4,[{file,"src/xmerl_lib.erl"},{line,180}]},{xmerl_lib,expand_content,4,[{file,"src/xmerl_lib.erl"},{line,224}]},{xmerl_lib,expand_element,4,[{file,"src/xmerl_lib.erl"},{line,187}]},{xmerl_lib,expand_content,4,[{file,"src/xmerl_lib.erl"},{line,224}]},...]}},...]},...}}
```

shino:
Comment about #977.

This error typically happen when key contains non-UTF-8 byte sequence, but it is possible for ASCII-only keys are affected by the bug.

[...]

PR #1040, opened 2015-01-06
Fix inaccurate s3 rewrite rule.

ksauzz:
Riak CS doesn't handle some specific characters in object name well.

see: https://gist.github.com/ksauzz/9f0dc13c9b8cb0f2a85e
related to #910 
##### This PR includes:
- Fix s3 rewrite rule which results in losing backward compability.
- Introduce riak_cs_s3_rewrite_legacy to keep backward compatibility.
- Add regression test for riak_cs_s3_rewrite_legacy.
##### Topics I'd like to discuss
- Is using `test-python` reasonable for regression test ? At least, boto test saved me from my bad codes.
##### How access log format changes after this patches.

before

```
172.17.42.1 - - [07/Jan/2015:08:27:07 +0000] "PUT /buckets/test/objects/path1%2Fpath2%2Fte%2Bst.txt HTTP/1.1" 200 0 "" ""
```

after

```
127.0.0.1 - - [07/Jan/2015:17:28:23 +0900] "PUT /buckets/test/objects/path1%2Fpath2%2Fte%252Bst.txt HTTP/1.1" 200 0 "" ""
```
##### TODO:
- Update releasenote about description the change of rewrite rule, and how to keep backward compatibility.


riak_test/tests/legacy_s3_rewrite_test.erl
```diff
@@ -0,0 +1,95 @@
...
+           {"CS_BUCKET",             ?TEST_BUCKET}],
+    case execute_cmd(Cmd, [{cd, CsSrcDir}, {env, Env}, {args, Args}]) of
+        ok ->
+            pass;
+        {error, Reason} ->
+            lager:error("Error : ~p", [Reason]),
+            error({external_client_tests, Reason})
+    end.
+
+execute_cmd(Cmd, Opts) ->
```
kuenishi: I'd recommend just using `rt_cs_dev:cmd/2` instead.

riak_test/tests/legacy_s3_rewrite_test.erl
```diff
@@ -0,0 +1,95 @@
...
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% ---------------------------------------------------------------------
+
+-module(legacy_s3_rewrite_test).
+
+%% @doc `riak_test' module for testing object get behavior.
```
kuenishi: This is a stale comment from copy source.

src/riak_cs_s3_rewrite_legacy.erl
```diff
@@ -0,0 +1,42 @@
...
+    {Path, QueryString, _} = mochiweb_util:urlsplit_path(Url),
+    riak_cs_s3_rewrite:rewrite_path_and_headers(Method,
+                                                Headers,
+                                                Url,
+                                                mochiweb_util:unquote(Path),
+                                                QueryString).
+
+
+-spec original_resource(term()) -> undefined | {string(), [{term(),term()}]}.
+original_resource(RD) ->
```
kuenishi: I'm afraid this function is not used as `riak_cs_s3_auth` and `riak_cs_s3_response` directly referring `riak_cs_s3_rewrite` module.

kuenishi:
I believe you should update `riak_cs_config:api/0` to handle `riak_cs_s3_rewrite_legacy`, or it does no harm to `riak_cs_wm_common:maybe_create_user/6` ?

kuenishi:
I also believe it is a good opportunity to write eqc tests on rewrite. If you can't afford time for now, I can do it later.

src/riak_cs_s3_rewrite_legacy.erl
ksauzz: Yes, riak_cs_keystone_auth also do that. It looks another issue. IMO, to resolve it totally, we should prepare riak_test for keystone but it can take so long to do it. It's reason I leave it for now.

ksauzz: Oops, my understanding is incorrect. Please ignore previous comment.

riak_test/tests/legacy_s3_rewrite_test.erl
```diff
@@ -0,0 +1,95 @@
...
+
+-export([confirm/0]).
+-include_lib("erlcloud/include/erlcloud_aws.hrl").
+-include_lib("eunit/include/eunit.hrl").
+
+-define(TEST_BUCKET, "legacy-s3-rewrite-test").
+
+confirm() ->
```
kuenishi, line 29: I'd be good if this test includes the exact case where the problem happened, like:
- put key "foo bar" with content "foo bar"
- put key "foo+bar" with content "foo+bar"
- read "foo bar" and "foo+bar" and verify they have different content

kuenishi:
Memo: to preserve old path translation behaviour, users should update app.config as `{rewrite_module, riak_cs_s3_rewrite_legacy}` from default when updating from 1.5.3 or older.

kuenishi:
After upgraded to the latest version and not to preserve old odd behaviour, new key for the old object put as "foo+bar", is "foo bar".

shino:
By removing URL decode at rewrite module, bucket part of path style
access is never decoded.

For example, using GET Bucket request (client can URL encode any characters in path part, `%74` is `t`):

```
s3curl.pl --id cs --get -- -x 127.0.0.1:15018 -i -s 'http://s3.amazonaws.com/%74est2'
```

Current `release/1.5` branch responds with 200 OK and contents:

```
HTTP/1.1 200 OK
Server: Riak CS
Date: Thu, 08 Jan 2015 02:14:39 GMT
Content-Type: application/xml
Content-Length: 253

<?xml version="1.0" encoding="UTF-8"?><ListBucketResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Name>test2</Name><Prefix></Prefix><Marker></Marker><MaxKeys>1000</MaxKeys><Delimiter></Delimiter><IsTruncated>false</IsTruncated></ListBucketResult>%
```

This branch (at commit d1c3d020d332b361daa0b2e5cc041487fd6e247c) responds with error:

```
HTTP/1.1 404 Object Not Found
Server: Riak CS
Date: Thu, 08 Jan 2015 02:14:08 GMT
Content-Type: application/xml
Content-Length: 185

<?xml version="1.0" encoding="UTF-8"?><Error><Code>NoSuchBucket</Code><Message>The specified bucket does not exist.</Message><Resource>/%74est2</Resource><RequestId></RequestId></Error>%
```

src/riak_cs_s3_rewrite_legacy.erl
kuenishi: I'd like you to get rid of this function.

riak_test/tests/legacy_s3_rewrite_test.erl
kuenishi, line 29: Maybe it's about testing legacy code, so I'd say this is enough to test old behaviour.

riak_test/tests/legacy_s3_rewrite_test.erl
kuenishi: I don't like duplicate code ^^;

kuenishi:
As for @shino 's comment, we'll leave it for bucket names bug just fix object keys. If legacy behaviour preferred, just use legacy rewrite module.

shino:
@kuenishi The bug was fixed by  40609c5 (if I understand correctly).

kuenishi:
ah :pray:

kuenishi:
All riak_test has passed as usual. test-riak_cs is still screwed, although.

kuenishi:
the AA

riak_test/tests/legacy_s3_rewrite_test.erl
ksauzz: rt_cs_dev:cmd/2 doesn't look suitable for this case because it doesn't handle exit code, or output log until the command finishes.I'd like to extract these function to rtcs module.

kuenishi: [In master branch of riak_test rtdev:cmd/2 handles the return value](https://github.com/basho/riak_test/blob/master/src/rtdev.erl#L674). We're just stale.

kuenishi: And also in master branch, `rt_cs_dev:cmd/2` does.

kuenishi:
```
quote state     RAW    escaped(1)  escaped(2)    example

                 *                               'baz/foo bar'        'baz/foo+bar'
client app       +--------+                      'baz/foo%20bar'      'baz/foo%2Bbar'
                          |
HTTP                      |
                          |
webmachine                |
before rewrite            |
rewrite                   +------------+         'baz%2Ffoo%2520bar'  'baz%2Ffoo%252Bbar'
after rewrite                          |
                                       |
extract_key      +---------------------+
                 |
                 v
                 *                               'baz/foo bar'        'baz/foo+bar'
```

shino:
Note: The above diagram is only on `objects` part of the path in case of "virtual host style" access. `buckets` part in "path style" and subresource part are different, not doubly escaped at any point of time.

ksauzz:
@borshop merge

shino:
This also fixed #977 which described the case of `%` character and subsequent `[0-9a-zA-Z]{2}` as well as `+` sign (in client application layer).

Basho-JIRA:
Reopening to get on sprint backlog - will close again shortly. 

**_[posted via JIRA by Deborah Rakow]**_

Basho-JIRA:
Re-closing after adding to sprint 

**_[posted via JIRA by Deborah Rakow]**_

Issue #1022, opened 2014-11-21
Disable previous object after backpressure_sleep 


shino:
From the comment https://github.com/basho/riak_cs/issues/1005#issuecomment-61354263

It seems that maybe_backpressure_sleep does harm.
It was added mainly for multipart upload because PUT Part
can be (massively) concurrent.

At normal (non-multipart) put, sleep after add_new_dict is bad idea.
Or (better?) we should set #state.riak_object/manifest to undefined
if we execute backpressure sleep in order not to use the previous result..

kuenishi:
I think this is a good change to add spec to `riak_cs_manifest_fsm:get_and_update/4`.

kuenishi:
All riak_test has passed. Will run sibling explosion test tomorrow.

Basho-JIRA:
Testing to try and get it to appear on Sprint backlog - will close again shortly. 

**_[posted via JIRA by Deborah Rakow]**_

Basho-JIRA:
Re-closing after adding to sprint 

**_[posted via JIRA by Deborah Rakow]**_

PR #1041, opened 2015-01-07
Disable previous riak object after back-pressure sleep is triggered

shino:
When back-pressure sleep is triggered by many siblings, holding
previous object can lead to more siblings. This commit change the
behaivior and discard it.

This setting has effect only when all the following conditions are met:
- Normal PUT Object, no effect for multipart upload
- Overwriting existing object, no effect for new object
- read_before_last_manifest_write=false, this setting was introduced in
  CS 1.5.3 and default is true

This PR addressed #1022 .

### DIFF
--- a/src/riak_cs_manifest_fsm.erl
+++ b/src/riak_cs_manifest_fsm.erl
@@ -314,8 +314,15 @@ get_and_update(RcPid, WrappedManifests, Bucket, Key) ->
         {ok, RiakObject, Manifests} ->
             case update(RcPid, Manifests, RiakObject, WrappedManifests, Bucket, Key) of
                 {ok, _, _} = Res ->
-                    maybe_backpressure_sleep(riakc_obj:value_count(RiakObject)),
-                    Res;
+                    case maybe_backpressure_sleep(riakc_obj:value_count(RiakObject)) of
+                        true ->
+                            %% Backpressure sleep has been triggered. Current object
+                            %% is to be discarded to prevent unnecessary interleaving
+                            %% with other concurrent manifest updates while sleeping.
+                            {ok, undefined, undefined};
+                        false ->
+                            Res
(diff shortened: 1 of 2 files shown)


kuenishi:
I think this is a good change to add spec to `riak_cs_manifest_fsm:get_and_update/4`.

src/riak_cs_manifest_fsm.erl
```diff
@@ -314,8 +314,14 @@ get_and_update(RcPid, WrappedManifests, Bucket, Key) ->
         {ok, RiakObject, Manifests} ->
             case update(RcPid, Manifests, RiakObject, WrappedManifests, Bucket, Key) of
                 {ok, _, _} = Res ->
-                    maybe_backpressure_sleep(riakc_obj:value_count(RiakObject)),
-                    Res;
+                    case maybe_backpressure_sleep(riakc_obj:value_count(RiakObject)) of
+                        true ->
+                            %% Backpressure sleep is triggered. Bacause holding
+                            %% previous object can lead to more siblings, discard it.
```
kuenishi: My comment recommendation is:
"Backpressure sleep has been triggered. Current manifests folded by this FSM is to be discarded to prevent unnecessary interleaving by other concurrent manifest updates while this sleep."

kuenishi:
All riak_test has passed. Will run sibling explosion test tomorrow.

shino:
Thanks for review.
Updated comment and added spec. Ready for review again.

kuenishi:
[Sibling explosion test result](https://gist.github.com/kuenishi/39a25a1da85bab6eabfb) indicates that it endures twice more than before. Compare the number of lines until fail.

Now working on bors, failing on PLT recreation.

kuenishi:
@borshop merge

PR #1042, opened 2015-01-13
Bugfix/disable prev after backpressure sleep 1.5

shino:
Backport of #1041 to `release/1.5` .

Original description (just copied) below

---

When back-pressure sleep is triggered by many siblings, holding
previous object can lead to more siblings. This commit change the
behaivior and discard it.

This setting has effect only when all the following conditions are met:
- Normal PUT Object, no effect for multipart upload
- Overwriting existing object, no effect for new object
- read_before_last_manifest_write=false, this setting was introduced in
  CS 1.5.3 and default is true

This PR addressed #1022 .

### DIFF
--- a/dialyzer.ignore-warnings.ee
+++ b/dialyzer.ignore-warnings.ee
@@ -1,6 +1,6 @@
 # Errors
 Function get_block_remote/5 has no local return
-riak_cs_config.erl:214: Function will never be called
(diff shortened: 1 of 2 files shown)


kuenishi:
@merge

shino:
@borshop merge

PR #1044, opened 2015-01-13
Remove one line from assertion of access log flush command

shino:
`access_stats_test` keeps failing on my laptop because
there is a race whether the line "~b more archives to flush~n" is
output or not.  Remove the line from assertion and use only another
line of "All access logs were flushed.~n".


shino:
@borshop merge

kuenishi:
@borshop merge

PR #1053, opened 2015-01-20
Suppress unnecessary warnings at startup

shino:
This PR removes two kinds of warning logs at riak-cs start-up.
- Warning about `access_archiver_max_workers`, including typo fix
- Warning on creation of new atoms from `riak-erlang-client`


kuenishi:
@borshop merge

PR #1056, opened 2015-01-21
Updating rebar.config to be set purely to tags/SHAs

gcymbalski:
As requested following conversations with Andrew Stone, here's a PR to pin deps to the latest tags/SHAs available in subrepos. Be aware that work may be needed in those subrepos to actually bring in commits that are expected, as previous builds have been against branches.


gcymbalski:
Dur, this was the /one/ project I didn't have a request to do this on.

PR #1067, opened 2015-02-10
Add date time skew check to both v2 and v4 authentication [JIRA: RCS-129]

kuenishi:
1/2 for RCS-118

### DIFF
--- a/src/riak_cs_s3_auth.erl
+++ b/src/riak_cs_s3_auth.erl
@@ -51,7 +51,6 @@ identify(RD,_Ctx) ->
         undefined ->
             identify_by_query_string(RD);
         AuthHeader ->
-            %% Query string auth does not require date/x-amz-date header
             parse_auth_header(AuthHeader)
     end.
 
--- a/src/riak_cs_s3_auth.erl
+++ b/src/riak_cs_s3_auth.erl
@@ -1,6 +1,6 @@
 %% ---------------------------------------------------------------------
 %%
-%% Copyright (c) 2007-2013 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2007-2015 Basho Technologies, Inc.  All Rights Reserved.
 %%
 %% This file is provided to you under the Apache License,
 %% Version 2.0 (the "License"); you may not use this file
@@ -52,20 +52,30 @@ identify(RD,_Ctx) ->
             identify_by_query_string(RD);
         AuthHeader ->
             %% Query string auth does not require date/x-amz-date header
-            Date = riak_cs_wm_utils:extract_date(RD),
-            case riak_cs_wm_utils:check_timeskew(Date) of
-                true ->
-                    parse_auth_header(AuthHeader);
-                false ->
-                    {failed, reqtime_tooskewed}
-            end
+            parse_auth_header(AuthHeader)
     end.
 
 -spec authenticate(rcs_user(), string() | {v4, v4_attrs()}, RD::term(), #context{}) ->
                           ok | {error, atom()}.
-authenticate(User, {v4, Attributes}, RD, _Ctx) ->
+authenticate(User, Signature, RD, Ctx) ->
+    case wrq:get_req_header("authorization", RD) of
+        undefined ->
+            authenticate_1(User, Signature, RD, Ctx);
+        _ ->
+            Date = riak_cs_wm_utils:extract_date(RD),
+            case riak_cs_wm_utils:check_timeskew(Date) of
+                true ->
+                    authenticate_1(User, Signature, RD, Ctx);
+                false ->
+                    {error, reqtime_tooskewed}
+            end
+    end.
+
+-spec authenticate_1(rcs_user(), string() | {v4, v4_attrs()}, RD::term(), #context{}) ->
+                          ok | {error, atom()}.
+authenticate_1(User, {v4, Attributes}, RD, _Ctx) ->
     authenticate_v4(User, Attributes, RD);
-authenticate(User, Signature, RD, _Ctx) ->
+authenticate_1(User, Signature, RD, _Ctx) ->
     CalculatedSignature = calculate_signature_v2(User?RCS_USER.key_secret, RD),
     case check_auth(Signature, CalculatedSignature) of
         true ->
--- a/src/riak_cs_s3_auth.erl
+++ b/src/riak_cs_s3_auth.erl
@@ -24,13 +24,14 @@
 
 -export([identify/2, authenticate/4]).
 
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-export([calculate_signature_v2/2]).
+-endif.
+
 -include("riak_cs.hrl").
 -include("s3_api.hrl").
 -include_lib("webmachine/include/webmachine.hrl").
-
--ifdef(TEST).
--include_lib("eunit/include/eunit.hrl").
--endif.
 
 -type v4_attrs() :: [{string(), string()}].
 
--- a/src/riak_cs_s3_auth.erl
+++ b/src/riak_cs_s3_auth.erl
@@ -43,13 +43,21 @@
 
 -spec identify(RD::term(), #context{}) ->
                       {string() | undefined,
-                       string() | {v4, v4_attrs()} | undefined}.
+                       string() | {v4, v4_attrs()} | undefined} |
+                      {failed, Reason::atom()}.
 identify(RD,_Ctx) ->
     case wrq:get_req_header("authorization", RD) of
         undefined ->
             identify_by_query_string(RD);
         AuthHeader ->
-            parse_auth_header(AuthHeader)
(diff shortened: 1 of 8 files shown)


src/riak_cs_s3_auth.erl
```diff
@@ -133,38 +141,32 @@ parse_auth_v4_header_value([$, | Rest], Acc) ->
...
+                   Expires ++ "\n"
+           end,
+    CMD5 = case wrq:get_req_header("content-md5", RD) of
+               undefined -> [];
+               CMD5_0 ->    CMD5_0
+           end,
+    ContentType = case wrq:get_req_header("content-type", RD) of
+                      undefined -> [];
+                      ContentType0 -> ContentType0
+                  end,
```
kuenishi, line 179: These changes above are just refactoring.

kuenishi:
I believe this is ready for review.

src/riak_cs_s3_auth.erl
```diff
@@ -43,13 +44,21 @@
...
-                       string() | {v4, v4_attrs()} | undefined}.
+                       string() | {v4, v4_attrs()} | undefined} |
+                      {failed, Reason::atom()}.
 identify(RD,_Ctx) ->
     case wrq:get_req_header("authorization", RD) of
         undefined ->
             identify_by_query_string(RD);
         AuthHeader ->
-            parse_auth_header(AuthHeader)
+            %% Query string auth does not require date/x-amz-date header
```
shino: This branch is for Authorization header. Why does this comment mention about query string?

src/riak_cs_s3_auth.erl
```diff
@@ -133,38 +142,32 @@ parse_auth_v4_header_value([$, | Rest], Acc) ->
...
-            ContentType = [];
-        ContentType ->
-            ok
-    end,
+                   undefined -> []; %% TODO: get noisy here?
+                   {Path,QS} -> [Path, canonicalize_qs(v2, QS)]
+               end,
+    Date = case wrq:get_qs_value("Expires", RD) of
+               undefined ->
+                   case proplists:is_defined("x-amz-date", Headers) of
```
shino, line 165: `riak_cs_wm_utils:extract_date(RD)` can be used.

shino, line 165: Oh, I was wrong. The value of `x-amz-date` is ignored.

src/riak_cs_wm_common.erl
```diff
@@ -407,8 +407,11 @@ authenticate({ok, {User, UserObj}}, RD, Ctx=#context{auth_module=AuthMod, submod
     riak_cs_dtrace:dt_wm_entry({?MODULE, Mod}, <<"authenticate">>, [], [atom_to_binary(AuthMod, latin1)]),
     case AuthMod:authenticate(User, AuthData, RD, Ctx) of
         ok ->
-            riak_cs_dtrace:dt_wm_return({?MODULE, Mod}, <<"authenticate">>, [1], [atom_to_binary(AuthMod, latin1)]),
+            riak_cs_dtrace:dt_wm_return({?MODULE, Mod}, <<"authenticate">>, [2], [atom_to_binary(AuthMod, latin1)]),
             {ok, User, UserObj};
+        {error, reqtime_tooskewed} ->
```
shino, line 412: It seems only `riak_cs_s3_auth:identify/2` returns `reqtime_tooskewed` atom.

src/riak_cs_wm_common.erl
```diff
@@ -472,6 +475,10 @@ post_authentication({error, bad_auth}, RD, Ctx, _, _) ->
     %% given keyid was found, but signature didn't match
     _ = lager:debug("bad_auth"),
     riak_cs_wm_utils:deny_access(RD, Ctx);
+post_authentication({error, reqtime_tooskewed} = Error, RD,
```
shino: If above comment is true, this clause is not needed.

src/riak_cs_s3_auth.erl
shino: I (probably) understood :smile: It's OK.

src/riak_cs_s3_auth.erl
```diff
@@ -43,13 +44,21 @@
...
+                      {failed, Reason::atom()}.
 identify(RD,_Ctx) ->
     case wrq:get_req_header("authorization", RD) of
         undefined ->
             identify_by_query_string(RD);
         AuthHeader ->
-            parse_auth_header(AuthHeader)
+            %% Query string auth does not require date/x-amz-date header
+            Date = riak_cs_wm_utils:extract_date(RD),
+            case riak_cs_wm_utils:check_timeskew(Date) of
```
shino, line 65: Just a proposal: How about move this check logic from `identify` to `authenticate`?
Main reason is skewed time errors should be counted for access stats.
But it will be little portion, so feel free to reject this proposal if it complicate the source code.

shino:
Nice work :smile_cat: All riak_test passes.

I want to discuss the comment just above before merge.

src/riak_cs_s3_auth.erl
kuenishi, line 65: Sounds reasonable. I'll send out pull request.

kuenishi, line 65: s/send out/update this/

kuenishi:
Updated and see whether `external_client_tests`

src/riak_cs_s3_auth.erl
```diff
@@ -43,20 +44,38 @@
 
 -spec identify(RD::term(), #context{}) ->
                       {string() | undefined,
-                       string() | {v4, v4_attrs()} | undefined}.
+                       string() | {v4, v4_attrs()} | undefined} |
+                      {failed, Reason::atom()}.
```
shino, line 48: This type is not used now.

src/riak_cs_s3_auth.erl
```diff
@@ -43,20 +44,38 @@
...
 identify(RD,_Ctx) ->
     case wrq:get_req_header("authorization", RD) of
         undefined ->
             identify_by_query_string(RD);
         AuthHeader ->
+            %% Query string auth does not require date/x-amz-date header
```
shino: This comment is not needed.

shino:
Should the last argument `AnonOk` be wildcard?
https://github.com/basho/riak_cs/blob/feature/check-signed-date/src/riak_cs_wm_common.erl#L478-L479

src/riak_cs_s3_auth.erl
kuenishi, line 48: `parse_auth_header/1` may return `{failed, ...}` although it's the second element is not atom.

kuenishi:
I haven't tested AWS behaviour. And looks like `AnonOk` doesn't make sense to me, the whole code. `parse_auth_header/1` is supposed to return `undefined|list()` and looks like never returns boolean. Am I wrong?

shino:
Memo to self
- `anon_ok` was introduced at https://github.com/basho/riak_cs/commit/28c98d975c3318067306ce75d639cdf978cdaf02
- only `riak_cs_wm_buckets` returns `false`, others used default impl which returns `true`

src/riak_cs_s3_auth.erl
shino, line 48: Oops. You are right :+1:

shino:
> parse_auth_header/1 is supposed to return undefined|list() and looks like never returns boolean.

`AnonOk` is not a return value from `parse_auth_header/1`. It's wm callback that
submodules may implement.

> And looks like AnonOk doesn't make sense to me, the whole code.

It looks like almost true.
If `riak_cs_wm_buckets` implements its own `authorize/2` and forbids anonymous access,
we can get rid of `anon_ok` callback completely.

src/riak_cs_wm_common.erl
```diff
@@ -472,6 +475,10 @@ post_authentication({error, bad_auth}, RD, Ctx, _, _) ->
...
+                    #context{response_module = ResponseMod} = Ctx, _, false) ->
+    _ = lager:debug("reqtime_tooskewed"),
+    ResponseMod:api_error(Error, RD, Ctx);
 post_authentication({error, {auth_not_supported, AuthType}}, RD,
                     #context{response_module=ResponseMod} = Ctx, _, _) ->
     %% given keyid was found, but signature didn't match
```
shino: Not for this PR, but would you please remove this comment line? It is simply wrong.

src/riak_cs_wm_common.erl
kuenishi, line 412: `idnetify/2` behaviour has changed not returning timeskew error.

kuenishi:
Looks like `AnonOk` is also generated from keystone authentication. It'd better be rain check.

kuenishi:
This is ready again for review.

shino:
I checked out slightly old riak_cs_multibag repo : `git checkout 7da4f32` for riak_test run.

shino:
All riak_test passed :bangbang:

shino:
@borshop merge

PR #1068, opened 2015-02-12
Feature/align config RCS-111

kuenishi:
c.f. [Configuration Mapping Table between 1.5 and 2.0](https://github.com/basho/riak_cs/wiki/Configuration-Mapping-Table-between-1.5-and-2.0-%5BRFC%5D)

### DIFF
--- a/rel/files/riak_cs.schema
+++ b/rel/files/riak_cs.schema
@@ -155,7 +155,7 @@
 %% @doc The time to retain the block for an object after it has been
 %% deleted.  This leeway time is set to give the delete indication
 %% time to propogate to all replicas.
-{mapping, "gc.leeway_seconds", "riak_cs.leeway_period", [
+{mapping, "gc.leeway_period", "riak_cs.leeway_seconds", [
   {default, "24h"},
   {datatype, {duration, s}}
 ]}.
@@ -247,7 +247,7 @@
 {translation,
  "riak_cs.storage_schedule",
  fun(Conf) ->
-   Keys = cuttlefish_variable:fuzzy_matches(["storage","stats","schedule","$time"], Conf),
+   Keys = cuttlefish_variable:fuzzy_matches(["stats","storage","schedule","$time"], Conf),
    [cuttlefish:conf_get("stats.storage.schedule." ++ Name, Conf)||{Key, Name} <- Keys]
  end
 }.
@@ -271,7 +271,7 @@
 
 %% @doc custom server name at http response header "Server: Riak CS"
 {mapping, "server.name", "webmachine.server_name", [
-  {default, "Riak CS" },
+  {default, "Riak CS"},
   {datatype, string}
 ]}.
 
@@ -545,7 +545,7 @@
 
 %% VM scheduler collapse, part 2 of 2
 {mapping, "erlang.schedulers.compaction_of_load", "vm_args.+scl", [
-  {default, false},
+  {default, "false"},
   merge
 ]}.
 
--- a/rel/files/riak_cs.schema
+++ b/rel/files/riak_cs.schema
@@ -15,7 +15,7 @@
 {validator,
  "valid_host",
  "should be a valid host",
- fun({Host, Port}) -> is_list(Host) andalso Port > 1024 andalso Port < 65536 end}.
+ fun({Host, Port}) -> is_list(Host) andalso 0 < Port andalso Port < 65536 end}.
 
 %% @doc Riak node to which Riak CS accesses
 {mapping, "riak_host", "riak_cs.riak_host", [
--- a/rel/files/riak_cs.schema
+++ b/rel/files/riak_cs.schema
@@ -353,7 +353,7 @@
 
 %% @doc If your Erlang virtual machine supports DTrace (or
 %% user-space SystemTap), set dtrace_support to true.
-{mapping, "dtrace_support", "riak_cs.dtrace_support", [
+{mapping, "dtrace", "riak_cs.dtrace_support", [
   {default, off},
   {datatype, flag}
 ]}.
--- a/rel/files/riak_cs.schema
+++ b/rel/files/riak_cs.schema
@@ -184,7 +184,7 @@
 ]}.
 
 {mapping, "gc.max_workers", "riak_cs.gc_max_workers", [
-  {default, 5},
+  {default, 2},
   {datatype, integer},
   hidden
 ]}.
--- a/rel/files/riak_cs.schema
+++ b/rel/files/riak_cs.schema
@@ -109,7 +109,8 @@
 %%  or retrieval of objects.
 {mapping, "pool.request.overflow", "riak_cs.connection_pools", [
   {default, 0},
-  {datatype, integer}
+  {datatype, integer},
+  hidden
 ]}.
 
 %% @doc Fixed pool size of secondary connection pool which is used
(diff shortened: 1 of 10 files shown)


kuenishi:
@shino @angrycub it's ready for review; could you take a look at this?

rel/files/riak_cs.schema
```diff
@@ -154,7 +156,6 @@
 %% for an object after it has been deleted.
 %% This leeway time is set to give the delete
 %% indication time to propogate to all replicas.
-%% 86400 is 24-hours.
 {mapping, "gc.leeway_seconds", "riak_cs.leeway_seconds", [
```
kuenishi: `leeway_period` because `leeway_seconds = 24h` is somewhat stupid.

rel/files/riak_cs.schema
```diff
@@ -181,7 +180,14 @@
 %% Riak nodes prior to version 1.4.0.
 {mapping, "gc.paginated_indexes", "riak_cs.gc_paginated_indexes", [
   {default, on},
-  {datatype, flag}
+  {datatype, flag},
+  hidden
+]}.
+
+{mapping, "gc.max_workers", "riak_cs.gc_max_workers", [
+  {default, 5},
```
shino: How about changing default value? Five workers can make rather large load. I want to recommend 2 to begin with.

rel/files/riak_cs.schema
```diff
@@ -189,7 +195,7 @@
 %% @doc How often to flush the access log; integer factor of
 %% access_archive_period (1 == once per period; 2 == twice per period,
 %% etc.)
-{mapping, "access.stats.log_flush_factor", "riak_cs.access_log_flush_factor", [
+{mapping, "access.stats.log.flush_factor", "riak_cs.access_log_flush_factor", [
```
shino: There are two kinds of statistics in riak cs, access and storage. If they are grouped, `stats.access.*` is better name.
If not,  `access_stats` is nice for prefix.

rel/files/riak_cs.schema
```diff
@@ -265,7 +270,7 @@
 
 %% @doc custom server name at http response header "Server: Riak CS"
 {mapping, "server.name", "webmachine.server_name", [
-  {default, "Riak CS"},
+  {default, "Riak CS" },
```
shino: The space added?

shino:
cuttlefish 2.0.1 had been tagged. Update?

shino:
Comments which are not for diff in this PR but cuttlefish'ization related.
- `riak_cs.conf` or `riak-cs.conf`? (underscore or hyphen)
  I want to +1 for hyphen because the command `riak-cs` uses hyphen.
- `dev_vars.confg` needs `devrel` flag:
  cf. https://github.com/basho/riak/blob/develop/rel/vars/dev_vars.config.src#L4
- Why is `cs_port` restricted to > 1024?
  If there is no strong reason, operators can use well-known ports at their own risk.
- I want `admin.ip` and `admin.port` is commented out, rather than hidden.
- Is it better to add `admin.` prefix for `anonymous_user_creation`?
  Operators will know this config item is about administration.
- How about introduce prefix `api.`? (or `s3_api.`?)
  e.g. `api.root_host`, `api.trust_x_forwared_for`, `api.region`, ...
- It's better to change `pool.*.overflow`'s hidden (because we don't test them)
- `request_limit` is defined with the prefix `storage.`.
  But `riak_cs.usage_request_limit` affects both storage and access.
- `storage.stats.request_limit` is defined by `{duration, h}`.
  On the other hand, `riak_cs.usage_request_limit` is used as count or just a number
  (not a duration).
  These two are coincide (accidentally) only if `storage.stats.archive_period`
  and `access.stats.archive_period` are `1h`.
- If we align to riak, `dtrace_support` will be `dtrace` as in `riak_core.schema`:
  cf. https://github.com/basho/riak_core/blob/develop/priv/riak_core.schema#L135-L142
- As well as `riak.schema`, it's good to enable `sfwi` and disable `scl`:
  https://github.com/basho/riak/blob/2.0/rel/files/riak.schema#L208-L219

rel/files/riak_cs.schema
kuenishi: I tried to introduce version number here but didn't work well.

kuenishi:
> `riak_cs.conf` or `riak-cs.conf`? (underscore or hyphen)

Changed

> `dev_vars.confg` needs `devrel` flag:

Added

> Why is `cs_port` restricted to > 1024?

Changed to > 0 

> I want `admin.ip` and `admin.port` is commented out, rather than hidden.

Commented out

> Is it better to add `admin.` prefix for `anonymous_user_creation`?

I don't think so. Too much categories..

> How about introduce prefix `api.`? (or `s3_api.`?)

Same above

> It's better to change `pool.*.overflow`'s hidden (because we don't test them)

Hidden.

> `request_limit` is defined with the prefix `storage.`.

Removed prefix and changed to `stats.usage_request_limit`

> `storage.stats.request_limit` is defined by `{duration, h}`.
>  On the other hand, `riak_cs.usage_request_limit` is used as count or just a number  (not a duration).
>  These two are coincide (accidentally) only if `storage.stats.archive_period` and `access.stats.archive_period` are `1h`.

Description updated

> If we align to riak, `dtrace_support` will be `dtrace` as in `riak_core.schema`:

Changed

> As well as `riak.schema`, it's good to enable `sfwi` and disable `scl`:

Added

rel/files/riak_cs.schema
kuenishi: Changed to 2

rel/files/riak_cs.schema
kuenishi: Changed to leeway_period

kuenishi:
Cuttlefish is also updated to 2.0.1.

kuenishi:
TODO: update configuration mapping and open as another pull request.

rel/files/riak_cs.schema
```diff
@@ -150,19 +152,16 @@
...
-%% @doc The number of seconds to retain the block
-%% for an object after it has been deleted.
-%% This leeway time is set to give the delete
-%% indication time to propogate to all replicas.
-%% 86400 is 24-hours.
-{mapping, "gc.leeway_seconds", "riak_cs.leeway_seconds", [
+%% @doc The time to retain the block for an object after it has been
+%% deleted.  This leeway time is set to give the delete indication
+%% time to propogate to all replicas.
+{mapping, "gc.leeway_seconds", "riak_cs.leeway_period", [
```
shino: `gc.leeway_seconds` ?

rel/files/riak_cs.schema
shino: Why not remove the space?

shino:
> riak_cs.conf or riak-cs.conf? (underscore or hyphen)

`cuttlefish_conf` in `vars.config` should also be changed.

rel/files/riak_cs.schema
kuenishi: heh

kuenishi:
Updated. Can we postpone work on +sfwi stuff? Looks like they're not working correctly and maybe involve cuttlefish.

shino:
> Can we postpone work on +sfwi stuff? Looks like they're not working correctly and maybe involve cuttlefish.

:accept:

shino:
All riak_test cases passed.

kuenishi:
@borshop merge

PR #1072, opened 2015-02-18
Yessir riak client, part 1

shino:
YESSIR riak_client part 1

Communication with Riak KV is stubbed out almost completely.
Main purpose is to benchmark/optimize on-memory logic of riak_cs.

_Caution_ Communication with Stanchion is NOT stubbed out.
Stanchion and Riak KV is needed for create user operation, etc.
Also, bucket properties verification at riak-cs startup interacts
with Riak KV actually.

---

Setting in advanced.config:

```
  {riak_client, riak_cs_yessir_riak_client}
```

For user credential, secret should be user key id + "+Secret",
example for s3cfg:

```
access_key = MZ59BPUOELYZNPTEPFR3
secret_key = MZ59BPUOELYZNPTEPFR3+Secret
```

Current Status is as following, not-yet-implemented functionalities
will come with subsequent PRs (hopefully):
- [x] list buckets (only one bucket)
- [x] list object (only one entry)
- [x] get object (10-byte, fixed content)
- [ ] put object
- [ ] delete object
- [x] GC list keys (empty key list)
- [ ] GC (with some keys)
- [x] access/storage stats PUT
- [ ] access/storage GET

### DIFF
--- a/rel/files/advanced.config
+++ b/rel/files/advanced.config
@@ -1,7 +1,31 @@
 [
  {riak_cs,
+  [
+   %% == Extra for easy development ==
+   {enforce_multipart_part_size, false},
+   {proxy_get, enabled},
+   {admin_auth_enabled, false},
 
-   [
+   {riak_client, riak_cs_yessir_riak_client},
+   %% {auth_bypass, true},
+   %% {auth_module, riak_cs_s3_passthru_auth},
 
-   ]}
+   {dummy, value}
+
+  ]},
+
+ {lager,
+  [
+   {colored, true},
+   {colors, [
+             {debug,     "\e[0;36m" }, %% 36 = cyan
+             {info,      "\e[0;35m" }, %% 35 = magenta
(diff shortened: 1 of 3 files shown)


slfritchie:
+1 I heartily approve of this hackery.

One idea might be to encode the object size in the S3 bucket name, e.g. the suffix is digits like "bucket4096" for objects with 4 Kbyte size?

shino:
> One idea might be to encode the object size in the S3 bucket name, e.g. the suffix is digits like 
> "bucket4096" for objects with 4 Kbyte size? 

Must be implemented in subsequent PRs :+1:

shino:
@borshop merge

kuenishi:
You should not change any default config in advanced.config unless we have any other moderate choice.

shino:
Sorry for that. The change was completely my mistake. 
It was removed in current develop.

PR #1075, opened 2015-02-23
Fix inaccurate webmachine log handler setting.

ksauzz:
`webmachine_log_handler` has been renamed to`webmachine_access_log_handler` on current webmachine.


shino:
Is it good idea to note the renaming in release note for who will move `app.config` to `advanced.config` in whole (and will not use `riak-cs.conf`)?

shino:
@borshop merge

PR #1084, opened 2015-03-09
Introduce getopt into riak_cs_inspector [JIRA: RCS-132]

shino:
`getopt` based riak_cs_inspector.

Before this PR, it was hard to change the script even if the change was minimal one.
This enables future enhancement in main line.

Based on @ksauzz 's work.


shino:
This also includes R16B compatibility (warning suppression about `crypto:md5`).

kuenishi:
I'd say command names are ... `users`, `objects`, `buckets` and `blocks`.

kuenishi:
Anyway it's great work. Sooner or later let's improve this as an internal inspection tool for operators.

kuenishi:
@borshop merge

PR #1051, opened 2015-01-19
Add switch to choose listkeys or fold_objects in storage calclation

kuenishi:
Cleaner version of https://github.com/basho/riak_cs/compare/ku-optimize-storage-calc . Maybe the default value can be `true` as it doesn't affect performance.

```
Eshell V5.10.3  (abort with ^G)
(riak-cs@127.0.0.1)1> timer:tc(fun() -> riak_cs_storage:sum_bucket(<<"junk">>) end).
{2771516,
 {struct,[{<<"Objects">>,58457},{<<"Bytes">>,2784575686}]}}
(riak-cs@127.0.0.1)2> application:set_env(riak_cs, use_2i_for_storage_calc, true).  
ok
(riak-cs@127.0.0.1)3> timer:tc(fun() -> riak_cs_storage:sum_bucket(<<"junk">>) end).
{2756002,
 {struct,[{<<"Objects">>,58457},{<<"Bytes">>,2784575686}]}}
```

This experiment was with vanilla Riak - so it doesn't yet have improvement. This needs another optimization like https://github.com/basho/riak_kv/compare/2.0...ku-optimize-mapred-listkeys .

PR #1089, opened 2015-03-11
Add switch to choose listkeys or fold_objects in storage calclation

kuenishi:
Rebased version of and supersedes #1051 (RCS-34).

### DIFF
--- a/src/riak_cs_config.erl
+++ b/src/riak_cs_config.erl
@@ -58,7 +58,8 @@
          max_buckets_per_user/0,
          read_before_last_manifest_write/0,
          region/0,
-         stanchion/0
+         stanchion/0,
+         use_2i_for_storage_calc/0
         ]).
 
 %% Timeouts hitting Riak
@@ -401,6 +402,15 @@ read_before_last_manifest_write() ->
 -spec region() -> string().
 region() ->
     get_env(riak_cs, region, ?DEFAULT_REGION).
+
+%% @doc This switch changes mapreduce input from listkeys to
+%% fold_objects after riak_kv's mapreduce input optimization
+%% merged and enabled. With fold_objects it will have 1/2 IO
+%% when scanning the whole bucket.
+-spec use_2i_for_storage_calc() -> boolean().
+use_2i_for_storage_calc() ->
+    riak_cs_list_objects_utils:fold_objects_for_list_keys()
+        andalso get_env(riak_cs, use_2i_for_storage_calc, false).
 
 %% ===================================================================
 %% ALL Timeouts hitting Riak
--- a/src/riak_cs_config.erl
+++ b/src/riak_cs_config.erl
@@ -409,8 +409,7 @@ region() ->
 %% when scanning the whole bucket.
 -spec use_2i_for_storage_calc() -> boolean().
 use_2i_for_storage_calc() ->
-    riak_cs_list_objects_utils:fold_objects_for_list_keys()
-        andalso get_env(riak_cs, use_2i_for_storage_calc, false).
+    get_env(riak_cs, use_2i_for_storage_calc, false).
 
 %% ===================================================================
 %% ALL Timeouts hitting Riak
--- a/src/riak_cs_storage.erl
+++ b/src/riak_cs_storage.erl
@@ -100,8 +100,12 @@ sum_bucket(BucketName) ->
         ok = riak_cs_riak_client:set_bucket_name(RcPid, BucketName),
         {ok, ManifestPbc} = riak_cs_riak_client:manifest_pbc(RcPid),
         ManifestBucket = riak_cs_utils:to_bucket_name(objects, BucketName),
+        Input = case riak_cs_config:use_2i_for_storage_calc() of
+                    true -> {index, ManifestBucket, <<"$bucket">>, <<0>>, <<255>>};
+                    false -> ManifestBucket
+                end,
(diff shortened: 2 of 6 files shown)


src/riak_cs_config.erl
```diff
@@ -402,6 +403,15 @@ read_before_last_manifest_write() ->
 region() ->
     get_env(riak_cs, region, ?DEFAULT_REGION).
 
+%% @doc This switch changes mapreduce input from listkeys to
+%% fold_objects after riak_kv's mapreduce input optimization
+%% merged and enabled. With fold_objects it will have 1/2 IO
+%% when scanning the whole bucket.
+-spec use_2i_for_storage_calc() -> boolean().
+use_2i_for_storage_calc() ->
+    riak_cs_list_objects_utils:fold_objects_for_list_keys()
```
shino: This setting item is already deprecated one. Should we care about it?

kuenishi: The deprecation plan is to mention in 2.0 release notes and remove the code later, like late 2.0.x or 2.1. That's why I still hold this config.

kuenishi: Actually, this was to ensure Riak is later than 1.4 - having 2i `<<"key">>` index and `return_value=true` in vnode folding. CS can know this thanks to `riak_pb_csbucket`.

src/riak_cs_config.erl
```diff
@@ -402,6 +403,15 @@ read_before_last_manifest_write() ->
...
+        andalso get_env(riak_cs, use_2i_for_storage_calc, false).
```
shino: Index query by `$bucket` is not 2i-ish, to my mind :sweat_smile: 
`fold_objects_for_storage_calc` is more sensible?

kuenishi: The interface is using 2i ... so that's why I named 2i. Better naming are welcomed!

shino: What happens if one set this switch `true` when underlying riak_kv does not support riak_kv#1072?

If it does no harm, then setting default value to `true` is better. If not, current default value `false` is mandatory (unless we have 2.0.6 before cs 2.0 :zap: ).

src/riak_cs_config.erl
shino: OK, understood.

src/riak_cs_storage.erl
```diff
@@ -100,8 +100,12 @@ sum_bucket(BucketName) ->
         ok = riak_cs_riak_client:set_bucket_name(RcPid, BucketName),
         {ok, ManifestPbc} = riak_cs_riak_client:manifest_pbc(RcPid),
         ManifestBucket = riak_cs_utils:to_bucket_name(objects, BucketName),
+        Input = case riak_cs_config:use_2i_for_storage_calc() of
+                    true -> {index, ManifestBucket, <<"$bucket">>, <<0>>, <<255>>};
```
shino: Probably caremad, but `<<255>>` is not the largest binary. For example, `riak_cs_list_objects_fsm_v2` uses 128 bytes of 255 ``binary:copy(255, 128). The one is also not the largest binary, but much safer.

shino:
Will the new configuration item go into cuttlefish schema?

kuenishi:
The reason why I still keep default value of `use_2i_for_storage_calc` as false is just I don't want to change many things at once. But I'd like to change default as true until all Riak <= 2.0.5 updated later version, at least. Maybe timing is good at 1.4 deprecation.

kuenishi:
I don't think we have to introduce this configuration in riak-cs.conf.
With the same reason I want to keep this parameter rather well-unknown. If we change default value, then do really operators have to know it? This is minor optimisation, so we'd better omit the responsibility to explain (or even announce loud) what the configuration exactly is. Just enough and worth for advanced operators.

kuenishi:
- Added very stupid riak_test
- Get rid of unnecessary configuration (fold_objects_for_list_keys)
- Bigger upper bound of 2i input

Ready for review again.

shino:
> The reason why I still keep default value of use_2i_for_storage_calc as false is just I don't want to change many things at once.

Understood. I won't test `use_2i_for_storage_calc=true` with Riak 2.0.5- .

shino:
Not disk intensive, but did micro benchmark on my laptop, with riak_ee 2.0 branch and got similar results as you had.

```
About 830 K keys

> application:set_env(riak_cs, use_2i_for_storage_calc, true).

13:07:46.817 [info] Finished storage calculation in 43 seconds.
13:09:20.953 [info] Finished storage calculation in 42 seconds.
13:10:22.254 [info] Finished storage calculation in 42 seconds.

> application:set_env(riak_cs, use_2i_for_storage_calc, false).

13:12:45.069 [info] Finished storage calculation in 81 seconds.
13:14:04.226 [info] Finished storage calculation in 75 seconds.
13:15:24.673 [info] Finished storage calculation in 78 seconds.

> application:set_env(riak_cs, use_2i_for_storage_calc, true). %% sanity check

13:16:23.934 [info] Finished storage calculation in 41 seconds.
```

shino:
riak_test passed with riak_ee 2.0 branch.

shino:
@borshop merge

shino:
Memo: The change https://github.com/basho/riak_kv/pull/1072 is included in Riak 2.0.6 and 2.1.0+.

PR #1090, opened 2015-03-13
Fix typos in relnote configuration table and source comment

shino:
- Remove non-existent configuration name
- Fix configuration names in release note table


kuenishi:
@borshop merge

PR #1110, opened 2015-03-27
Default number of `gc_max_workers` was not updated

kuenishi:
Even though we claimed at release notes.


kuenishi:
Looks like @borshop is flaky because github oftenly refusing port 22 frequently.

kuenishi:
Manually merging... as I checked xref and dialyzer and it was successful.

Issue #1109, opened 2015-03-27
Missed items in riak-cs.conf [JIRA: RCS-162]


kuenishi:
Missed 2.0.0
- `max_buckets_per_user`
- `gc.batch_size` (`gc_batch_size` in app.config)

kuenishi:
Lacking access log handler is not working :sob:

PR #1114, opened 2015-04-02
Remove disabling access log add add known issues

shino:
Update release note first before fixing bugs.

cf. https://github.com/basho/riak_cs/issues/1109

PR #1115, opened 2015-04-02
Bugfix/missing conf in schema

shino:
- Add switch for access log output
- Add two configs `gc.batch_size` and `max_buckets_per_user` those are missing at 2.0.0

Addresses #1109


rel/files/riak_cs.schema
```diff
@@ -271,6 +285,12 @@
   {datatype, string}
 ]}.
 
+%% @doc Whether to enable access log.
+{mapping, "log.access", "webmachine.log_handlers", [
+  {default, on},
+  {datatype, flag}
+]}.
```
kuenishi, line 292: Let's add unit test for this in `default_config_test` .

rel/files/riak_cs.schema
```diff
@@ -136,6 +136,14 @@
  end
 }.
 
+%% @doc Max number of buckets that each user can create.
+%% If this limit is exceeded, bucket creation will fails
+%% for the specific user.
+{mapping, "max_buckets_per_user", "riak_cs.max_buckets_per_user", [
```
kuenishi, line 142: How can we set `unlimited` here? Just a string?

kuenishi:
Basically the code looks nice, ... and good catch! I also want you to update configuration mapping table in the release notes. +known issues.

rel/files/riak_cs.schema
shino, line 292: How can we do that? It seems that `defualt_config_test` asserts against application environment keys.

rel/files/riak_cs.schema
shino, line 142: oh, forgot about it.

shino:
Will rebase and force push to include changes from #1114

shino:
Thanks for review. Updated, ready for review again.

kuenishi:
Opened cuttlefish case: https://github.com/basho/cuttlefish/issues/188

kuenishi:
Oddness might be wrong word for this. Sorry!

shino:
+1 for your commit :smile:

kuenishi:
running rt now

kuenishi:
I think no degradation can happen with this PR, caremad.

shino:
I could not find the bug you found, it was because I skipped r_t (and `generate` before that) :hear_no_evil:

kuenishi:
@borshop merge

PR #1118, opened 2015-04-06
Rough sketch of user quota and bandwidth limiting

kuenishi:
- Still no tests
- only applies for object and object part uploads / gets
- `riak_cs_quota` is a behaviour where programmers can make thier own module (see callback definition)
- `riak_cs_simple_quota` is a simple quota limiter which takes care of Object puts
- `riak_cs_bw_limitter` is a stupid bandwidth / access rate limitter on objects
- they are only controllable from attached console: type `riak_cs_simple_quota:help()` for detail usage
- to use them, write `{quota_modules, [riak_cs_simple_quota, riak_cs_bw_limitter]}` in `advanced.config`
- to change quota modules, config change and restart are required
- All data size units are in **Bytes**.

For @joecaswell 's review.

Remaining works or notes on usage:
- separate owner and accessor case: bandwidth and access is for accessor, whild owner is for usage
- retrieving user's latest storage usage from `moss.storage` won't scale for many CS nodes
- add tests
- Add some cool http headers
- Multibag


kuenishi:
# Quota

simple idea of quota
- Bahaviour named `riak_cs_quota` that define callback spec
- Plugin-architecture
- Live users' usage, access and bandwidth monitoring
- Limits of users' usage, access and bandwidth

Use cases
- Stop upload as Quota over
- Stop upload as too much bandwidth
- Stop upload as too much request rate, rate limitation

-> three infos stored in moss.users? no nah
- Quota => on bucket owners

*\* For usage: cluster-wide / soft quota / both user and bucket level / persistent state
*\* For bandwidth and counts: Node-wide / soft quota / both user and bucket level / non-persistent
- check hook point is right after policy chedck
- update hook point is right after in riak_cs_access_log_handler

kuenishi:
Other notification style on rate limiting / quota: below are all quotations -

## [Twitter API](https://dev.twitter.com/rest/public/rate-limiting)

Note that these HTTP headers are contextual. When using app-only auth, they indicate the rate limit for the application context. When using user-based auth, they indicate the rate limit for that user-application context.
- X-Rate-Limit-Limit: the rate limit ceiling for that given request
- X-Rate-Limit-Remaining: the number of requests left for the 15 minute window
- X-Rate-Limit-Reset: the remaining window before the rate limit resets in UTC epoch seconds

When an application exceeds the rate limit for a given API endpoint, the Twitter API will now return an HTTP 429 “Too Many Requests” response code instead of the variety of codes you would find across the v1’s Search and REST APIs.

## [Foursquare](https://developer.foursquare.com/overview/ratelimits)

Please note that rate limits are not per endpoint, but per top-level endpoint group. For example, in an hour, if you make 2,500 userless requests to venues/search and 2,500 userless requests to venues/explore, you would have exhausted your venues/\* rate limits for that hour; however, you would still have 500 requests left for tips/*.

Generally rate limits follow the rules above; however, some endpoints may have different limits due to security or other considerations. For the most current and accurate rate limit usage information for a particular request type, inspect the X-RateLimit-Remaining and  X-RateLimit-Limit HTTP headers of API responses.

If you are currently over limits, our API will return a 403 error, and the response object returned by our API will be empty. We will also include a X-RateLimit-Reset header in the response, which is a timestamp that corresponds to when your rate limits will reset.

## [Etsy](https://www.etsy.com/developers/documentation/getting_started/api_basics)

Headers: Every API response bears rate limit headers, as described below:

```
X-RateLimit-Limit: 10000
X-RateLimit-Remaining: 9924
```

These default rate limits are subject to change at any time, without notice.

## [Calais](http://www.opencalais.com/forums/geek-speak-technical-usage-and-issues-questions-and-hints/usage-quota-error-code)

kuenishi:
- please dialyzer
- [Change status code to 429](http://tools.ietf.org/html/rfc6585#section-4) for bandwidth / rate limit
- [Maybe change status code to 507](http://tools.ietf.org/html/rfc4918#section-11.5) for full quota?

They're done.

kuenishi:
r_t passed for me. I believe it's ready for review.

src/riak_cs_quota.erl
```diff
@@ -0,0 +1,207 @@
...
+
+%% @doc Behaviour for Quota / QoS hook point, to save CS systems by
+%% limiting loads. The idea of Quota and QoS system is, to limit
+%% things by pluggable modules. This is because requirements vastly
+%% varies and it is too hard to cover all of them with single
+%% implementation.
+%% This is experimental feature. These interfaces and behaviousr
+%% may or may not change until general availability.
+%%
+%% To use 'inherited' your own module, it should be configured in
```
shino, line 30: What does `inherited` mean?

src/riak_cs_quota.erl
```diff
@@ -0,0 +1,207 @@
...
+%% are evaluated every time 1) incoming requests passed every other
+%% authentication, authorization works (including ACL, Policy), and 2)
+%% all incoming requests successfully finished even if those requests
+%% resulted in 50x or whatever.
+%%
+%% 1) is a callback called `allow/4' and 2) is `update/2` callback.
+%% These callbacks are called in each process context, 1) is at
+%% mochiweb acceptor and 2) is webmachine log handler.  Other
+%% callbacks are for maintenance or management, which is called at the
+%% context of gen_server process, a child process of
```
shino: Which gen_server? Should callback implement gen_server behavior?

src/riak_cs_quota.erl
```diff
@@ -0,0 +1,207 @@
...
+%% `riak_cs_quota_sup'. Note that these processes can't be added nor
+%% removed at runtime, please just control such enabled/disabled
+%% status within plugin modules.
+%%
+%% Modules may return one of these values as defined in callback:
+%% - ok, will return 20x
+%% - disk quota exceed, returns 507
```
shino: If user should not break quota, response status should be 4xx. I don't know about 507 but it's server's error because it starts with 5 (just a guess, the situation for the error is when server's disk is full but user is not bad).

src/riak_cs_quota.erl
```diff
@@ -0,0 +1,207 @@
...
+%% These callbacks are called in each process context, 1) is at
+%% mochiweb acceptor and 2) is webmachine log handler.  Other
+%% callbacks are for maintenance or management, which is called at the
+%% context of gen_server process, a child process of
+%% `riak_cs_quota_sup'. Note that these processes can't be added nor
+%% removed at runtime, please just control such enabled/disabled
+%% status within plugin modules.
+%%
+%% Modules may return one of these values as defined in callback:
+%% - ok, will return 20x
```
shino: Can callback modules specify response headers and/or response body (or part of it)?

src/riak_cs_quota.erl
```diff
@@ -0,0 +1,207 @@
...
+%% - disk quota exceed, returns 507
+%% - bandwidth exceed, returns 429
+%% - access rate exceed, returns 429
+%%
+%% Known limitations:
+%% - currently this is only for objects.
+%% - separate owner and accessor case: bandwidth and access should be
+%%   for accessor, whild owner is for usage
```
shino: Nice catch for corner case! One possible example for bandwidth should be counted for owner: public readable objects.
It may be better to pass both access user and owner user to callback (and riak cs does not know how they are cooked.)

src/riak_cs_quota.erl
```diff
@@ -0,0 +1,207 @@
...
+%% - no integration tests, please report bugs if you find any!
+%% - No cool configuration systems
+%% - No cool operation console: use `riak-cs console' and send msg to processes
```
shino: s/console/attach/ ?

src/riak_cs_quota.erl
```diff
@@ -0,0 +1,207 @@
...
+%% - Multibag
```
shino, line 64: What kind of limitations?

src/riak_cs_quota.erl
```diff
@@ -0,0 +1,207 @@
...
+         update_all_states/2,
+         get_local_state/1]).
+
+-export([start_link/1, invoke_callback/4, update/2]).
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2,
+         terminate/2, code_change/3]).
+
+%% Callbacks as a behaviour
+
+%% @doc initalize a global state of a quota instance. This state is
```
shino, line 78: initialize

src/riak_cs_quota.erl
shino: OK. Callback should implement gen_server behavior.

src/riak_cs_quota.erl
kuenishi, line 30: Behaviour implementation.

src/riak_cs_quota.erl
kuenishi: No. This is actually a gen_server, but this is just a process.

src/riak_cs_quota.erl
kuenishi: Yes. It can update RD and Ctx as you like.

src/riak_cs_quota.erl
kuenishi: Yes, but fwiw it's left for callback writers who has whole RD and Ctx.

src/riak_cs_quota.erl
kuenishi: nice

src/riak_cs_quota.erl
kuenishi, line 64: Actually this is for `riak_cs_simple_quota`, it just knows master pool, to get user stats. Maybe that's fair enough? If so, I just haven't tested it yet.

Another use cases is to set quotas per bags, depending on cluster status (healthy, new, empty, unstable, obsolete).

src/riak_cs_simple_quota.erl
```diff
@@ -0,0 +1,218 @@
...
+%%
+%% reset()
+%%
+%%   Resets current quota cache.
+%%
+%% --
+%% Known limitations (TODO):
+%% - user's latest storage usage from moss.storage won't scale
+%%   for many CS node
+%% - no integra5tion tests
```
shino: Leet or typo?

src/riak_cs_bw_limitter.erl
```diff
@@ -0,0 +1,197 @@
...
+-spec reset() -> true.
+reset() ->
+    {ok, #state{table=Table}} = riak_cs_quota:get_local_state(?MODULE),
+    ets:delete_all_objects(Table).
+
+-spec init() -> {ok, #state{}}.
+init() ->
+    %% TableName = list_to_atom(lists:flatten([atom_to_list(?MODULE), $_, "user"])),
+    TableName = ?MODULE,
+    Table = ets:new(TableName, [{write_concurrency, true}, {keypos, 2}, public, set]),
```
shino: I prefer `#user_state.user` instead of `2` for readability. It explains this table stores `#user_state` records.

src/riak_cs_bw_limitter.erl
```diff
@@ -0,0 +1,197 @@
...
+    ets:delete_all_objects(Table),
+    _ = maybe_schedule_refresh(Sec),
+    {noreply, State};
+handle_info(_, State) ->
+    {noreply, State}.
+
+update_state(Data, #state{table=Table} = _State) ->
+    ets:delete_all_objects(Table),
+    {DurationSec, BandwidthMax, AccessCountMax} = Data,
+    _ = maybe_schedule_refresh(DurationSec),
```
shino: If this function is called 1000 times per sec, many refresh events will fire! :fire:

src/riak_cs_s3_response.erl
```diff
@@ -43,8 +43,11 @@
 -type xmlElement() :: #xmlElement{}.
 
 -type error_reason() :: atom()
-                  | {'riak_connect_failed', term()}
-                  | {'malformed_policy_version', string()}.
+                      | {'riak_connect_failed', term()}
+                      | {'malformed_policy_version', string()}
+                      | {'disk_quota', non_neg_integer(), non_neg_integer()}
+                      | {'bandwidth', non_neg_integer(), non_neg_integer()}
+                      | {'access_count', non_neg_integer(), non_neg_integer()}.
```
shino: It seems taking these error reasons into riak cs itself is a little too tightly coupled.
Or, if these three reasons added here is general enough, it may be better to add callback functions for error message, error code and api_error.

src/riak_cs_simple_quota.erl
```diff
@@ -0,0 +1,218 @@
...
+    end.
+
+sum_all_buckets(UsageList) ->
+    lists:foldl(fun({<<"StartTime">>, _}, Sum) -> Sum;
+                   ({<<"EndTime">>, _}, Sum) -> Sum;
+                   ({_BucketName, {struct, Data}}, Sum) ->
+                        case proplists:get_value(<<"Bytes">>, Data) of
+                            I when is_integer(I) -> Sum + I;
+                            _ -> Sum
+                        end
```
shino: I don't care completeness because this module is an example implementation.
If someone use it or implement his/her own one, please add clause `{BucketName, ErrorMsgBin}`, or catch-all-and-just through clause.

src/riak_cs_bw_limitter.erl
shino: Why not use `named_table`? It can be eliminated table handle in state.

kuenishi:
I need more to please dialyzer

kuenishi:
I think this is ready for review again. Will run r_t next week, and if we agree on basic ideas, then I'll squash commits and create a cleanly rebased pull request.

src/riak_cs_bw_limitter.erl
```diff
@@ -0,0 +1,235 @@
...
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% ---------------------------------------------------------------------
+
+%% @doc Simple node-level rate limit implementation, per user
+%% This is experimental feature.
+%%
+%% riak_cs_bw_limitter help
```
shino: By google'ing, there are both "limitter" and "limiter" (difference of one "t" or two). I don't know their difference. :sweat_smile:

src/riak_cs_bw_limitter.erl
```diff
@@ -0,0 +1,235 @@
...
+                _ -> Bytes0
+            end,
+    UpdateOps = [{#user_state.bandwidth, Bytes},
+                 {#user_state.access_count, 1}],
+    try
+        case ets:update_counter(?MODULE, User, UpdateOps) of
+            [_, _] = R ->
+                ok;
+            Error0 ->
+                lager:warning("something wrong? ~p", [Error0]),
```
shino: Depending on refresh interval, this warning logs can be spammy. Race example steps:
1. get user_state record from ets
2. refresh triggered
3. ets truncated
4. update_counter

shino: No. Sorry. The steps above results in exception and caught in catch block below.

src/riak_cs_quota.erl
```diff
@@ -0,0 +1,130 @@
...
+%% under the License.
+%%
+%% ---------------------------------------------------------------------
+
+%% @doc Behaviour for Quota / QoS hook point, to save CS systems by
+%% limiting loads. The idea of Quota and QoS system is, to limit
+%% things by pluggable modules. This is because requirements vastly
+%% varies and it is too hard to cover all of them with single
+%% implementation.
+%% This is experimental feature. These interfaces and behaviousr
```
shino: s/behaviousr/behaviousr/

src/riak_cs_bw_limitter.erl
kuenishi: Limitter 

and

Limiter

I'll change the whole words to limiter, and module name will be `riak_cs_simple_bw_limiter`, aligning to the quota one.

src/riak_cs_quota.erl
shino: I want it to be explicitly written because my memory is so bad :smile: 
From source code, for `allow(Owner, Access, Context)`, `Owner` is bucket owner,  `Access` and `Context` have access user (maybe `undefined` for anonymous) on the other hand.

shino: For `update(Key, LogData)`, `Key` is the user key of bucket owner. Access users may be retrieved from headers in `LogData`.

src/riak_cs_quota.erl
```diff
@@ -0,0 +1,130 @@
...
+%% preserved while this quota process is alive. As no system
+%% configurations provided, module programmers should set their own
+%% configuration system. So aren't command-line console.
+-callback start_link() ->
+    {ok, pid()} | ignore | {error, Reason :: term()}.
+
+%% @doc note that this callback is not called at single process, but
+%% called at each mochiweb acceptor context.
+-callback allow(
+            Owner :: rcs_user(), %% Owner of the target object
```
shino: This can be `undefined` for anonymous access.

shino: Also there is possibility for `string()` for non-owner access but allowed by ACL/Policy .

src/riak_cs_simple_quota.erl
```diff
@@ -0,0 +1,228 @@
...
+maybe_usage(_, _Ctx = #context{riak_client=undefined}) ->
+    %% can't happen here
+    error(no_riak_client);
+maybe_usage(User0, _Ctx = #context{riak_client=RiakClient}) ->
+    User = binary_to_list(User0),
+    Usage = case get_latest_usage(RiakClient, User) of
+                {error, notfound} ->
+                    lager:warning("No storage stats data was found. Starting as no usage."),
+                    0;
+                {ok, [Usage0|_]} ->
```
shino: I guess the last element is latest.

src/riak_cs_quota.erl
kuenishi: Heh, dialyzer told me nothing about this :(

src/riak_cs_simple_quota.erl
kuenishi: Good pick.

src/riak_cs_bw_limitter.erl
```diff
@@ -0,0 +1,235 @@
...
+        case application:get_env(riak_cs, bw_limitter_interval) of
+            {ok, V} when is_integer(V) -> V;
+            _ -> ?DEFAULT_REFRESH_INTERVAL_SEC
+        end,
+    case test_bandwidth_state(IntervalSec, UserState) of
+        ok ->
+            case test_access_state(IntervalSec, UserState) of
+                ok ->
+                    {ok, RD, Ctx};
+                {error, {_, Current, Threshold}} = Error ->
```
shino: This is not rate, but accumulated.

src/riak_cs_bw_limitter.erl
```diff
@@ -0,0 +1,235 @@
...
+                    lager:warning("User ~p has exceeded access limit: usage, limit = ~p, ~p (/sec)",
+                                  [OwnerKey, Current, Threshold]),
+                    Error
+            end;
+        {error, {_, Current, Threshold}} = Error2 ->
```
shino: Ditto.

src/riak_cs_bw_limitter.erl
```diff
@@ -0,0 +1,235 @@
...
+                     #user_state{bandwidth = Bandwidth}) ->
+    BandwidthMax =
+        case application:get_env(riak_cs, bw_limitter_bandwidth_limit) of
+            {ok, V} when is_integer(V) -> V;
+            _ -> -1
+        end,
+    case Bandwidth / DurationSec < BandwidthMax  of
+        true -> ok;
+        false when BandwidthMax < 0 -> ok; %% No quota set
+        false -> {error, {bandwidth, Bandwidth, BandwidthMax}}
```
shino: `Bandwidth` is accumulated number.

src/riak_cs_simple_bwlimiter.erl
```diff
@@ -0,0 +1,236 @@
...
+%% This is experimental feature.
+%%
+%% riak_cs_bwlimiter help
+%%
+%%    This module enables node-level rate limiting.
+%%
+%%  set_limits(DurationSec, BandwidthMax, AccessCountMax)
+%%
+%%    Set bandwidth and access limit. DurationSec is refresh rate.
+%%    set_limits(-1, -1, -1) will remove all rate limitation.
```
shino: This example does not work.

src/riak_cs_simple_bwlimiter.erl
```diff
@@ -0,0 +1,236 @@
...
+    TableName = ?MODULE,
+    TableName = ets:new(TableName, [{write_concurrency, true},
+                                    {keypos, #user_state.user},
+                                    named_table, public, set]),
+    Pid = proc_lib:spawn_link(fun() -> refresher() end),
+    register(?MODULE, Pid),
+    {ok, Pid}.
+
+refresher() ->
+    IntervalSec = case application:get_env(riak_cs, bw_limitter_interval) of
```
shino: Should be `bwlimiter_interval`

src/riak_cs_simple_quota.erl
```diff
@@ -0,0 +1,230 @@
...
+
+%% @doc Only puts affected
+
+-spec allow(rcs_user(), access(), #context{}) ->
+                   {ok, #wm_reqdata{}, #context{}} |
+                   {error, {disk_quota, non_neg_integer(), non_neg_integer()}}.
+allow(Owner, #access_v1{req = RD, method = 'PUT'} = _Access, Ctx) ->
+    OwnerKey = iolist_to_binary(riak_cs_user:key_id(Owner)),
+    lager:debug("access => ~p", [OwnerKey]),
+    {_, Usage} = case ets:lookup(?MODULE, OwnerKey) of
```
shino, line 123: Lookup key is `binary()`, but...

src/riak_cs_simple_quota.erl
```diff
@@ -0,0 +1,230 @@
...
+                    lager:warning("No storage stats data was found. Starting as no usage."),
+                    0;
+                {ok, Usages} ->
+                    sum_all_buckets(lists:last(Usages))
+            end,
+    %% Here's a race condition where if so many concurrent access
+    %% come, each access can yield a new fresh state and then
+    %% receives not access limitation.
+    try
+        ets:insert_new(?MODULE, {User, Usage}),
```
shino: Inserted as `string()`.

kuenishi:
I think it's fixed now :8ball:

src/riak_cs_simple_quota.erl
```diff
@@ -0,0 +1,234 @@
...
+       #wm_log_data{method = Method,
+                    headers = Headers} = _LogData)
+  when Method =:= 'PUT' orelse Method =:= 'DELETE' ->
+
+    RequestLength = mochiweb_headers:get_value("content-length", Headers),
+    Bytes = case catch list_to_integer(RequestLength) of
+                Len when is_integer(Len) ->
+                    case Method of
+                        'PUT' -> Len;
+                        'DELETE' -> - Len
```
shino: DELETE request length is almost zero.

src/riak_cs_wm_utils.erl
```diff
@@ -799,16 +799,36 @@ object_access_authorize_helper(AccessType, Deletable, SkipAcl,
...
                 {ok, {actor_is_not_owner_but_allowed_policy, OwnerId}} ->
                     %% actor is not the owner
-                    actor_is_not_owner_but_allowed_policy(User, OwnerId, RD, Ctx, LocalCtx);
+                    %% Quota hook here
+                    case riak_cs_quota:invoke_all_callbacks(OwnerId, Access, Ctx) of
+                        {ok, RD2, Ctx2} ->
+                            actor_is_not_owner_but_allowed_policy(User, OwnerId, RD2, Ctx2, LocalCtx);
+                        {error, Module, Reason, RD3, Ctx3} ->
+                            riak_cs_quota:handle_error(Module, Reason, RD3, Ctx3)
+                            %%riak_cs_s3_response:api_error(Reason, RD, Ctx)
```
shino: Can we remove this line?

src/riak_cs_simple_quota.erl
```diff
@@ -0,0 +1,234 @@
...
+    NowASec = calendar:datetime_to_gregorian_seconds(Now),
+    get_latest_usage(Pid, User, NowASec, 0, 10).
+
+-spec get_latest_usage(pid(), string(),
+                       non_neg_integer(), non_neg_integer(), non_neg_integer()) ->
+                              {ok, list()} | {error, notfound}.
+get_latest_usage(_Pid, _User, _, N, N) -> {error, notfound};
+get_latest_usage(Pid, User, EndSec, N, Max) ->
+    End = calendar:gregorian_seconds_to_datetime(EndSec),
+    ADayAgo = calendar:gregorian_seconds_to_datetime(EndSec - 86400),
```
shino: If hard coded `86400` is meant for the default value of `storage_archive_period`, using application environment value would be better.

shino:
Both bw and quota limiter modules work well, great! :sunny: 
I will :+1: after above comments are addressed and dialyzer failure is fixed.

src/riak_cs_simple_quota.erl
kuenishi: Ooops

kuenishi:
Updated ^^;

src/riak_cs_simple_quota.erl
```diff
@@ -0,0 +1,241 @@
...
+    Body = riak_cs_xml:to_xml([XmlDoc]),
+    ReqData = wrq:set_resp_header("Content-Type", ?XML_TYPE, RD),
+    UpdReqData = wrq:set_resp_body(Body, ReqData),
+    {StatusCode, UpdReqData, Ctx}.
+
+-spec get_latest_usage(pid(), string()) -> {ok, list()} | {error, notfound}.
+get_latest_usage(Pid, User) ->
+    Now = calendar:now_to_datetime(os:timestamp()),
+    NowASec = calendar:datetime_to_gregorian_seconds(Now),
+    LeewaySeconds = riak_cs_gc:leeway_seconds(),
```
shino: `riak_cs_storage:archive_period()` ?

shino:
Added the last comment :sweat_smile:

kuenishi:
oh, conflicts....

kuenishi:
Conflicts resolved.

kuenishi:
@borshop merge

Basho-JIRA:
Release note already written: see an item starting "Module-level hook point for .."   in https://github.com/basho/riak_cs/blob/develop/RELEASE-NOTES.md#additions  

Issue #1125, opened 2015-04-17
Cannot set gc.interval as infinity in riak-cs.conf [JIRA: RCS-189]


Basho-JIRA:
Current workaround is to use advanced.config: `{gc_interval, infinity}` .

**[Created in JIRA by Kota Uenishi]**

PR #1126, opened 2015-04-17
Fix gc.interval doesn't accept infinity.

ksauzz:


kuenishi:
Addresses #1125 (RCS-189).

kuenishi:
@borshop merge

PR #1124, opened 2015-04-17
Feature/offline gc tool

kuenishi:
Discussion topics:
- how much does it work with old Bitcask old format?
- performance?
- bugs?

PR #1128, opened 2015-04-23
Orphan block scanner scripts

shino:
## Steps for scanning orphan blocks
1. List keys in `moss.buckets`
2. Iterate over all CS buckets (optionally ones listed in command line)
   1. List keys (streaming listkeys) on a `<<0b:...>>` bucket
      - _We MUST list keys of blocks at first_ to avoid race from API calls
   2. Collect all blocks, as `{UUID, [list of existing block seqs]}`
   3. List manifest UUIDs on `<<0o:...>>` bucket (by streaming `csbucketfold`)
   4. Check existence of corresponding manifest by subtracting the two
      results of list keys
   5. Complete maybe orphaned UUID list
3. Filter out false-orphaned block UUIDs.
   This is because list keys (or csbucketfold) of manifest buckets will
   be done by R=1, then replica deficit (e.g. by force-replace) leads to
   false-positiveness for orphan block extraction at previous step.
   R=quorum (or R=all) manifest GETs is needed to avoid false-positive.
   1. Get block object and get CS key of manifests from user metadata
   2. Get manifest for the key and verify existence of corresponding manifest
   3. Remove block UUIDs if manifest exists from orphaned list

Usage (for debugging)

```
% riak-cs escript /path/to/priv/tools/internal/block_audit.erl  \
      -h 127.0.0.1 -p 10017 \
      -dd \
      -o $PWD/blocks.local
% riak-cs escript /path/to/priv/tools/internal/block_audit2.erl \
      -h 127.0.0.1 -p 10017 \
      -dd \
      -i $PWD/blocks.local -o $PWD/blocks2.local
```

This PR adds scripts for gathering block {UUID, Sequence}s to be deleted.
For actual deletion, we can re-using block cleaner script in #1124.
But there is still some interface (file format) discrepancy, subsequent
PR is needed.

### DIFF
--- a/priv/tools/internal/block_audit.erl
+++ b/priv/tools/internal/block_audit.erl
@@ -70,10 +70,9 @@ option_spec() ->
      {host, $h, "host", string, "Host of Riak PB"},
      {port, $p, "port", {integer, 8087}, "Port number of Riak PB"},
      {bucket, $b, "bucket", string, "CS Bucket to audit, repetitions possible"},
-     {output, $o, "output", {string, "orphaned-blocks"}, "Directory to output resutls"},
-     {debug, $d, "debug", {integer, 0}, "Enable debug (-dd for more verbose)"},
-     {page_size, $s, "page-size", {integer, 1000}, "Specify page size for 2i listing"},
-     {dry_run, undefined, "dry-run", {boolean, false}, "if set, actual update does not happen"}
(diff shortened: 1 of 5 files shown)


shino:
riak_test case added, rebased & force-push'ed.
Ready for review  now.

shino:
:speak_no_evil:

Issue #1129, opened 2015-04-24
Multiple delete fails to delete objects with key containing 2+ subsequent space characters  [JIRA: RCS-191]


shino:
~~2015-04-28 Update: Changed subject to indicate the bug of list objects API. Multiple delete API itself does no bad. Some tools like s3cmd issues list objects before multiple delete to grab keys, so such tools are affected.~~

Confirmed at version 2.0.0, probably 1.5.x are affected too.

Steps:

```
% s3cmd -c .s3cfg.15018.alice mb s3://test3
Bucket 's3://test3/' created
% s3cmd -c .s3cfg.15018.alice put tools.mk s3://test3/'three   spaces'
tools.mk -> s3://test3/three   spaces  [1 of 1]
 5751 of 5751   100% in    0s   418.06 kB/s  done
% s3cmd -c .s3cfg.15018.alice ls s3://test3/
2015-04-24 02:57      5751   s3://test3/three   spaces
% s3cmd -c .s3cfg.15018.alice del -r -f s3://test3/
File s3://test3/three   spaces deleted
% s3cmd -c .s3cfg.15018.alice ls s3://test3/
2015-04-24 02:57      5751   s3://test3/three   spaces
```

Expected output from the final step (`s3cmd ls`) is empty, but
there was one.

From the output of console debug log from multiple detele path,
sebsequnt spaces became corrupted to single space:

```
11:57:36.407 [debug] deleting keys at <<"test3">>: [<<"three spaces">>]
```

zd://10707

shino:
Spaces are trimmed at XML parsing [1], which was introduced by [2] fixing other bug.

The function is called by:

```
./riak_cs_wm_user.erl:143:    case riak_cs_xml:scan(Body) of
./riak_cs_wm_user.erl:160:    case riak_cs_xml:scan(Body) of
./riak_cs_wm_object_upload_part.erl:212:        {ok, ParsedData} = riak_cs_xml:scan(Body),
./riak_cs_s3_response.erl:335:    {ok, ParsedData} = riak_cs_xml:scan(Xml),
./riak_cs_wm_bucket_delete.erl:161:    case riak_cs_xml:scan(Body) of
./riak_cs_acl_utils.erl:305:    case riak_cs_xml:scan(Xml) of
```

[1] https://github.com/basho/riak_cs/blob/develop/src/riak_cs_xml.erl#L69
[2] https://github.com/basho/riak_cs/pull/795

PR #1135, opened 2015-04-28
Fix corruption of successive space charactors in XML request

shino:
Originally reported by zd://10707 and bug description is at #1129 (RCS-191).

Affected API
- user administration by XML
- Multiple delete

Other APIs that accepts XML as requests does not allow spaces in text nodes, so no actual harmful effect.

This PR fixes the bug by:
- Removing space normalization option for xml parsing by xmerl and
- broadening pattern match that depends on the option


kuenishi:
@borshop merge

PR #1136, opened 2015-04-28
Update Release Notes for 2.0.1

lrnrthr:


lrnrthr:
This file should take the place of the RELEASE-NOTES.md file.

kuenishi:
Thank you for grammar correction, but I don't understand why we're changing item listing style at version update from 2.0.0 to 2.0.1. I'd love unified style in a single text file.

lrnrthr:
@kuenishi I changed the item listing style because I was asked to create a standard style for release notes across Basho. However, your comment has made me think that it may be best to do so at 2.1 rather than 2.0.1. I will update my PR.

kuenishi:
@laurenrother that makes sense.

kuenishi:
@borshop merge

PR #1140, opened 2015-05-01
Fix race between reading file info and bitcask merge delete

shino:
Between listing a directory and reading each file info's, merge
delete can interleave. Then read file info results in enoent error.
This commit treats it as zero size file.

Typical error message of this race is:

```
{{badmatch,{error,enoent}},
 [{cs_suites,'-check_data_files_are_small_enough/3-lc$^0/1-0-',1,
             [{file,"riak_test/src/cs_suites.erl"},{line,538}]},
```

This PR includes only change of riak_test code.


kuenishi:
@borshop merge

PR #1141, opened 2015-05-01
Fix URL encoding bug around AWS v4 auth [JIRA: RCS-203]

shino:
AWS v4 auth defined URL encoding algorithm that is used for
generating signature. Mochiweb's utility function convert
space (" ") to plus ("+") but AWS spec says it should be
percent encoded. i.e. "%20".

This commit fixes the encoding bug for resource (URL path) and
query parameters.

This PR only affects AWS v4 authentication path that is still preliminary support.

cf.
Authenticating Requests by Using the Authorization Header
  (Compute Checksum of the Entire Payload Prior to Transmission)
- Signature Version 4 - Amazon Simple Storage Service
  http://docs.aws.amazon.com/ja_jp/AmazonS3/latest/API/sig-v4-header-based-auth.html

### DIFF
--- a/client_tests/python/boto_tests/boto_test.py
+++ b/client_tests/python/boto_tests/boto_test.py
@@ -164,6 +164,17 @@ class BasicTests(S3ApiVerificationTestBase):
         conn = self.make_connection(bad_user)
         self.assertRaises(S3ResponseError, conn.get_canonical_user_id)
 
+    def test_auth_weird_query_param(self):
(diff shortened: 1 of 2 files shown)


kuenishi:
@borshop merge

kuenishi:
@borshop: retry

Basho-JIRA:
No release note needed as AWS v4 auth is not supported in 2.0 nor 2.1. 

PR #1144, opened 2015-05-07
Extract out console facing FSM from riak_cs_gc_d

kuenishi:
riak_cs_gc_d used to have bunch of complex states not only name of
states exposed in gen_fsm, but also many elements in gc_d_state
record. This commit extracts operator-facing states such as idle,
paused, running from riak_cs_gc_d. riak_cs_gc_d is not a long-running
worker under supervision tree any more, but riak_cs_gc_manager is.
The lifecycle of riak_cs_gc_d is now as long as one GC batch - if the
GC has finished then riak_cs_gc_d process terminates.

See comments in both riak_cs_gc_d.erl and riak_cs_gc_manager.erl for
detail state diagram and description.


include/riak_cs_gc_d.hrl
```diff
@@ -106,3 +93,14 @@
 -define(DEFAULT_GC_BATCH_SIZE, 1000).
 -define(DEFAULT_GC_WORKERS, 2).
 -define(EPOCH_START, <<"0">>).
+
+-record(gc_manager_state, {
```
shino, line 97: This state record is referenced only from manager module, I guess.
Maybe just historical(?) reason, how about putting it into manger module file?

src/riak_cs_gc_d.erl
```diff
@@ -178,193 +80,62 @@ stop() ->
...
-                                      initial_delay=InitialDelay,
-                                      max_workers=MaxWorkers,
-                                      testing=Testing}),
-    {ok, idle, SchedState}.
+init([State]) ->
+    State1 = maybe_fetch_first_key(State),
+    NextState = maybe_start_workers(State1),
+    case has_batch_finished(NextState) of
+        true ->
+            {stop, normal, NextState};
```
shino: Only 2-tuple here `{stop, Reason}`.

src/riak_cs_gc_manager.erl
```diff
@@ -0,0 +1,294 @@
...
+    Leeway = proplists:get_value(leeway, Options,
+                                 riak_cs_gc:leeway_seconds()),
+                                 
+    %% set many items to GCDState here
+    GCDState = #gc_d_state{
+                  batch_start=BatchStart,
+                  leeway=Leeway,
+                  max_workers=MaxWorkers},
+
+    case riak_cs_gc_d:start_link(GCDState) of
```
shino: This call may be blocked by 2i fetch and it may be long depending on cluster / nodes status.

include/riak_cs_gc_d.hrl
kuenishi, line 97: This is because they're also used in quickcheck tests.

src/riak_cs_gc_manager.erl
kuenishi: Will change...

src/riak_cs_gc_batch.erl
```diff
@@ -0,0 +1,313 @@
...
+
+%% @doc this fsm has no special upgrade process
+code_change(_OldVsn, StateName, State, _Extra) ->
+    {ok, StateName, State}.
+
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+
+maybe_fetch_first_key(?STATE{batch_start=BatchStart,
```
shino: This function fetches keys regardless of any condition. Then, "maybe" is not needed and s/key/keys/.

src/riak_cs_gc_console.erl
```diff
@@ -127,56 +124,32 @@ set_leeway(Leeway) ->
 handle_batch_start(ok) ->
     output("Garbage collection batch started."),
     ok;
-handle_batch_start({error, already_deleting}) ->
-    output("Error: A garbage collection batch"
-           " is already in progress."),
-    error;
-handle_batch_start({error, already_paused}) ->
-    output("The garbage collection daemon was already paused."),
+handle_batch_start({error, no_data_to_collect}) ->
```
shino: Who emits this error `no_data_to_collect` ?

src/riak_cs_gc_manager.erl
```diff
@@ -0,0 +1,285 @@
...
+%% start -> idle -(start)-> running
+%%            ^               |
+%%            +----(finish)---+
+%%            +----(stop)-----+
+%%
+%% Message excange chart (not a sequence, but just a list)
+%%
+%%  Message\  sdr/rcver  gc_manager    gc_batch
+%%   spawn_link)                   --->
+%%   cancel)                       --->
```
shino, line 33: Start is used in the state diagram above, cancel is used here and command line interface.
Because stop is used in gen_(fsm|server) contexts as to stop process itself but it stops only child process, cancel is better wording?

shino, line 33: After reading code below, I noticed no `stop` event :smile: It seems like just a typo in the state diagram.

include/riak_cs_gc.hrl
```diff
@@ -106,3 +93,14 @@
 -define(DEFAULT_GC_BATCH_SIZE, 1000).
 -define(DEFAULT_GC_WORKERS, 2).
 -define(EPOCH_START, <<"0">>).
+
+-record(gc_manager_state, {
+          next :: undefined | non_neg_integer(),
+          gc_batch_pid :: undefined | pid(),
+          batch_history = [] :: list(#gc_batch_state{}),
```
shino, line 100: Why does manager hold whole history? I could not find the place it was used.

src/riak_cs_gc_manager.erl
```diff
@@ -0,0 +1,285 @@
...
+
+init([]) ->
+    process_flag(trap_exit, true),
+    InitialDelay = riak_cs_gc:initial_gc_delay(),
+    State = case riak_cs_gc:gc_interval() of
+                infinity ->
+                    #gc_manager_state{};
+                Interval when is_integer(Interval) ->
+                    Next = riak_cs_gc:timestamp() + InitialDelay,
+                    TimerRef = erlang:send_after(InitialDelay * 1000, self(), {start, []}),
```
shino: This message will be ignored.

src/riak_cs_gc_manager.erl
```diff
@@ -0,0 +1,285 @@
...
+    {reply, Reply, running, State}.
+
+%% @private
+%% @doc Not used.
+handle_event(_Event, StateName, State) ->
+    {next_state, StateName, State}.
+
+%% @private
+%% @doc
+handle_sync_event({set_interval, Initerval}, _From, StateName, State) ->
```
shino, line 183: It's better to set application environment variable here. I guess original (before refactoring) code lacks it... :sweat_smile: 

More simply, just set app env var and remove interval field from state record is alternative, as `set_leeway` does.

src/riak_cs_gc_manager.erl
```diff
@@ -0,0 +1,285 @@
...
+    %% for tests
+    {stop, normal, ok, State};
+handle_sync_event(_Event, _From, StateName, State) ->
+    Reply = {error, illegal_sync_send_all_state_call},
+    {reply, Reply, StateName, State}.
+
+%% @private
+%% @doc Because process flag of trap_exit is true, gc_batch
+%% failure will be delivered as a message.
+handle_info({'EXIT', _Pid, normal}, StateName, State) ->
```
shino: - No logging in this clause?
- If it is in `running` state, transition to `idle` is needed?
- Pattern match with `gc_batch_pid` is unncessary? Pid in EXIT message may be from workers, riak_client, etc.

src/riak_cs_gc_key_list.erl
```diff
@@ -56,6 +56,12 @@ next(#gc_key_list_state{current_riak_client=RcPid,
     {#gc_key_list_result{bag_id=BagId, batch=Batch},
      State#gc_key_list_state{continuation=UpdContinuation}}.
 
+-spec has_next(gc_key_list_state()) -> boolean().
```
shino, line 60: `s/has_next/no_more_result/` ( ﾟ∀ﾟ)･∵. ｸﾞﾊｯ!!

src/riak_cs_gc_batch.erl
```diff
@@ -0,0 +1,313 @@
...
+                     maybe_fetch_next_keys(State);
+                 _ ->
+                     State
+             end,
+    case has_batch_finished(State2) of
+        true ->
+            {stop, normal, State2};
+        _ ->
+            %%?debugHere,
+            State3 =  maybe_start_workers(State2),
```
shino: Maybe, `maybe_fetch_next_keys` and `maybe_start_workers` should be called interleaved way, even after just single `batch_complete` event. It's because `?STATE.batch` list has at most only one element (under current impl).

src/riak_cs_gc_batch.erl
```diff
@@ -0,0 +1,313 @@
...
+        riak_cs_gc_key_list:new(BatchStart, Leeway),
+    #gc_key_list_result{bag_id=BagId, batch=Batch} = KeyListRes,
+    _ = lager:debug("Initial batch keys: ~p", [Batch]),
+    State?STATE{batch=Batch,
+                key_list_state=KeyListState,
+                bag_id=BagId}.
+
+maybe_fetch_next_keys(?STATE{key_list_state=undefined} = State) ->
+    State;
+maybe_fetch_next_keys(?STATE{key_list_state=KeyListState} = State) ->
```
shino: If `key_list_state` is not `undefined`, it should be checked there is more results before issuing 2i fetch.
The cost of 2i fetch is rather high (at least much higher than on memory checking of record fields).

src/riak_cs_gc_manager.erl
kuenishi, line 183: The reason why I'm not using application environment is for `pause` and `resume` command, which needs somewhat 'default' original value before pause.

shino, line 183: Interesting coincidence. These are the very reason I thought app env was better so that `resume` could resume with the interval value that had been set last.

shino, line 183: So, different angle. Which is _natural_ after `riak_cs_gc_manager` dies and restarts, back to interval in riak-cs.conf or the value set last by `set-interval`?

kuenishi, line 183: Is that possible to re-read the value from config file? I though there were no way other than getting app env to get values from config files.

include/riak_cs_gc.hrl
kuenishi, line 100: Wouldn't it be good if we can know stats of three or ten last run of garbage collection?

src/riak_cs_gc_manager.erl
shino, line 183: Sorry that I misunderstood the behavior. Re-read code around pause and resume, the implementation is perfect.

include/riak_cs_gc.hrl
shino, line 100: That's nice. What I'm concerned about is indefinite growth of the list.
5min interval (for worse case estimation, somewhat smaller than default) then 8640 entries for a month, 10^5 for a year.

include/riak_cs_gc_d.hrl
shino, line 97: (￣ー￣)ｂ

shino:
`riak-cs-admin` script has usage infromation, which should be updated.

```
                echo "Usage: $SCRIPT gc { batch [<leeway_seconds>] | status | pause | resume | cancel |"
                echo "                          set-interval <interval_seconds> | set-leeway <leeway_seconds> }"
```

src/riak_cs_gc_manager.erl
```diff
@@ -0,0 +1,325 @@
...
+%% under the License.
+%%
+%% ---------------------------------------------------------------------
+%%
+%% State Diagram
+%%
+%% start -> idle -(start)-> running
+%%            ^               |
+%%            +----(finish)---+
+%%            +----(stop)-----+
```
shino: s/stop/cancel/

src/riak_cs_gc_manager.erl
```diff
@@ -0,0 +1,325 @@
...
+running(cancel, _From, State = #gc_manager_state{gc_batch_pid=Pid}) ->
+    %% stop gc_batch here
+    catch riak_cs_gc_batch:stop(Pid),
+    NextState = schedule_next(State),
+    {reply, ok, idle, NextState#gc_manager_state{gc_batch_pid=undefined}};
+running({finished, Report}, _From, State) ->
+    %% Add report to history
+    NextState=schedule_next(State),
+    {reply, ok, idle,
+     NextState#gc_manager_state{batch_history=[Report]}};
```
shino, line 171: Better to set `gc_batch_pid` to `undefined`? I guess there's no harm if not set, just for good correspondence with one above branch of `cancel`.

src/riak_cs_gc_manager.erl
```diff
@@ -0,0 +1,325 @@
...
+        Error ->
+            _ = lager:error("Cannot start batch. Reason: ~p", [Error]),
+            NextState = schedule_next(State),
+            {next_state, idle, NextState}
+    end;
+handle_info(Info, StateName, State) ->
+    _ = lager:warning("Error detected at GC process (~p): ~p",
+                      [StateName, Info]),
+    _ = lager:warning("Error detected at GC process (~p): ~p",
+                      [StateName, State]),
```
shino: - Why two different log lines?
- Warning level log whose message says error, maybe confusing.

kuenishi:
I don't think any usage of `riak-cs-admin` has changed?

src/riak_cs_gc_manager.erl
kuenishi: This is sorta debug log, will fix.

src/riak_cs_gc_manager.erl
kuenishi, line 171: No. `gc_batch_pid` is set undefined at `handle_info` as the process traps exits.

include/riak_cs_gc.hrl
kuenishi, line 100: Well, thanks for agreement. Although it's good idea I think, somewhat it's out of the scope of refactoring and will do in future. For now, the history will be a list of just one element. Please see the latest code in manager receiving `finished` message.

src/riak_cs_gc_batch.erl
```diff
@@ -0,0 +1,309 @@
...
+     {files_deleted, State?STATE.batch_count},
+     {files_skipped, State?STATE.batch_skips},
+     {files_left, if is_list(State?STATE.batch) -> length(State?STATE.batch);
+                     true                       -> 0
+                  end}].
+
+
+%% ===================================================================
+%% Test API and tests
+%% ===================================================================
```
shino: `elapsed` below is not (only) for testing.

src/riak_cs_gc_manager.erl
shino, line 171: Understood the flow. Then, why manually set to `undefined` at `cancel` event? In the event, `EXIT` will happen.

include/riak_cs_gc.hrl
shino, line 100: :+1:

src/riak_cs_gc_manager.erl
shino, line 171: It seems `schedule_next` will be called twice by `finished` and `EXIT`.
How about the arrangement as
- this branch has things after successful batch and
- `EXIT` event handling has things in general regardless of success or failure
  ?

shino, line 171: Multiple calls for `schedule_next` is _safe_ :smile: So the above comment is only for clearness, not mandatory.

shino:
> I don't think any usage of riak-cs-admin has changed?

True. Sorry for my misunderstanding (again) :disappointed:

src/riak_cs_gc_manager.erl
kuenishi, line 171: For now I don't have idea to cleanly and safely handle these events ... maybe to be left for future refactoring.

shino:
Eunit, xref , dialyzer succeeded.
Much cleaner code. Great refactoring.

+1 to merge.

shino:
@borshop merge

Basho-JIRA:
The pull request was merged Monday, without starting the sprint (Most of the work was finished, so I was working on next work). 

Basho-JIRA:
No release note needed as this is just a code cleanup. 

PR #1156, opened 2015-05-29
Add case clause for tombstone return

kuenishi:


shino:
Seems like existing implementation is based on buggy behaivior [1]

[1] https://github.com/basho/riak-erlang-client/pull/196

shino:
@borshop merge

PR #1157, opened 2015-06-02
[wip] Feature/share manifest tools

kuenishi:
This is a part of efforts of improving delete performance by skipping GC but actively deleting on requests - aka Yak Shaving or refactoring before change. This pull request moves manifests related codes out of current repository to a new repo called `riak_cs_manifests`. That repo is also referred from [riak_kv](https://github.com/basho/riak_kv/tree/feature/bundle-cs-manifest) so as to avoid ugly `add_paths` configuration from advanced.config.

If this is merged and riak_kv is not yet updated, the value of `add_paths` should be changed to `riak-cs/lib/riak_cs_manifests-<version>/ebin` from `riak-cs/lib/riak_cs-<version>/ebin`.

`riak_cs_manifests` is [here](https://github.com/basho/riak_cs_manifests/pull/1).


shino:
Could you describe the purpose and principle for separation?

(I'm not against the separation :smile:  I think it's good to separate data structure from connection management or REST API handling )

kuenishi:
Current diff is for just 'having-no-regression' and 'moving-least-files' moving. Next step is changing function call structure.

kuenishi:
Operators interested in storage calculation might also take the handy work to add `add_paths` to advanced.config in Riak.

PR #1168, opened 2015-06-18
Update riak_cs_multibag tag to 1.5.4p1

shino:
There are several unreleased changes in release/1.5 branches of riak_cs and riak_cs_multibag.
This PR updates deps for riak_cs_multibag to tag/1.5.4p1 to avoid one step of manual checkout under deps.


kuenishi:
You've been also assigned to update the version of all other dependencies in develop. :imp:

shino:
@borshop merge

PR #1170, opened 2015-06-23
Fix erlang client test compilation

kuenishi:
Looks like it have been broken since #1144 (RCS-199).


shino:
Nice bug finder :eyes:

kuenishi:
@borshop merge

PR #1175, opened 2015-06-30
Add deprecated warning to old CLIs  [JIRA: RCS-219]

ksauzz:
It would be better to show a user a deprecated warning message before removing these commands.

### DIFF
--- a/rel/files/riak-cs-access
+++ b/rel/files/riak-cs-access
@@ -1,3 +1,10 @@
 #!/bin/sh
 
+cat<<EOS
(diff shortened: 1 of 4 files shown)


kuenishi:
@borshop merge

Basho-JIRA:
For release note: Administration CLIs including `riak-cs-storage`, `riak-cs-gc`, `riak-cs-access` and `riak-cs-stanchion` are replaced with `riak-cs-admin` command. Those commands will be removed in future released. 

PR #1176, opened 2015-07-01
Warn in startup if dvv_enabled is false [JIRA: RCS-224]

kuenishi:


shino:
For completeness, one more assertion of `lww=false` is desirable, isn't it?

kuenishi:
add check on lww=false, too. Also, result of `riakc_pb_socket:get_bucket_type(Pid, <<"default">>)` does not include `dvv_enabled` explicitly, while HTTP API does. It may be another issue, but any thoughts?

kuenishi:
Ooops, dialyzer.

kuenishi:
Okay, pleased dialyzer

shino:
I added following lines to riak.conf and start riak and riak-cs, then got warning about `dvv_enabled`

```
[warning] dvv_enabled is strongly recommended to be true (currently undefined)
```

Configuration is wrong?

kuenishi:
I also tested with this config:

```
buckets.default.allow_mult = true
buckets.default.merge_strategy = 2
```

My test result: https://gist.github.com/kuenishi/eeb601889e985a6072fb

My conclusion: it's like a Riak bug in PB API.

kuenishi:
Found the culprit: https://github.com/basho/riak_pb/blob/develop/src/riak_pb_codec.erl#L241-L299

The decoder ignores `dvv_enabled` and `write_once` .

shino:
Then, this PR will be deferred to 2.1.1?

kuenishi:
Opened https://github.com/basho/riak_pb/pull/127 as a right fix. We'll have to work on workaround, too...

kuenishi:
> Then, this PR will be deferred to 2.1.1?

If fortunately that is included in Riak 2.1.2, I'd be happy but will push my workaround.

src/riak_cs_app.erl
```diff
@@ -57,14 +55,11 @@ stop(_State) ->
 sanity_check(true, {ok, true}) ->
     riak_cs_sup:start_link();
 sanity_check(false, _) ->
-    _ = lager:error("You must update your Riak CS app.config. Please see the"
+    _ = lager:error("You must update your Riak CS riak-cs.conf. Please see the"
                     "release notes for more information on updating you"
```
shino, line 59: It's better to add white spaces at the end of these two lines because concatenated string is shown in the log.

kuenishi:
Dropping this off of 2.1.0 because it's silly to connect to HTTP port just for checking dvv_enabled. Will revive this after https://github.com/basho/riak_pb/pull/127 is merged.

shino:
This is on-hold because of riak kv side issue. This is worth to put in future version of riak_cs.

PR #1181, opened 2015-07-09
Add default riak_cs_wm_common's multiple_choices callback  [JIRA: RCS-221]

shino:
And remove the identical implementations in sub-modules.

Without this default function, `riak_cs_wm_common:undefined/2` is
called and eventually `code_server:call()` is hit unless sub-module
does not implement it. This is one of possible bottleneck in
processing HTTP request in Riak CS.

### DIFF
--- a/src/riak_cs_wm_bucket_uploads.erl
+++ b/src/riak_cs_wm_bucket_uploads.erl
@@ -34,7 +34,6 @@
          allowed_methods/0,
          malformed_request/2,
          content_types_accepted/2,
-         multiple_choices/2,
          finish_request/2]).
 
 -include("riak_cs.hrl").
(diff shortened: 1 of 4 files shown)


kuenishi:
Checked manually it's working by

```
> redbug:start("error_handler:undefined_function -> return", [{msgs,100},{max_msg_size, 9999999999999}]).
```

and hitting some API with curl `curl -v -s -i http://localhost:8080/test-bucket/a` .

```
{error_handler,
                                              undefined_function,
                                              [riak_cs_wm_common,
                                               undefined, ....
```

kuenishi:
@borshop merge

Basho-JIRA:
For release note:

Add default callback of `multiple_choices` to `riak_cs_wm_common` to prevent bothering code_server, which could possibly be bottleneck. 

PR #1188, opened 2015-07-15
Sibling resolver [JIRA: RCS-249]

kuenishi:
This is very well tested. Not really.

### DIFF
--- a/src/riak_cs_console.erl
+++ b/src/riak_cs_console.erl
@@ -230,6 +230,8 @@ resolve_siblings(Pid, RawBucket, RawKey, GetOptions, GetTimeout, PutOptions, Put
                     R = riakc_pb_socket:put(Pid, RiakObj2, PutOptions, PutTimeout),
                     _ = lager:info("Resolution result: ~p~n", [R]),
                     R;
+                ok ->
+                    lager:info("No siblings in ~p:~p~n", [RawBucket, RawKey]);
                 {error, _} = E ->
                     _ = lager:info("Not updating ~p:~p: ~p~n", [RawBucket, RawKey, E]),
                     E
@@ -244,7 +246,9 @@ resolve_siblings(Pid, RawBucket, RawKey, GetOptions, GetTimeout, PutOptions, Put
 %%%===================================================================
 
 -spec resolve_ro_siblings(riakc_obj:riakc_obj(), binary(), binary()) ->
-                                 {ok, riakc_obj:riakc_obj()} | {error, term()}.
+                                 {ok, riakc_obj:riakc_obj()} | %% Needs update to resolve siblings
+                                 ok | %% No siblings, no need to update
+                                 {error, term()}. %% Some other reason that prohibits update
 resolve_ro_siblings(_, ?USER_BUCKET, _) ->
     %% This is not supposed to happen unless something is messed up
     %% with Stanchion. Resolve logic is not obvious and needs operator decision.
@@ -276,8 +280,8 @@ resolve_ro_siblings(RO, <<"0b:", _/binary>>, _) ->
         {E, true} ->
             _ = lager:info("Cannot resolve: ~p~n", [E]),
             {error, E};
-        {E, false} ->
-            {error, E}
+        {_, false} ->
+            ok
     end;
 resolve_ro_siblings(RiakObject, <<"0o:", _/binary>>, _RawKey) ->
     [{_, Manifest}|_] = Manifests =
--- a/src/riak_cs_console.erl
+++ b/src/riak_cs_console.erl
@@ -189,14 +189,7 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
 resolve_siblings(RawBucket, RawKey) ->
     {ok, RcPid} = riak_cs_riak_client:start_link([]),
     try
-        {ok, Pid} = case RawBucket of
-                        <<"0o:", _/binary>> ->
-                            riak_cs_riak_client:manifest_pbc(RcPid);
-                        <<"0b:", _/binary>> ->
-                            riak_cs_riak_client:block_pbc(RcPid);
-                        _ ->
-                            riak_cs_riak_client:master_pbc(RcPid)
-                    end,
+        {ok, Pid} = riak_cs_riak_client:master_pbc(RcPid),
         resolve_siblings(Pid, RawBucket, RawKey)
     after
         riak_cs_riak_client:stop(RcPid)
@@ -237,9 +230,8 @@ resolve_siblings(Pid, RawBucket, RawKey, GetOptions, GetTimeout, PutOptions, Put
                     R = riakc_pb_socket:put(Pid, RiakObj2, PutOptions, PutTimeout),
                     _ = lager:info("Resolution result: ~p~n", [R]),
                     R;
-                {error, R} = E ->
-                    _ = lager:info("Not updating ~p:~p: ~p~n",
-                                   [RawBucket, RawKey, R]),
+                {error, _} = E ->
+                    _ = lager:info("Not updating ~p:~p: ~p~n", [RawBucket, RawKey, E]),
                     E
             end;
         {error, Reason} = E ->
--- a/src/riak_cs_console.erl
+++ b/src/riak_cs_console.erl
@@ -28,6 +28,8 @@
          cleanup_orphan_multipart/1
         ]).
 
+-export([resolve_siblings/2, resolve_siblings/3, resolve_siblings/7]).
+
 -include("riak_cs.hrl").
 -include_lib("riakc/include/riakc.hrl").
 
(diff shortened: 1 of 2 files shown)


kuenishi:
Good update of original script on timeout by Taka:

``` erlang
Fun = fun(Host, Port, B, K) ->
              {ok, Pid} = riakc_pb_socket:start_link(Host, Port),
              try
                  {ok, RiakObject} = riakc_pb_socket:get(Pid, B, K, [{r, all},{timeout, 60000*5}], 60000*5),
                  case lists:any(fun({_, <<>>}) -> true; ({MD, _}) -> dict:is_key(<<"X-Riak-Deleted">>, MD) =:= true end, riakc_obj:get_contents(RiakObject)) of
                      true ->
                          VClock = riakc_obj:vclock(RiakObject),
                          ok = riakc_pb_socket:delete_vclock(Pid, B, K, VClock);
                      false ->
                          {Meta, Block} = hd(riakc_obj:get_contents(RiakObject)),
                          RO1 = riakc_obj:update_metadata(RiakObject, Meta),
                          RO2 = riakc_obj:update_value(RO1, Block),
                          ok = riakc_pb_socket:put(Pid, RO2,[{timeout, 60000*5}],60000*5)
                  end
              after
                  riakc_pb_socket:stop(Pid)
              end
      end.
```

src/riak_cs_console.erl
```diff
@@ -100,10 +102,84 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
...
+%% `riak-cs attach` RawKey and RawBucket should be provided via Riak
+%% log. Example usage is:
+%%
+%% > {ok, Pid} = riakc_pb_socket:start_link(Host, Port).
+%% > riak_cs_console:resolve_siblings(Pid, <<...>>, <<...>>).
+%% > riakc_pb_socket:stop(Pid).
+%%
+%% Note: this does not support multibag itself
+-spec resolve_siblings(pid(), RawBucket::binary(), RawKey::binary()) -> ok | {error, term()}.
+resolve_siblings(Pid, RawBucket, RawKey) ->
```
shino, line 214: It's good to output time at start and end, which may help investigating timeout errors.

src/riak_cs_console.erl
```diff
@@ -100,10 +102,84 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
...
+    GetOptions = [{r, all}],
+    _ = lager:info("Trying to resolve siblings of ~p:~p", [RawBucket, RawKey]),
+    case riakc_pb_socket:get(Pid, RawBucket, RawKey, GetOptions, ?DEFAULT_RIAK_TIMEOUT) of
+        {ok, RiakObj} ->
+            case resolve_ro_siblings(RiakObj, RawBucket, RawKey) of
```
shino, line 228: Output value count here?

src/riak_cs_console.erl
shino, line 214: Another subtle parameter in resolving (possibly) large object is custom timeout value.  Using large (like 10min) timeout value by default or add it as the fourth argument so that user can specify it.

src/riak_cs_console.erl
```diff
@@ -100,10 +102,84 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
...
 
+%% @doc Resolve siblings as quick operation. Assuming login via
+%% `riak-cs attach` RawKey and RawBucket should be provided via Riak
+%% log. Example usage is:
+%%
+%% > {ok, Pid} = riakc_pb_socket:start_link(Host, Port).
+%% > riak_cs_console:resolve_siblings(Pid, <<...>>, <<...>>).
+%% > riakc_pb_socket:stop(Pid).
+%%
+%% Note: this does not support multibag itself
```
shino: It is not obvious for me at first glance. Using Host/Port for proper bag suffices? Or this comment says just that it is not tested?

kuenishi:
I would like to include Taka's awesome tool to this pull request. Pending to next sprint.

src/riak_cs_console.erl
kuenishi, line 228: Output value count is printed out especially in case of manifests. That should depend on object types and not outputtting right here.

shino, line 228: I don't understand yet :sweat_smile: Why is sibling count less interesting for blocks / users / buckets?

kuenishi, line 228: I meant the merge functions do not have always clean interfaces to show sibling numbers (+ manifest histories), but I added the one just shows results of `riakc_obj:value_count/1`.

kuenishi:
Updated this branch; I believe I've addressed all your comments ^^;.

ksauzz:
testtest

src/riak_cs_console.erl
```diff
@@ -100,10 +102,104 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
...
+resolve_siblings(Pid, RawBucket, RawKey) ->
+    GetOptions = [{r, all}],
+    PutOptions = [{w, all}],
+    resolve_siblings(Pid, RawBucket, RawKey,
+                     GetOptions, ?DEFAULT_RIAK_TIMEOUT,
+                     PutOptions, ?DEFAULT_RIAK_TIMEOUT).
+
+resolve_siblings(Pid, RawBucket, RawKey, GetOptions, GetTimeout, PutOptions, PutTimeout)
+  when is_integer(GetTimeout) andalso is_integer(PutTimeout) ->
+    io:format("~p: getting ~p:~p~n", [erlang:now(), RawBucket, RawKey]),
```
shino: I prefer `os:timestamp()` unless monotonicity is needed.

src/riak_cs_console.erl
```diff
@@ -100,10 +102,104 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
...
+                      [erlang:now(), riakc_obj:value_count(RiakObj), RawBucket, RawKey]),
+            _ = lager:info("Trying to resolve ~p sibling(s) of ~p:~p",
+                           [riakc_obj:value_count(RiakObj), RawBucket, RawKey]),
+            case resolve_ro_siblings(RiakObj, RawBucket, RawKey) of
+                {ok, RiakObj2} ->
+                    R = riakc_pb_socket:put(Pid, RiakObj2, PutOptions, PutTimeout),
+                    io:format("~p: Resolution result: ~p~n", [erlang:now(), R]);
+                {error, R} = E ->
+                    io:format("~p: Not updating ~p:~p: ~p~n",
+                              [erlang:now(), RawBucket, RawKey, R]),
```
shino: Ditto.

src/riak_cs_console.erl
```diff
@@ -100,10 +102,104 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
...
+                     PutOptions, ?DEFAULT_RIAK_TIMEOUT).
+
+resolve_siblings(Pid, RawBucket, RawKey, GetOptions, GetTimeout, PutOptions, PutTimeout)
+  when is_integer(GetTimeout) andalso is_integer(PutTimeout) ->
+    io:format("~p: getting ~p:~p~n", [erlang:now(), RawBucket, RawKey]),
+    case riakc_pb_socket:get(Pid, RawBucket, RawKey, GetOptions, GetTimeout) of
+        {ok, RiakObj} ->
+            io:format("~p: Trying to resolve ~p sibling(s) of ~p:~p~n",
+                      [erlang:now(), riakc_obj:value_count(RiakObj), RawBucket, RawKey]),
+            _ = lager:info("Trying to resolve ~p sibling(s) of ~p:~p",
```
shino, line 226: Why lager is used only here?

shino, line 226: I don't think it's wrong. Just want to know coherence :sunglasses:

src/riak_cs_console.erl
kuenishi: To align with timestamps in logs, I'd prefer `calendar:universal_time()`. thoughts?

src/riak_cs_console.erl
shino: This log line is to indicate error occurrence. Is it better to include the word "error" (or "ERROR" or "Error ...) for grep'pability?

kuenishi:
Given that this function could be used via external scripts (not only console), I'll change all `io:format` to log output.

src/riak_cs_console.erl
```diff
@@ -100,10 +102,104 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
...
+    {ok, Obj};
+resolve_ro_siblings(RO, <<"0b:", _/binary>>, _) ->
+    case riak_cs_utils:resolve_robj_siblings(riakc_obj:get_contents(RO)) of
+        {{MD, Value}, true} when is_binary(Value) ->
+            RO1 = riakc_obj:update_metadata(RO, MD),
+            {ok, riakc_obj:update_value(RO1, Value)};
+        {E, true} ->
+            io:format("Cannot resolve: ~p~n", [E]),
+            {error, E};
+        {E, false} ->
```
shino: `false` in second element represents there are no siblings (single value), right?

src/riak_cs_console.erl
```diff
@@ -100,10 +102,104 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
...
+            {error, E}
+    end;
+resolve_ro_siblings(RiakObject, <<"0o:", _/binary>>, _RawKey) ->
+    [{_, Manifest}|_] = Manifests =
+        riak_cs_manifest:manifests_from_riak_object(RiakObject),
```
shino, line 288: There is possibility that no manifest history after `manifests_from_riak_object` (e.g. all UUIDs are pruned.)

src/riak_cs_console.erl
```diff
@@ -100,10 +102,104 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
...
+    %% up), operator must choose who is true owner.
+    {error, {not_supported, ?BUCKETS_BUCKET}};
+resolve_ro_siblings(RO, ?GC_BUCKET, _) ->
+    Resolved = riak_cs_gc:decode_and_merge_siblings(RO, twop_set:new()),
+    Obj = riak_cs_utils:update_obj_value(RO,
+                                         riak_cs_utils:encode_term(Resolved)),
+    {ok, Obj};
+resolve_ro_siblings(RO, <<"0b:", _/binary>>, _) ->
+    case riak_cs_utils:resolve_robj_siblings(riakc_obj:get_contents(RO)) of
+        {{MD, Value}, true} when is_binary(Value) ->
```
shino, line 277: Wondering how tombstones should be treated... Put them back without deleting or delete them explicitly.

src/riak_cs_console.erl
```diff
@@ -100,10 +102,104 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
...
 
+%% @doc Resolve siblings as quick operation. Assuming login via
+%% `riak-cs attach` RawKey and RawBucket should be provided via Riak
+%% log. Example usage is:
+%%
+%% > {ok, Pid} = riakc_pb_socket:start_link(Host, Port).
+%% > riak_cs_console:resolve_siblings(Pid, <<...>>, <<...>>).
+%% > riakc_pb_socket:stop(Pid).
+%%
+%% or, for CS Bucket and Key name:
```
shino: The code below are for manifests. "for manifests with CS Bucket and Key name:" ?

src/riak_cs_console.erl
```diff
@@ -100,10 +102,104 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
...
+                      [erlang:now(), Reason]),
+            E
+    end.
+
 %%%===================================================================
 %%% Internal functions
 %%%===================================================================
 
+-spec resolve_ro_siblings(riakc_obj:riakc_obj(), binary(), binary()) ->
+                                 {ok, riakc_obj:riakc_obj()} | {error, term()}.
```
shino: Third argument is not used and riakc_obj has the key inside it. Unnecessary?

kuenishi: For now, yes. But I just wanted a case to match something with keys

src/riak_cs_console.erl
kuenishi, line 277: block server also puts back the most likely block, so here's the same logic.

src/riak_cs_console.erl
kuenishi, line 288: That shouldn't happen I'm afraid, and no need to resolve it, I think.

src/riak_cs_console.erl
kuenishi: Well, yes

kuenishi:
force-pushed :imp:

src/riak_cs_console.erl
```diff
@@ -183,10 +185,119 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
     _ = io:format("~nAll old unaborted orphan multipart uploads have been deleted.~n", []).
 
 
+-spec resolve_siblings(binary(), binary()) -> any().
+resolve_siblings(RawBucket, RawKey) ->
+    {ok, RcPid} = riak_cs_riak_client:start_link([]),
+    try
+        {ok, Pid} = case RawBucket of
+                        <<"0o:", _/binary>> ->
+                            riak_cs_riak_client:manifest_pbc(RcPid);
```
shino: This does not work for multibag environment because riak_cs_riak_client does not know which bag it should use for manifest.

kuenishi: Can we just take this function for single-bag cluster and tests?

src/riak_cs_console.erl
```diff
@@ -183,10 +185,119 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
...
+            _ = lager:info("Trying to resolve ~p sibling(s) of ~p:~p",
+                           [riakc_obj:value_count(RiakObj), RawBucket, RawKey]),
+            case resolve_ro_siblings(RiakObj, RawBucket, RawKey) of
+                {ok, RiakObj2} ->
+                    R = riakc_pb_socket:put(Pid, RiakObj2, PutOptions, PutTimeout),
+                    _ = lager:info("Resolution result: ~p~n", [R]),
+                    R;
+                {error, R} = E ->
+                    _ = lager:info("Not updating ~p:~p: ~p~n",
+                                   [RawBucket, RawKey, R]),
```
shino: This log line is to indicate error occurrence. Is it better to include the word "error" (or "ERROR" or "fail" ...) for grep'pability?

src/riak_cs_console.erl
```diff
@@ -183,10 +185,119 @@ cleanup_orphan_multipart(Timestamp) when is_binary(Timestamp) ->
...
+    {ok, Obj};
+resolve_ro_siblings(RO, <<"0b:", _/binary>>, _) ->
+    case riak_cs_utils:resolve_robj_siblings(riakc_obj:get_contents(RO)) of
+        {{MD, Value}, true} when is_binary(Value) ->
+            RO1 = riakc_obj:update_metadata(RO, MD),
+            {ok, riakc_obj:update_value(RO1, Value)};
+        {E, true} ->
+            _ = lager:info("Cannot resolve: ~p~n", [E]),
+            {error, E};
+        {E, false} ->
```
shino: false in second element represents there are no siblings (single value), right?

src/riak_cs_console.erl
shino, line 288: It happened actually, but I accept it as restriction of this scripts :smile:

src/riak_cs_console.erl
kuenishi: I don't think so, if that's error it should have log level as error but this is manual operation tool so I assume the operator's watching the log and see it. If needed, it should be grep'd with `Not updating` .

src/riak_cs_console.erl
kuenishi: Yes

src/riak_cs_console.erl
shino: Ah, OK. Then I think three branches are not good because it wrongly looks like multibag ready. Just using `master_pbc` is sufficient.

src/riak_cs_console.erl
shino: Is it so hard to `s/R/E/` ?

kuenishi: No :)

kuenishi:
Updated

shino:
Simple and effective :+1:

shino:
@borshop merge

Basho-JIRA:
For release note: Add a generic function for manual operations to resolve siblings of manifests and blocks 

PR #1191, opened 2015-07-16
Fast user get option [JIRA: RCS-240]

kuenishi:
If `fast_user_get` is turned on, all requests to get user record skip PR=3 get to Riak to avoid unnecessary unavailability. The case this merits is where one of Riak nodes is ~~down~~ slow or going to down, get_fsm should wait for all of them or request fails. For consistency PW=3 when updating user record would be enough. `fast_user_get` is off by default now, but will be on by default in some future.

All riak_test items have passed with this option (except `stats_test` which is currently failing continously), but I don't think additional riak_test for this because current riak_test does not simulate any failure (which we have to do some).

### DIFF
--- a/rel/files/riak_cs.schema
+++ b/rel/files/riak_cs.schema
@@ -216,8 +216,9 @@
 %% to omit PR=all get at retrieving user object. When this is turned
 %% on, instead gets are performed with R=quorum and PR=one to avoid
 %% waiting for responses from slow node. The price for this option is
-%% a very slight possibility of stale user object (such as a list of
-%% owned buckets) being read.
+%% possibility to read stale user object (such as a list of owned
+%% buckets) being read, as well as the list of buckets in a user
+%% record never pruned.
 {mapping, "fast_user_get", "riak_cs.fast_user_get", [
   {default, off},
   {datatype, flag}
--- a/rel/files/riak_cs.schema
+++ b/rel/files/riak_cs.schema
@@ -211,7 +211,13 @@
   {datatype, bytesize}
 ]}.
 
-%% @doc Whether to omit PR=all get at retrieving user info.
+%% @doc When some of the nodes are slow, getting user with PR=all may
+%% make all requests of the user slow. This option stands for whether
+%% to omit PR=all get at retrieving user object. When this is turned
+%% on, instead gets are performed with R=quorum and PR=one to avoid
+%% waiting for responses from slow node. The price for this option is
+%% a very slight possibility of stale user object (such as a list of
+%% owned buckets) being read.
 {mapping, "fast_user_get", "riak_cs.fast_user_get", [
   {default, off},
   {datatype, flag}
--- a/rel/files/riak_cs.schema
+++ b/rel/files/riak_cs.schema
@@ -211,6 +211,12 @@
   {datatype, bytesize}
 ]}.
 
+%% @doc Whether to omit PR=all get at retrieving user info.
+{mapping, "fast_user_get", "riak_cs.fast_user_get", [
+  {default, off},
+  {datatype, flag}
+]}.
+
 %% == Access statistics ==
 
 %% @doc How often to flush the access stats; integer factor of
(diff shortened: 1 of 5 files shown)


shino:
Nice PR! I agree this is great direction for latency improvements.

I give big :+1: for the simple(st) approach of current diff is sufficient for some users like private object storage purpose.  Maybe not in the scope of this PR, for more general purpose or finally to make this fast path by default setting, I want to discuss on pruning of bucket list [1] and consistency.

Pruning path along with fast path can be achieved by using the strong option to get user objects in some of APIs.  To avoid latency issue, such APIs should be not hot APIs.  The tentative list is:
- PUT Bucket S3 API and
- DELETE Bucket S3 API.

For consistency, it would be safer to use strong option for user modifying API calls. These includes following ones as well as above two S3 APIs.
- POST User admin API (create user)
- PUT User admin API (modify user)

It's no problem if these discussion will not be reflected in this PR :) I will just create new issue or PR based on this comment.  Aside from it, any comments/correction/etc are appreciated.

---

[1] Short description of pruning deleted buckets

skip this if you already know about it :)

There is a bucket list in user records and deleted buckets will be there as a kind of tombstone.  These tombstones should be removed in any way eventually in general purpose, e.g. public object storage, to keep the list bounded.  The pruned object will be overwritten _only_ when user objects has been fetched by the strong option.

kuenishi:
Those thoughts are definitely needed when we make this option on by default. However, I thought such APIs finally go to Stanchion and PR=3 and PW=3 saves consistency ... but it does not sound like that ...

In reality, currently most `get_user` call is done via `riak_cs_wm_common` , for all API calls. Maybe we have to make the common module API-aware, which is a little bit more hustle. We also need another riak_test injecting node failure by kill or anything. Let's do it after 2.1.0 get out. Will open another issue for that.

shino:
> However, I thought such APIs finally go to Stanchion and PR=3 and PW=3 saves consistency ... but it does not sound like that ...

Agreeeeeee. Currently some update for bucket is done by stanchion for bucket objects (and maybe user objects) and done by riak_cs for user objects. This is a problem but totally another issue :see_no_evil:

rel/files/riak_cs.schema
```diff
@@ -211,6 +211,12 @@
   {datatype, bytesize}
 ]}.
 
+%% @doc Whether to omit PR=all get at retrieving user info.
```
shino: If this is not hidden, warning should be included that pruning never happens.

shino:
Will :+1: after discussing on riak-cs.conf documentation and rebasing on current develop.

kuenishi:
Updated.

rel/files/riak_cs.schema
kuenishi: What pruning did you mean?

kuenishi:
Updated.

src/riak_cs_user.erl
```diff
@@ -30,6 +30,7 @@
          is_admin/1,
          get_user/2,
          get_user_by_index/3,
+         from_riakc_obj/2,
```
shino: Not used from outside?

kuenishi: Ah. This was remain of my effort to try consistency updating user object. Fruitless effort without strong consistency.

kuenishi:
Updated.

shino:
@borshop merge

kuenishi:
For release note:

~~Add an option to replace N=all user get option with N=quorum just before authentication. This improves all latency especially in presence of slow (or actually-failing) node, who blocks the whole request flow because of N=all. The price of turning this on is, A user's owned bucket list is never pruned after a bucket is deleted, instead just marked as deleted (RCS-242 / #1193 (RCS-242) (RCS-242) ).~~ This is wrong.

kuenishi:
For release note (correct version):

Add an option to replace `PR=all` user get option with `PR=one` and `R=quorum` just before authentication. This improves all latency especially in presence of slow (or actually-failing) node, who blocks the whole request flow because of `PR=all`. The price of turning this on is, A user's owned bucket list is never pruned after a bucket is deleted, instead just marked as deleted (RCS-242 / #1193 (RCS-242) (RCS-242) (RCS-242) (RCS-242) (RCS-242) (RCS-242) )

PR #1195, opened 2015-07-17
Fix two lines of wrong rtcs:pass() usage in access_stats_test

shino:
The bug was introduced at commit a07eb28cbb3500d2295e395504851c42f29e5055.
This PR is a partial revert ot it.


kuenishi:
@borshop merge

PR #1196, opened 2015-07-21
Add max_scheduled_delete_manifests config into cuttlefish schema [RCS-247]

ksauzz:
This PR contains:
- add max_scheduled_delete_manifests into riak_cs.schema
- change default value of max_scheduled_delete_manifests from unlimited to 50

Changing the default value seems to need a discussion. However, in my opinion, it would be worth to change the default value from `unlimited`, since sometimes manifest histories grows up to large size in some use-case.

### DIFF
--- a/src/riak_cs_gc.erl
+++ b/src/riak_cs_gc.erl
@@ -311,7 +311,7 @@ set_leeway_seconds(_LeewaySeconds) ->
 max_scheduled_delete_manifests() ->
     case application:get_env(riak_cs, max_scheduled_delete_manifests) of
         undefined ->
-            unlimited;
+            ?DEFAULT_MAX_SCHEDULED_DELETE_MANIFESTS;
         {ok, MaxCount} ->
             MaxCount
     end.
--- a/include/riak_cs_gc.hrl
+++ b/include/riak_cs_gc.hrl
@@ -94,6 +94,7 @@
 -define(DEFAULT_GC_BATCH_SIZE, 1000).
 -define(DEFAULT_GC_WORKERS, 2).
 -define(EPOCH_START, <<"0">>).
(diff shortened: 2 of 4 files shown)


src/riak_cs_gc.erl
```diff
@@ -311,7 +311,7 @@ set_leeway_seconds(_LeewaySeconds) ->
 max_scheduled_delete_manifests() ->
     case application:get_env(riak_cs, max_scheduled_delete_manifests) of
```
shino, line 312: Case branch for `unlimited` is needed.

ksauzz, line 312: `unlimited` matches `{ok, MaxCount}` branch, right?

shino, line 312: You are right! :see_no_evil:

ksauzz, line 312: m9(^Д^)

shino:
> it would be worth to change the default value from unlimited, since
>  sometimes manifest histories grows up to large size in some
>  use-case.

:+1:

Because `scheduled_delete` state is tombstone in order to prevent resurrection of
deleted UUIDs, finite value for `max_scheduled_delete_manifests` may results
in resurrection.  Nonetheless, bloated manifest object history will do
harm in performance and/or memory footprint.
The usage pattern that is affected by finite value is frequent overwriting
for certain key, then one can expect resurrection will be repaired soon
even if it happens.

P.S. (just for information)

There is another pruning logic in riak cs based on deleted time,
enabled by default.  If a `scheduled_delete` manifest that is older
than leeway period (by default, 24hours), it is pruned in memory. If
manifest overwrite happens (e.g. PUT Object), pruned history is stored
in Riak.

shino:
@borshop merge

PR #1198, opened 2015-07-23
Misc improvements around build

shino:
This PR consists of three independent commits.
- Add compile-riak-test to dependency of copile target c2f8057
- Skip deps in clean-riak-test 03628e9
- Add `--all` option to "git add" to suppress deprecation warning 01c7b81

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ clean-client-test:
 	@./rebar client_test_clean
 
 clean-riak-test:
-	@./rebar riak_test_clean
+	@./rebar riak_test_clean skip_deps=true
 
 deps:
 	@./rebar get-deps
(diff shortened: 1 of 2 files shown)


shino:
Example of warning message from `git add` without `--all` is like this:

```
warning: You ran 'git add' with neither '-A (--all)' or '--ignore-removal',
whose behaviour will change in Git 2.0 with respect to paths you removed.
Paths like 'current/dev/dev1/data/generated.configs/app.2015.07.23.15.21.35.config' that are
removed from your working tree are ignored with this version of Git.

* 'git add --ignore-removal <pathspec>', which is the current default,
  ignores paths you removed from your working tree.

* 'git add --all <pathspec>' will let you also record the removals.

Run 'git status' to check the paths you removed from your working tree.
```

shino:
@borshop merge

PR #1218, opened 2015-08-14
Bugfix: Missing keys in JSON for detailed storage calc of empty buckets [JIRA: RCS-267]

shino:
Fix missing `"Objects"` and `"Bytes"` keys in storage stats for empty
buckets.

Before this PR, detailed storage calc JSON result for empty
buckets looks like:

```
{
    "bucket-1": {},
    "bucket-2": {},
    "EndTime": "20150814T023756Z",
    "StartTime": "20150814T023756Z"
}
```

Because non-detialed storage calc adds keys of `"Objects"` and
`"Bytes"` even for empty buckets, this PR adds such keys for detailed
ones and riak_test assertions for such empty bucket storage calc.


Basho-JIRA:
For relnote: No need to mention because a bug fix for 2.1.0 development cycle. 

**_[posted via JIRA by Shunichi Shinohara]**_

kuenishi:
@borshop merge

PR #1223, opened 2015-08-25
Add requestPayment and lifecycle stub callbacks [JIRA: RCS-272]

shino:
Before this PR,
- GET Bucket Lifecycle returns 403 Access Denied because `lifecycle` is not
  included in subresources definition of Riak CS. 
- GET Bucket RequestPayment returns 404 Not Found because no dispatch rule for it.

In order to response a little more understandable response, this PR changes them as:
- For requestPayment, just reply XML with "BucketOwner" after authn/authz.
- For lifecycle, just reply error XML of "NotImplemented" without anthentication.

---

To be truthful, motivation for these changes were from s3cmd error for `s3cmd info <bucket>` :secret:

### DIFF
--- a/src/riak_cs_web.erl
+++ b/src/riak_cs_web.erl
@@ -71,9 +71,13 @@ base_resources() ->
      {["buckets", bucket, "objects"], riak_cs_wm_common, props(riak_cs_wm_objects)},
      {["buckets", bucket, "delete"], riak_cs_wm_common, props(riak_cs_wm_bucket_delete)},
      {["buckets", bucket, "acl"], riak_cs_wm_common, props(riak_cs_wm_bucket_acl)},
+     %% Just stub, no dynamic contents
      {["buckets", bucket, "location"], riak_cs_wm_common, props(riak_cs_wm_bucket_location)},
      {["buckets", bucket, "versioning"], riak_cs_wm_common, props(riak_cs_wm_bucket_versioning)},
+     {["buckets", bucket, "requestPayment"], riak_cs_wm_common, props(riak_cs_wm_bucket_request_payment)},
+     %% NYI
      {["buckets", bucket, "versions"], riak_cs_wm_common, props(riak_cs_wm_not_implemented)},
+     {["buckets", bucket, "lifecycle"], riak_cs_wm_common, props(riak_cs_wm_not_implemented)},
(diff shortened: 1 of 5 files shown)


src/riak_cs_web.erl
```diff
@@ -71,9 +71,13 @@ base_resources() ->
      {["buckets", bucket, "objects"], riak_cs_wm_common, props(riak_cs_wm_objects)},
      {["buckets", bucket, "delete"], riak_cs_wm_common, props(riak_cs_wm_bucket_delete)},
      {["buckets", bucket, "acl"], riak_cs_wm_common, props(riak_cs_wm_bucket_acl)},
+     %% Just stub, no dynamic contents
      {["buckets", bucket, "location"], riak_cs_wm_common, props(riak_cs_wm_bucket_location)},
```
kuenishi, line 74: location is no more a stub because it's return value is configurable via configuration.

src/riak_cs_web.erl
```diff
@@ -71,9 +71,13 @@ base_resources() ->
...
      {["buckets", bucket, "versioning"], riak_cs_wm_common, props(riak_cs_wm_bucket_versioning)},
+     {["buckets", bucket, "requestPayment"], riak_cs_wm_common, props(riak_cs_wm_bucket_request_payment)},
```
kuenishi, line 77: Can we just move versioning and requestPayment to NYI section and be handled via riak_cs_wm_not_implemented ? Does s3cmd get messed up with that?

shino, line 77: What is good point to use not_implemented callback than response that contains not-zero information?

src/riak_cs_web.erl
shino, line 74: Ah, reasonable :) I did so just because location does not have meaning anywhere else in riak cs (except a possibility of v4 auth.)

src/riak_cs_web.erl
kuenishi, line 77: ah, ok.

shino:
Updated.

shino:
@borshop merge

Basho-JIRA:
Release note: Add 501 stub for Bucket lifecycle and simple stub for Bucket requestPayment 

PR #1228, opened 2015-08-27
Increase request pool size on external client tests [JIRA: RCS-280]

ksauzz:
This PR has 2 changes:
- Improve rtcs:setup/x to merge specified prams into default prams
- Increase request_pool size on external client tests

Sometimes request pool workers of poolboy is exhausted on external_client_tests, then it causes the
test fails, since external_client_tests runs individual client tests concurrently using 'make -j 8'.
So pool size should be increased enough.
I changed rtcs:setup to merge specified params into default configs before Increasing pool size, to do it easily.

### DIFF
--- a/riak_test/src/rtcs.erl
+++ b/riak_test/src/rtcs.erl
@@ -33,7 +33,7 @@
 -define(CSDEV(N), list_to_atom(?CSDEVS(N))).
 
 setup(NumNodes) ->
-    setup(NumNodes, rtcs_config:default_configs(), current).
+    setup(NumNodes, [], current).
 
 setup(NumNodes, Configs) ->
     setup(NumNodes, Configs, current).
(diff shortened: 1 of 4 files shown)


kuenishi:
The reason why we had small connection pools by default is to find bugs by setting more tight parameters. If it's an issue in specific test, the parameters should be specified in each test setup, external_client_tests for this case.

kuenishi:
Fm, I was wrong ^^;

ksauzz:
ha, my 3rd line in the description was unclear. Updated.

shino:
Simpler API than before. Nice change :smile:

shino:
+1 14c47a1

shino:
@borshop merge

kuenishi:
No release note needed as this is a test code update

Issue #1201, opened 2015-07-29
Timeout in get user with strong option makes subsequent weak get fail [JIRA: RCS-250]


shino:
Symptom : single slow node may cause user fetch failure.
- Getting CS user is in two steps, first step with PR=all,
  in which single slow riak node can cause timeout error at client.
- When timeout occurs in `riakc_pb_socket`, it disconnects TCP connection
  and goes into wait-and-retry loop.
- Then, CS user get 2nd phase with weak option, but it's likely that reconnect
  does not happen yet, fails with disconnected error.

If "slow" node is completely frozen (no action will come out from it),
after health check timeout, strong get fails by "insufficient vnodes"
and weak get should work well.  For this case, certain user can not
access Riak CS for finite time period, 60 sec by default.

---

Reproduction (or simulation)
- Create 4-node cluster (`{get_user_timeout, 3000}` in advanced.config may help)
- Memo dev2 pid
  
  ```
  DEV2=`ps aux | grep riak_ee | grep dev1 | grep beam.smp | awk '{print $2;}' `
  ```
- Freeze it: `kill -s SIGSTOP $DEV2` (keep your fingers crossed, if unfortunate, freeze another node :hear_no_evil:)
- Do any access,

shino:
For release note, short version: Improve user object fetch logic when some nodes are slow or silently failed.
For longer version, please refer the description of this issue.

shino:
Addressed by #1230

PR #1230, opened 2015-08-31
Give grace period for `riakc_pb_socket` to reconnect

shino:
This PR addresses #1201 (RCS-250) (RCS-250) .

When `riakc_pb_socket` detects timeout errors, it disconnects the
TCP connection and enters reconnection phase.  Immediate next
request to `riakc_pb_socket` process will likely fail with
infamous `disconnect` errors.

Typical case is described in #1201 (RCS-250) (RCS-250) for user object fetch.
Also, there is another case for block get. If the first
primary vnode for a block is very slow, the reuqest will
timeout because it is `N=one`. Then subsequent `N=all`
request will likely fail.

This PR fixes these two cases by adding grace period so that
`riakc_pb_socket` can reconnect if possible.

### DIFF
--- a/src/riak_cs_block_server.erl
+++ b/src/riak_cs_block_server.erl
@@ -294,6 +294,7 @@ try_local_get(RcPid, FullBucket, FullKey, GetOptions1, GetOptions2,
                           Why == disconnected;
                           Why == <<"{insufficient_vnodes,0,need,1}">>;
                           Why == {insufficient_vnodes,0,need,1} ->
+            _ = riak_cs_pbc:pause_to_reconnect(block_pbc(RcPid), Why, Timeout),
             handle_local_notfound(RcPid, FullBucket, FullKey, GetOptions2,
                                   ProceedFun, RetryFun,
                                   [{local_one, Why}|ErrorReasons], UseProxyGet,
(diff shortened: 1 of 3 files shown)


shino:
Rebased on develop 1643c8d to incorporate r_t fix #1231

kuenishi:
~~Currently we have `auto_reconnect` as `false` by default, so the recovery from `disconnected` state is by just re-launching the process by poolboy?~~ [edit]We had auto_reconnect as true by default. I was wrong.

shino:
> by just re-launching the process by poolboy?

Does "the process" indicate `riak_cs_riak_client` process? Or `riakc_pb_socket`?

kuenishi:
I think it was riak_cs_riak_client.

shino:
It may be my mind model of riak cs code base... some processes share single riak_cs_riak_client checked out from a pool, and (at least I think) the pid is unchangable in a request-response lifecycle. If it should be changed or renewed, all the processes that share it should be notified and renew to new pid. At the current implementation, I'm afraid that it might be a big diff covering many modules in the final stage of 2.1.0 development cycle.

By the example of this PR's current diff, block server process detects `timeout` but it can not notify put/get FSM and/or mochiweb acceptor processes which has the pid in `#context`.

Although, I prefer any other way to just sleeping in general if possible. This PR is not mandatory for 2.1.0 because it's long standing issue :sweat:

Some random thoughts for possible next PR creator:
- Slow node or silent node failure may prevent some users from accessing Riak CS system _completely_ (because user object can not be fetched) typically for 60 seconds.
- Using `queue_if_disconnected` encapsulates `disconnected` state in `riakc_pb_socket` and hides it from riak cs, it will be not good for visibility.

shino:
@borshop merge

kuenishi:
GC also fails at `riak_cs_gc_key_list.erl` like this:

```
2015-09-07 18:17:29.172 [warning] <0.10473.4>@riak_cs_gc_key_list:eligible_manifest_keys:122 Error occurred trying to query from time <<"0">> to <<"1441531049">>in gc key index. Reason: disconnected
```

Maybe we'd also need this [here](https://github.com/basho/riak_cs/blob/66d081601302c174ac2e0442ff54084757bb843f/src/riak_cs_gc_key_list.erl#L169). ... but the code change couldn't be small. fmm.

PR #1232, opened 2015-09-01
Suppress noisy logs  [JIRA: RCS-281]

shino:
This PR consists of two independent commits.
### commit 8960fe2386b3dd9e1de47b0e95f89650eea1f0f1

This commit Suppresses warning log of unsatisfied PR in strong user get.
It is noisy because single node down should be considered as normal state
and neither direct client effects (weak user get will follow) nor operator
action needed.

To reproduce, setup multiple node cluster and stop one of the nodes.
Almost any request to riak cs let warning log like :

```
18:01:37.070 [warning] Fetching user record with strong option failed:
  <<"{pr_val_unsatisfied,3,2}">>
```
### commit f95acf0d14fe2ebee3d90cf3a774c0f9472147db

This filters out undefined log data notes at the beginning of handle_event.
It is a fix of bug of 2.1 development cycle (#1130).

Following log is an example which is output when GET `?torrent` is called.

```
12:15:29.888 [error] gen_event riak_cs_access_log_handler installed in
  webmachine_log_event terminated with reason: bad argument in call to
  lists:keyfind({access,user}, 1, undefined) in
  riak_cs_access_log_handler:handle_event/2 line 189
```

The fix revives log of "No WM route" that have been exists prior 2.1.
This commit also changes its log severity from error to debug
because the error indicates client error except in development
phase.  Just responding 4xx is sufficient.

### DIFF
--- a/src/riak_cs_riak_client.erl
+++ b/src/riak_cs_riak_client.erl
@@ -331,6 +331,8 @@ get_user_with_pbc(MasterPbc, Key, true) ->
 get_user_with_pbc(MasterPbc, Key, false) ->
     case strong_get_user_with_pbc(MasterPbc, Key) of
         {ok, _} = OK -> OK;
+        {error, <<"{pr_val_unsatisfied,", _/binary>>} ->
+            weak_get_user_with_pbc(MasterPbc, Key);
         {error, Reason} ->
             _ = lager:warning("Fetching user record with strong option failed: ~p", [Reason]),
             weak_get_user_with_pbc(MasterPbc, Key)
--- a/src/riak_cs_access_log_handler.erl
+++ b/src/riak_cs_access_log_handler.erl
@@ -182,9 +182,13 @@ handle_call(_Request, State) ->
     {ok, ok, State}.
 
 %% @private
+handle_event({log_access, #wm_log_data{notes=undefined,
+                                       method=Method, path=Path, headers=Headers}},
+             State) ->
+    lager:debug("No WM route: ~p ~s ~p\n", [Method, Path, Headers]),
+    {ok, State};
 handle_event({log_access, LogData},
              #state{table=T, size=S, max_size=MaxS}=State) ->
-
     %% Updates for quotas
     case lists:keyfind(?STAT(user), 1, LogData#wm_log_data.notes) of
         {?STAT(user), Key} ->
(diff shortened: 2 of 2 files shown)


shino:
@borshop merge

src/riak_cs_riak_client.erl
```diff
@@ -331,6 +331,8 @@ get_user_with_pbc(MasterPbc, Key, true) ->
 get_user_with_pbc(MasterPbc, Key, false) ->
     case strong_get_user_with_pbc(MasterPbc, Key) of
         {ok, _} = OK -> OK;
+        {error, <<"{pr_val_unsatisfied,", _/binary>>} ->
```
kuenishi, line 334: How about matching fully with `{pr_val_unsatisfied,3,2}` ? `3,2` should be negligible but could `3,1` happen in faulty cluster and that's the case we want to warn operators?

kuenishi, line 334: fmm... 3 is not a good case when `n_val` is not 3.

shino, line 334: > when n_val is not 3. 

Exactly :bulb: 

> could 3,1 happen in faulty cluster and that's the case we want to warn operators?

Worth consideration :thought_balloon: Just my opinion, cluster membership should be monitored in riak side, not from riak cs logs. The reason is monitoring target should be one-time event (node-up, node-down) but riak cs can, at best, provide noisy, repeating-over-and-over log lines of warnings.

kuenishi, line 334: We can match it when n_val is not 3 (but less than 10).

``` erlang
24> F=fun(<<"{pr_val_unsatisfied,", N, ",",M, "}">>) when (M-$0)*2 > (N-$0) -> ok; (_) -> ng end.
#Fun<erl_eval.6.80484245>
25> F(<<"{pr_val_unsatisfied,3,2}">>).                                          
ok               
26> F(<<"{pr_val_unsatisfied,5,3}">>).                                          
ok               
27> F(<<"{pr_val_unsatisfied,3,1}">>).
ng
28> F(<<"{pr_val_unsatisfied,5,2}">>).
ng
```

I think `{3,2}` case is unnecessary and noisy but `{3,1}` case is _NOT_.

shino, line 334: The PR which includes this commits have been merged already. Please create new PR for the change :smile: 

Assuming 2,000 requests and 100 bytes per log line, log size 11MB per
minutes or 670 MB per hour. If lagger setting were default, 5-minute old log
would start disappearing.

Basho-JIRA:
DUP with RCS-274 

Issue #1229, opened 2015-08-31
Race conditions for gc status [JIRA: RCS-273]


shino:
Two race conditions:
- State (idle or runnning) and detialed information
  - StateName is come from riak_cs_gc_manager (gen_fsm)
  - Detailed information is from riak_cs_gc_batch (gen_server)
  - it's possible that the command says "A garbage collection batch is
    in progress" but no detailed information
- riak_cs_gc_batch may be dead at riak_cs_gc_batch:current_state(Pid)

Basho-JIRA:
No release note needed: new bug during development cycle. 

**_[posted via JIRA by Kota Uenishi]**_

Basho-JIRA:
This was introduced at RCS-199 

**_[posted via JIRA by Kota Uenishi]**_

PR #1234, opened 2015-09-03
Bugfix/gc status race

shino:
Addresses #1229 (RCS-273)

Before this PR, there are two race conditions:
1. Manager's state (idle or running) and detailed information from a batch process.
   It's possible that the command says "A garbage collection batch is
   in progress" but less detailed information
2. riak_cs_gc_batch may had been dead at riak_cs_gc_batch:current_state(Pid).

Introduced 'finishing' state for command output (not an actual state of certan
fsm). Just output "finishing" but no details for it in command line.
Also added try/catch to riak_cs_gc_batch:current_state() call

The first commit is the actual fix and second one is just for cosmetics.


kuenishi:
@borshop merge

kuenishi:
@borshop: retry

Issue #1166, opened 2015-06-16
Header parsing in AWS v4 should ignore spaces after commas [JIRA: RCS-214]


gdoteof:
Riak_cs appears to be strict in it's parsing of the `Authorization` header.

Header that looks for example like (NOTE SPACES BETWEEN FIELDS): 

```
Authorization: AWS4-HMAC-SHA256 Credential=QQS2D16QNR6NI1EXAMPLE/20150616/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=242a08fc8b580025f3e02dde5bcdf2dd5d36c614d401252489a8EXAMPLE
```

Will parse the AuthResult as (note space before `SignedHeaders`)

```
[{"Credential":"QQ.."}, {" SignedHeaders":"host.."}..]
```

And then the lookup will fail

here(ish):

https://github.com/basho/riak_cs/blob/03032ffb326a20ac9961b1b3c6e1b9fbffc18583/src/riak_cs_s3_auth.erl#L198

shino:
@gdoteof Thanks for reporting.

By reading the AWS S3 doc [1], it seems that the separator for these
elements is only a comma, but a space is not mentioned.

```
The following is an example of the Authorization header value. Line
breaks are added to this example for readability:

Authorization: AWS4-HMAC-SHA256 Credential=AKIAIOSFODNN7EXAMPLE/20130524/us-east-1/s3/aws4_request,
SignedHeaders=host;range;x-amz-date,
Signature=fe5f80f77d5fa3beca038a248ff027d0445342fe2855ddc963176630326f1024

Note that there is space between the first two components,
AWS4-HMAC-SHA256 and Credential, and that the subsequent components,
Credential, SignedHeaders, and Signature are separated by a comma.
```

Does your example with spaces after commas work well with AWS S3?

[1] http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html

gdoteof:
@shino Yes I did see that as well, I think it's rather ambiguous; especially as they added line breaks for "readability!"

Note here their python example _does_ use a space: http://docs.aws.amazon.com/general/latest/gr/sigv4-signed-request-examples.html

The library I found this with is https://github.com/brendanhay/amazonka -- and yes it does work with amazon s3 (this library hard codes the spaces in the header)

shino:
[...]

PR #1235, opened 2015-09-04
Trim spaces around commas

shino:
Addresses #1166 (RCS-214) 

AWS S3 python example does use a space after each comma:
http://docs.aws.amazon.com/general/latest/gr/sigv4-signed-request-examples.html


kuenishi:
I won't do manual reproduction with haskell libraries.... although `external_client_tests` passed.

shino:
@borshop merge

PR #1237, opened 2015-09-04
Return 'ok' from riak_cs_console:status/1

ksauzz:
riak-cs-admin status always returns 1 since riak_cs_console:status/1 return list of 'ok' even though [nodetool expects 'ok' as success of RPC call](https://github.com/basho/node_package/blob/2.0.3/priv/base/nodetool#L81-L92).

### DIFF
--- a/src/riak_cs_console.erl
+++ b/src/riak_cs_console.erl
@@ -38,7 +38,7 @@
 status([]) ->
     Stats = riak_cs_stats:get_stats(),
     _ = [io:format("~p : ~p~n", [Name, Value])
-     || {Name, Value} <- Stats],
+         || {Name, Value} <- Stats],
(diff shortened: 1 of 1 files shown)


src/riak_cs_console.erl
```diff
@@ -37,8 +37,9 @@
 
 status([]) ->
     Stats = riak_cs_stats:get_stats(),
-    [io:format("~p : ~p~n", [Name, Value])
-     || {Name, Value} <- Stats].
+    _ = [io:format("~p : ~p~n", [Name, Value])
+     || {Name, Value} <- Stats],
```
shino: I guess the indent is wrong.

shino:
@borshop merge

Issue #1219, opened 2015-08-14
Add start/end log to each storage calculation batch [JIRA: RCS-265]


kuenishi:
Found it's inconvenient while testing CS cluster.

Basho-JIRA:
For release note: Add info log when starting a storage calculation batch. 

**_[posted via JIRA by Kota Uenishi]**_

PR #1238, opened 2015-09-04
Add storage calculation start log

shino:
Addresses #1219 (RCS-265).

Finish log has been already there.


kuenishi:
I'd like some storage calculation options logged out as well as GC.

shino:
Updated

kuenishi:
@borshop merge

PR #1241, opened 2015-09-08
Add code formatter into Makefile

ksauzz:
Add a code formatter which is based on @kuenishi 's work.

`make code-format` lookup specified source codes and apply code formatter.


kuenishi:
Let's go here: https://github.com/basho/tools.mk

shino:
@ksauzz @kuenishi ping

PR #1247, opened 2015-09-11
Fix false state transition riak_cs_gc_batch to waiting_workers without any more tasks [RCS-291]

shino:
Before this commit, batch completion is verified _before_
maybe_start_workers call and maybe_start_workers always
make transitions to waiting_workers state.
When the count of gc keys is just multiple of gc_batch_size,
continuation indicates there (maybe) more results but
no more tasks actually. The riak_cs_gc_batch process is in
waiting_workers but no workers who triggers event transition.

This commit only changes the order of maybe_start_workers and
has_batch_finished.

---

Steps to reproduce:
1. Set gc.batch_size and gc.max_workers to 1
2. Put one object to riak cs and delete it, wait 1 sec.
3. Execute `riak-cs-admin gc batch 1` and wait.
4. GC does not finish at least several minutes.

### DIFF
--- a/src/riak_cs_gc_batch.erl
+++ b/src/riak_cs_gc_batch.erl
@@ -147,14 +147,12 @@ waiting_for_workers(_Msg, _From, State) ->
 
 %% @doc there are no all-state events for this fsm
 handle_event({batch_complete, WorkerPid, WorkerState}, StateName, State0) ->
-    %%?debugFmt("~w", [State0]),%% WorkerState#gc_worker_state.batch_count,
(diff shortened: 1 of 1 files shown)


kuenishi:
The patch looks good, although I couldn't reproduce in my MacOS with your steps.

kuenishi:
Repro'd. You'll also need `gc.max_workers=1` in riak-cs.conf.

shino:
Thanks! Updated the description.

kuenishi:
I've successfully reproduced this via eqc test; will follow up with it.

shino:
@borshop merge

PR #1248, opened 2015-09-14
test/gh1247 [RCS-292]

kuenishi:
- Use batch_size preset when starting riak_cs_gc_batch process
- Add QuickCheck test to address #1247 
- Add console option (`-b`) to `riak-cs-admin gc batch` to manually set batch size in each batch

### DIFF
--- a/src/riak_cs_gc_batch.erl
+++ b/src/riak_cs_gc_batch.erl
@@ -102,7 +102,7 @@ init([#gc_batch_state{
             %% EndKey <= BatchStart - Leeway
             _ = lager:info("Starting garbage collection: "
                            "(start, end) = (~p, ~p), "
-                           "leeway=~p, batch_start=~p, max_workers=~p, batch_size=~p",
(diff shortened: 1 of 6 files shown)


src/riak_cs_gc_batch.erl
```diff
@@ -101,8 +102,8 @@ init([#gc_batch_state{
             %% EndKey <= BatchStart - Leeway
             _ = lager:info("Starting garbage collection: "
                            "(start, end) = (~p, ~p), "
-                           "leeway=~p, batch_start=~p, max_workers=~p",
-                           [StartKey, EndKey, Leeway, BatchStart, MaxWorkers]),
+                           "leeway=~p, batch_start=~p, max_workers=~p, batch_size=~p",
```
shino: Log messages is a kind of end user / operator interface, too. It's better to use the same word here as command line option.

kuenishi:
I've force-pushed, again, sorry :imp:

shino:
The EQC prop actually catched the bug #1247  :cool:

shino:
@borshop merge

PR #1257, opened 2015-10-06
Supercluster by cuttlefish schema (aka multibag name chage)

shino:
A little change for `s/multibag/supercluster/` by cuttlefish schema.

This PR depends on following PRs:
- https://github.com/basho/riak_cs_multibag/pull/31
- https://github.com/basho/stanchion/pull/106

PR #1258, opened 2015-10-06
Update exometer_core version to 1.4

kuenishi:
To use this fix https://github.com/uwiger/setup/commit/346bff47d7ae4ee24dec74336a026280fbc8b404  included in setup 1.5. Found via https://github.com/basho/riak_cs/pull/1257#issuecomment-145771625 .


shino:
The change between 1.2 and 1.4 is small and it seems to me that no necessity to update in (or even after) the last minutes of 2.1.0 release. Thoughts?

kuenishi:
Yes. it was wrong and not an issue.

PR #1260, opened 2015-10-07
Don't use any bags

shino:
Enforce to use supercluster and some backward compatibility workaround.

Rule of thumb: Use `riak-cs.conf`.


shino:
For release note, see https://github.com/basho/riak_cs/pull/1257#issuecomment-146088919

shino:
@borshop merge

shino:
For release note, see https://github.com/basho/riak_cs/pull/1257#issuecomment-146088919

PR #1261, opened 2015-10-07
fall back on /etc/riak-cs/app.config when no generated one is found

joecaswell:
Prevent the error "cannot access /usr/lib64/riak-cs//var/lib/riak-cs/generated.config/vm.*.args" when a using app.config/vm.args instead of riak.conf, such as when upgrading from 1.5.x


shino:
Fixed by #1263 at `2.1` branch. The fix will be foward merged to `develop`.

kuenishi:
@joecaswell Do you need this included in 2.0? We'll include this in 2.1 anyway

PR #1262, opened 2015-10-08
Bugfix/still rt is failing oh my bad

shino:
Fix <del>two</del> three bugs, <del>only related with tests</del> almost only for test fix except small error handling of `riak_cs_console:resolve_sibling/2` function. <del>Two</del> Three commits fix <del>two</del> three different bugs respectively, independently.


kuenishi:
@borshop merge

PR #1265, opened 2015-10-09
Bugfix/gh1263 backport

ksauzz:


ksauzz:
Backporting gh1236 is also needed. I'll open another PR.

PR #1267, opened 2015-10-13
[WIP] Forward merge 2.0 dd0bc89 to 2.1

shino:
Still WIP because 2.1 scope is not strictly defined. This PR shows that merging 2.0 to 2.1 have only one conflict which is expected and simple.


shino:
Forward merge will happen after 2.0.x releases. Close this.

PR #1271, opened 2015-10-22
Put a cap on list objects v1

kuenishi:
Fix a bug by removing features that should have been removed in 2.1.0.


shino:
`riak_cs_list_objects_fsm.erl` can be removed?

kuenishi:
Removing everything carefully takes time so I just capped it. Removing v1 list objects will sure tempt me to clean up other codes.

kuenishi:
...and broke the dialyzer.

shino:
>  so I just capped it.

Understood :smile:  Then, is this PR urgent work?

kuenishi:
Never urgent.

shino:
It's not urgent but this PR is worth to be added for code simplicity.

PR #1273, opened 2015-11-30
Introduce intercepts into riak test

ksauzz:
[Intercepts](https://github.com/basho/riak_test#intercepts) are riak test's feature which changes behaviour of modules easily on riak test.

All riak_cs and stanchion instances load intercepts files from `riak_test/intercepts` in riak_cs repo during startup. This PR contains a copy of intercepts.erl from riak_test repo since it's necessary to be located intercepts directory. Without the copy, we should specify intercepts directory in riak_test repo, but it causes to load all intercepts files from there. It looks too much to me.

### DIFF
--- a/riak_test/intercepts/intercept.erl
+++ b/riak_test/intercepts/intercept.erl
@@ -54,7 +54,7 @@
 %%             functions.
 %%
 %%             E.g. `[{{update_perform,2}, sleep_update_perform}]'
--spec add(module(), module(), mapping()) -> ok.
+-spec add(module(), module(), mapping(), string()) -> ok.
 add(Target, Intercept, Mapping, OutDir) ->
     Original = ?ORIGINAL(Target),
     TargetAC = get_abstract_code(Target),
@@ -66,6 +66,7 @@ add(Target, Intercept, Mapping, OutDir) ->
     ok = compile_and_load(Original, OrigAC, OutDir),
     ok = compile_and_load(Target, ProxyAC, OutDir).
 
+-spec add(module(), module(), mapping()) -> ok.
 add(Target, Intercept, Mapping) ->
     add(Target, Intercept, Mapping, undefined).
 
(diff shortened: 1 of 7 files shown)


ksauzz:
ready for review.

riak_test/intercepts/intercept.erl
```diff
@@ -0,0 +1,392 @@
...
+%%
+%% `Intercept' - The module containing intercept definitions.
+%%               E.g. `hashtree_intercepts'
+%%
+%% `Mapping' - The mapping from target functions to intercept
+%%             functions.
+%%
+%%             E.g. `[{{update_perform,2}, sleep_update_perform}]'
+-spec add(module(), module(), mapping()) -> ok.
+add(Target, Intercept, Mapping, OutDir) ->
```
shino, line 58: spec says arity is 3 but 4 from the function definition

shino, line 58: Ah, `/3` is defined just below this function. It's not bad to declare spec anywhere, but closer is better?

shino:
It seems like _adding_ intercepts with empty mapping

```
rt_intercept:add(ErrCSNode, {riak_cs_riak_client, []})
```

makes the module behave as it is original one.

Not must but just a small refactoring idea, by defining

```
clear(Node, Module) -> add(Node, {Module, []}). 
```

in rt_intercept module, some function definitions in
intercepts modules can be removed and test code looks
a little more readable, I guess.

riak_test/intercepts/intercept.erl
ksauzz, line 58: Agree, but I don't like to change intercepts.erl in riak_cs repo since it should remain just a copy of one in riak_test repo.

shino, line 58: +1

ksauzz:
I prepared rt_intercept:clean/2 at https://github.com/basho/riak_test/pull/961, and also updated intercept's specs you mentioned above. :bowtie:

shino:
@borshop merge

PR #1281, opened 2015-12-15
don't use non-portable cp options

kuenishi:


shino:
Nice finding!

shino:
@borshop merge

PR #1284, opened 2015-12-24
Add a test to verify a retry logic after getting the block fails with nval=1

ksauzz:
Verify a retry logic to get a block after getting the block fails, and
introduce intercept.hrl as a copy of one in riak_test repo.


shino:
Nice addition! Big :+1: for partial failure test case!

shino:
@borshop merge

Issue #1292, opened 2016-02-02
Riak CS does not correctly handle subresources when creating an AWS signature [JIRA: RCS-349]


bashoshaun:
I have a customer that uses s3browser (s3browser.com) to interface with Riak CS.

Multi-part uploads fail for him.  This is due to the way Riak treats the uploadID subresource when creating an AWS signature.

Here's the steps taken and the results to provide the example:

User does a GET on a multi-part upload - this is the data captured:

```
GET /npo-training/020MB_txt?uploadId=YY0RyQnrTyiTOD3rCejqfQ%3D%3D HTTP/1.1
User-Agent: S3 Browser 5-7-3 http://s3browser.com
Authorization: AWS JQS79NNW-QPN_VZDYSCC:1e8MpKjK8F+ZRktLetv8IbWrmKs=
x-amz-date: Mon, 01 Feb 2016 14:43:31 GMT
Host: s3-emean-1.eecloud.nsn-net.net
```

Notice, that in the uploadID, part of it has been urlEncoded.  There's a %3D instead of an equals sign.

The signature is then created to authenticate with Riak CS.  Some clients, including s3cmd, would encode the following text:

```
GET



x-amz-date:Mon, 01 Feb 2016 14:43:31 GMT
/npo-training/020MB_txt?uploadId=YY0RyQnrTyiTOD3rCejqfQ==
```

As you can see, the uploadId was first URL decoded before signing.  In the user's example, it produced the hash of 1e8MpKjK8F+ZRktLetv8IbWrmKs= which matches the signature sent in the GET.

However, s3browser doesn't decode the URL.  It signs the following:

```
GET



x-amz-date:Mon, 01 Feb 2016 14:43:31 GMT
/npo-training/020MB_txt?uploadId=YY0RyQnrTyiTOD3rCejqfQ%3D%3D
```

This provides a different hash value, which means uploads are rejected.

Amazon has a description on how to properly create a signature [here](http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html).  Riak is treating the uploadId parameter as a _query string parameter_, which should be decoded from a URL encoding before being hashed.  However, Amazon considers an uploadID to be a _subresource_ which should be hashed in the encoding as it appears in the URL.

So when dealing with an uploadID, Riak CS should _not_ decode this part for the purposes of creating a signature (it can, of course, later decode it for other purposes).

[...]

Issue #1063, opened 2015-01-30
uploadId should be url-safe [JIRA: RCS-350]


kuenishi:
In v4 authentication, query parameters are used at signing like this (s3cmd example):

```
DEBUG: Canonical Request:
PUT
/40MB
partNumber=1&uploadId=8YSbOd4lQVqbgprjS-VKAw==
content-length:15728640
```

Characters like `=` should not be there or it should be url encoded by clients.
c.f. #1062

shino:
It's nice to avoid client-side corner case bugs because such query param normalization is specific to S3 AND v4 auth. But I want to note that it is not a bug of riak cs, and s3cmd should be fixed.

bashoshaun:
The uploadId is not considered a query parameter in Amazon's [signing documentation](http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html).  It is considered a subresource, and therefore falls under different rules of URL encoding/decoding as query parameters.

Subresources form part of the CanonicalizedResource and when signing, should be left either URL encoded or decoded, as specified in the original URI.

I believe Riak CS is treating uploadId as a query parameter, and following the rules for signing by first URL-decoding them.  Under these rules, it looks like Riak CS is exhibiting a bug.

Basho-JIRA:
Approved by Dave McCrory to create a 2.0 patch for Nokia.  

**_[posted via JIRA by Patricia Brewer]**_

Basho-JIRA:
Previously fixed and included in Riak CS 2.1.2

Nokia patch is issue RCS-366 

**_[posted via JIRA by Ted Burghart]**_

Basho-JIRA:
Changing this back to "To Do" status because the code, and the test for it, in the merged fix is incorrect.

The code happens to work when encoding/decoding UUIDs because their encoded length hits the midpoint in calculating how many '=' characters belong at the tail the properly encoded Base64, but variable length data will trigger a crash if the Base64 encoding would have an odd number of trailing '=' characters.

Additionally, the regular expression used to test the new code will never match, even if the code produced the wrong result.  Since the test only uses UUIDs, which by chance produce the right result because of their length, the test never sees data that exposes the decoding flaw. 

**_[posted via JIRA by Ted Burghart]**_

Basho-JIRA:
2.1.2 will need to be rebuilt, when we are the point of working on it again, correct?  

**_[posted via JIRA by Patricia Brewer]**_

[...]

PR #1293, opened 2016-02-04
Remove equal at every Base64 URL encode/decoding

kuenishi:
Multipart upload id in S3 does not includes equal '=' which made its
handling diverse among client libraries. Contrary to this, Riak CS
issues multipart upload ids that include '='. With this change
base64url module changes its behaviour, to automatically removing
trailing '=' at encoding time, and to automatically adding trailing
'=' at decoding time. This is harmless because in Base64 codes '=' is
just a filler.

This commit also removes several unused functions.

This addresses #1292 (RCS-349) and #1063 (RCS-350).

### DIFF
--- a/src/base64url.erl
+++ b/src/base64url.erl
@@ -90,17 +90,18 @@ urldecode_digit(D)  -> D.
 -ifdef(TEST).
 equal_strip_amend_test() ->
     %% TODO: rewrite this with EQC
-    [begin
-         UUID = druuid:v4(),
-         Encoded = base64url:encode(UUID),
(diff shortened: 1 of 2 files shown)


src/base64url.erl
```diff
@@ -27,30 +27,47 @@
```
ksauzz: I guess it would be better to add description about omitting '=' here.

ksauzz:
@borshop merge

kuenishi:
Note that during update to version which includes this, a '==' omitted uploadIds might be passed to CS node with old version, which may lead to process-crash by failing on decoding uploadId.

Issue #1025, opened 2014-11-25
Copying an object fails when connecting via https [JIRA: RCS-358]


iby:
Both AWS SDK (PHP) and s3cmd fail when I try to do copy an object within the same bucket. AWS SDK says `The specified key does not exist.`, while s3cmd says `AttributeError: 'NoneType' object has no attribute 'getchildren`. Here's the full output from s3cmd.

``` shell
$ s3cmd cp s3://tst/1234.jpg s3://tst/5678.jpg

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
    An unexpected error has occurred.
  Please try reproducing the error using
  the latest s3cmd code from the git master
  branch found at:
    https://github.com/s3tools/s3cmd
  If the error persists, please report the
  following lines (removing any private
  info as necessary) to:
   s3tools-bugs@lists.sourceforge.net

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Invoked as: /usr/local/bin/s3cmd cp s3://tst/1234.jpg s3://tst/5678.jpg
Problem: AttributeError: 'NoneType' object has no attribute 'getchildren'
S3cmd:   1.5.0-rc1
python:   2.7.6 (default, Sep  9 2014, 15:04:36) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)]
environment LANG=en_GB.UTF-8

Traceback (most recent call last):
  File "/usr/local/bin/s3cmd", line 2523, in <module>
    rc = main()
  File "/usr/local/bin/s3cmd", line 2441, in main
    rc = cmd_func(args)
  File "/usr/local/bin/s3cmd", line 752, in cmd_cp
    return subcmd_cp_mv(args, s3.object_copy, "copy", u"File %(src)s copied to %(dst)s")
  File "/usr/local/bin/s3cmd", line 739, in subcmd_cp_mv
    response = process_fce(src_uri, dst_uri, extra_headers)
  File "/usr/local/Cellar/s3cmd/1.5.0-rc1/libexec/S3/S3.py", line 628, in object_copy
    response = self.send_request(request)
  File "/usr/local/Cellar/s3cmd/1.5.0-rc1/libexec/S3/S3.py", line 883, in send_request
    e = S3Error(response)
  File "/usr/local/Cellar/s3cmd/1.5.0-rc1/libexec/S3/Exceptions.py", line 54, in __init__
[...]

PR #1300, opened 2016-03-15
Bugfix/cs1025

kuenishi:
Addresses #1025 (RCS-358).


kuenishi:
@tburghart Once you got all r_t modules running I'd like you to review this, as this includes a minimal example of S3 client implementation in Erlang, which might be a good example to learn S3 API.

When it comes to contradiction of proxy and SSL, I expressed my idea in the comment of 4d9b923 .

kuenishi:
Go see #1305.

PR #1303, opened 2016-03-17
Fix riak_test setup race condition

kuenishi:
This has been resident since 59f7e32 . Blame me because I was inappropriate.


tburghart:
+1 e658a77

kuenishi:
@borshop merge

Issue #1100, opened 2015-03-19
Riak CS should respond '400 Bad request' on multipart completion if the request has no part item. [JIRA: RCS-156]


ksauzz:
API incompatibility issue which is found by [this test](https://github.com/basho/s3-tests/blob/a38c056122ef2e8fbd5bc61b387463a16d228a59/s3tests/functional/test_s3.py#L4476-L4486).

Although Amazon S3 respond 400 Bad request if multipart completion request has no part item, Riak CS accept the request.

A response sample from S3.

```
 HTTP/1.1 400 Bad Request
 x-amz-request-id: 7DCAB2138707FE68
 x-amz-id-2: hgAmdxlfDD7YxWsoktlpUx38KltSbSr4SXvjPi6wSQOPtAmnYMQYKYMuaLpvJ7Qn
 Content-Type: application/xml
 Date: Thu, 19 Mar 2015 16:23:46 GMT
 Connection: close
 Server: AmazonS3

<Error><Code>MalformedXML</Code><Message>The XML you provided was not well-formed or did not validate against our published schema</Message><RequestId>7DCAB2138707FE68</RequestId><HostId>hgAmdxlfDD7YxWsoktlpUx38KltSbSr4SXvjPi6wSQOPtAmnYMQYKYMuaLpvJ7Qn</HostId></Error>
```

mrallen1:
For clarity's sake, this means the Content-Length is 0, correct?

jadeallenx:
OK, so it looks like the test is supposed to do the following:
- Initiate a multipart upload
- Send 0 upload parts
- Send the "completed" upload message

Is my reading of this test case correct?

kuenishi:
Yes, you're right, Mark.

Basho-JIRA:
See https://github.com/basho/riak_cs/pull/1304 

**_[posted via JIRA by Mark Allen]**_

PR #1304, opened 2016-03-18
Return 400 Bad Request if multipart upload does not contain upload parts

jadeallenx:
During a multipart upload, the "complete" message should contain a list of part ids and MD5 hashes. If it is empty, we should return a 400 Bad Request, Malformed XML message like S3.  The associated test for this test is [here](https://github.com/basho/s3-tests/blob/riakcs-2.1/s3tests/functional/test_s3.py#L4596-L4606). Addresses #1100

### DIFF
--- a/src/riak_cs_wm_object_upload_part.erl
+++ b/src/riak_cs_wm_object_upload_part.erl
@@ -117,6 +117,11 @@ process_post(RD, Ctx=#context{local_context=LocalCtx, riak_client=RcPid}) ->
     case {parse_body(Body), catch base64url:decode(UploadId64)} of
         {bad, _} ->
             {{halt,477}, RD, Ctx};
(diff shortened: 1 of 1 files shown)


tburghart:
+1 b4491c6

tburghart:
@borshop merge

kuenishi:
@mrallen1  The test you mentioned is omitted because of the decorator `fails_on_rcs` ; https://github.com/basho/s3-tests/blob/riakcs-2.1/s3tests/functional/test_s3.py#L4599 which is written [here](https://github.com/basho/riak_cs/blob/2.1/client_tests/python/ceph_tests/Makefile#L13).

kuenishi:
I meant [this line](https://github.com/basho/s3-tests/blob/riakcs-2.1/s3tests/functional/test_s3.py#L4599) must be removed and get this tested.

PR #1305, opened 2016-03-18
Fix for issue 1025 [RCS-358]

tburghart:
This merges branches bugfix/cs1025 with current 2.1 (for full-suite testing) and replaces the hard-coded host/port destinations for the ssl object copy test with values from the configuration.

It's smaller to merge back to 2.1 than back to bugfix/cs1025, so it supersedes https://github.com/basho/riak_cs/pull/1300


kuenishi:
+1 772902c

kuenishi:
@borshop merge

PR #1308, opened 2016-03-25
skip repl_v3_test if not using an ee configuration for riak_test

tburghart:
if the build_type from the configuration is not 'ee' the test immediately returns 'pass'


kuenishi:
+1 6a4de74

kuenishi:
@borshop merge

PR #1309, opened 2016-03-26
Stop access_stats_test from being flappy

tburghart:
Previously, access_stats_test performed all of its operations very quickly, so it would intermittently request stats with the same begin and end time and get an empty result set back from CS.

Whether CS should return an empty result set in such a case is a different question, depending on whether a time specification represents an instant or an interval, but in either case that's not what this test is for.  Instead, it now simply spends enough time to make sure the begin and end times are different.

This PR inadvertently includes the changes in https://github.com/basho/riak_cs/pull/1308 so that PR should be merged before this one so they each reflect the appropriate change set.

### DIFF
--- a/riak_test/src/rtcs.erl
+++ b/riak_test/src/rtcs.erl
@@ -1,6 +1,6 @@
-%% ---------------------------------------------------------------------
+%% -------------------------------------------------------------------
 %%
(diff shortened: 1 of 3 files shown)


kuenishi:
@borshop merge

kuenishi:
+1 f014d51

kuenishi:
@borshop merge

PR #1313, opened 2016-05-18
Change loglevel regarding secret

ksauzz:
This should be written as debug level.


ksauzz:
My bad. This is **KEY**.

PR #1316, opened 2016-06-04
Corrected fix for RCS-350

tburghart:
This is a revision of the changes in https://github.com/basho/riak_cs/pull/1293

That code happened to work when encoding/decoding UUIDs because their encoded length hit the midpoint in calculating how many '=' characters belong at the tail the properly encoded Base64. The padding length was calculated incorrectly, though, so variable length data wound up improperly formatted and triggered a crash if the Base64 encoding had an odd number of trailing '=' characters.

Additionally, the regular expression used in the test code would never match an improperly encoded result. Since the test only used UUIDs, which by chance produced the right result because of their length, the test never saw data that exposed the padding flaw.

These changes are cherry-picked from the RCS-366 patch branch, which was based on a different version, so some of its changes were already present (hence the apparently trivial changes). The relevant changes are those in the `base64url` module.

### DIFF
--- a/riak_test/tests/mp_upload_test.erl
+++ b/riak_test/tests/mp_upload_test.erl
@@ -1,6 +1,6 @@
-%% ---------------------------------------------------------------------
+%% -------------------------------------------------------------------
 %%
-%% Copyright (c) 2007-2013 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2007-2016 Basho Technologies, Inc.
 %%
 %% This file is provided to you under the Apache License,
 %% Version 2.0 (the "License"); you may not use this file
@@ -16,7 +16,7 @@
 %% specific language governing permissions and limitations
 %% under the License.
 %%
-%% ---------------------------------------------------------------------
+%% -------------------------------------------------------------------
 
 -module(mp_upload_test).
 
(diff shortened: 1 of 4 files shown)


riak_test/tests/mp_upload_test.erl
```diff
@@ -154,6 +154,7 @@ aborted_upload_test_case(Bucket, Key, Config) ->
     lager:info("Initiating multipart upload"),
     InitUploadRes = erlcloud_s3_multipart:initiate_upload(Bucket, Key, [], [], Config),
     UploadId = erlcloud_s3_multipart:upload_id(InitUploadRes),
+    lager:info("Upload ID: ~p", [UploadId]),
```
JeetKunDoug: Do we still need these as `lager:info`? Could they be `lager:debug` instead?

riak_test/tests/mp_upload_test.erl
```diff
@@ -216,6 +217,7 @@ basic_upload_test_case(Bucket, Key, Config) ->
...
```
JeetKunDoug: Ditto

riak_test/tests/mp_upload_test.erl
```diff
@@ -274,6 +276,7 @@ parts_too_small_test_case(Bucket, Key, Config) ->
     lager:info("Initiating multipart upload (bad)"),
     InitUploadRes = erlcloud_s3_multipart:initiate_upload(Bucket, Key, [], [], Config),
     UploadId = erlcloud_s3_multipart:upload_id(InitUploadRes),
+    lager:info("Upload ID: ~p", [UploadId]),
```
JeetKunDoug: Ditto

JeetKunDoug:
+1 f69aea8

Issue #1060, opened 2015-01-27
v4 authentication - aws-chunked upload [JIRA: RCS-115]


Basho-JIRA:
- [ ] Investigate which client/SDK supports aws-chunked uploads
  - Java SDK implements aws-chunked
    https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/AWSS3V4Signer.java#L53
- [ ] Authorization header / aws-chunked / no tranfer encoding case
  - Content-Length header does not represent object size, how to know/compute?
  - s3cmd and boto do not implement this.
- [ ] Authorization header / aws-chunked / chunked tranfer encoding case
  - No Content-Length header, c-l will be determined after transfer?
- [ ] Access stats log for chunked upload
  - there are possibly three kinds of length: HTTP request payload, HTTP Content-Length, Object size

Issue #1058, opened 2015-01-27
v4 authentication - query parameter [JIRA: RCS-113]


Basho-JIRA:
- [ ] Calculate SHA256 of Payload (along with MD5? Necessary?)
  - What error happens if sha256 is wrong?

shino:
As for Multipart Upload Part, AWS S3 adds 'ETag' which is MD5 hash of the content of part object.
Also, `ETag` of completed object is MD5 hash of concatenated parts' MD5 hash values. [1]

If riak cs follow this S3 behaivior, both MD5 and SHA256 of payload should be calculated.

[1] https://gist.github.com/shino/29676f0cf438fb56f3df

shino:
For faked `x-amx-contet-sha256`, S3 responds with error
`XAmzContentSHA256Mismatch` [1].  It can be thought that
authentication is 2-phase, signature calculation (require headers
only) and post accept body sha256 verification.

```
<?xml version="1.0" encoding="UTF-8"?>
<Error>
  <Code>XAmzContentSHA256Mismatch</Code>
  <Message>The provided 'x-amz-content-sha256' header does not match what was computed.</Message>
  <ClientComputedContentSHA256>0000000000000000000000000000000000000000000000000000000000000000</ClientComputedContentSHA256>
  <S3ComputedContentSHA256>809fa1ed21450f59827d1e9aec720bbc4b687434fa22283c6cb5dd82a47ab9c0</S3ComputedContentSHA256>
  <RequestId>A3D32EA3F062DB15</RequestId>
  <HostId>vE48bS9V+VHXMkrCiKdcTGbL/wiRPzFhOor4O5TV1ZT5NUeOFvHx9dOjkHvFrwCv21CvQOXJH4w=</HostId>
</Error>
```

[1] https://gist.github.com/shino/54cdcb2c8475d9699789 (it includes
faked Content-MD5 case too)

PR #1351, opened 2018-07-31
Remove EE and OSS differentiation. signature_v4 access via riak-cs.conf

Bob-The-Marauder:
Some parts commented rather than removed at this stage in case they later prove necessary.


shino:
Some comments from old guy.
At first, I'm very happy to look at the new activity in this repo :smile:

This comment does not to intent two points below should be addressed
before merge, but just to notice some old issues/concerns.

1. `rebar.config.script` was just for OSS/EE deference. It can be removed completely
   if there is only one OSS flavor. `rebar.config` will just suffice.
2. v4 authentication is *NOT COMPLETED* yet.  The current status is not as secure as S3 is [1], and some API does not support v4 [2] [3].
   I agree to make it accessible from conf, but it may be better to include some "unfinished" caution in the `@doc`.

[1] https://github.com/basho/riak_cs/issues/1059
[2] https://github.com/basho/riak_cs/issues/1058
[3] https://github.com/basho/riak_cs/issues/1060

Bob-The-Marauder:
I tried removing `rebar.config.script` but CS refused to make packages without it. Guessing it's a packaging issue. I will see if I can correct for that in future updates.
Thank you for the additional v4 comments. I have changed the comment in our internal copy of riak-cs.conf to "experimental support" and that will probably be reflected here sooner or later.

shino:
Thanks for the update.

Just for the future change, a comment about rebar.config.script (my previous comment was insufficient, sorry).

It creates pkg.vars.config file for packaging. Because the file was different for OSS and EE versions, it was created dynamically. Now OSS version is the only one, the file can be static.
What should be done are:
- remove the line from .gitignore https://github.com/basho/riak_cs/blob/develop/.gitignore#L14, and
- add the file to git

Thanks again!

shino:
Oops, I missed another thing the script does.

`git mv dialyzer.ignore-warnings.ee dialyzer.ignore-warnings` is needed.

shino:
Looks nice!

Bob-The-Marauder:
Thank you for your help. Looking to tag this as 2.1.2rc2

